PR #25, opened 2018-12-06
update readme alias section

domgreen:
I assume this is what you intended ... please close if not :+1:

---
This change is [Reviewable](https://reviewable.io/reviews/peltoche/lsd/25)


domgreen:
Actually you probably want to just remove the d from the second line ... will update

Peltoche:
Indeed... Thanks for the PR!

PR #34, opened 2018-12-08
Set license instead of license-file

Crestwave:
From The Cargo Book:

```
# This is an SPDX 2.1 license expression for this package.  Currently
# crates.io will validate the license provided against a whitelist of
# known license and exception identifiers from the SPDX license list
# 2.4.  Parentheses are not currently supported.
#
# Multiple licenses can be separated with a `/`, although that usage
# is deprecated.  Instead, use a license expression with AND and OR
# operators to get more explicit semantics.
license = "..."

# If a package is using a nonstandard license, then this key may be specified in
# lieu of the above key and must point to a file relative to this manifest
# (similar to the readme key).
license-file = "..."
```


Peltoche:
LGTM, thanks @Crestwave  for the PR.

PR #35, opened 2018-12-08
show - for size when if it is not a file

### LINKED ISSUE #23: Do not print the size in case of folders or symlinks
The current output is: 
```sh
drwxr-xr-x  peltoche  users    4 KB  Thu Dec  6 00:34:14 2018    src/
.rw-r--r--  peltoche  users    8 KB  Thu Dec  6 00:29:01 2018    Cargo.lock
```
The 4KB is the folder size but doesn't give much information as it's not the content size. In order to simplify the output I propose to make the output looks like:
```sh
drwxr-xr-x  peltoche  users       -  Thu Dec  6 00:34:14 2018    src/
.rw-r--r--  peltoche  users    8 KB  Thu Dec  6 00:29:01 2018    Cargo.lock
```

The size is replaced by a dash `-` like the `exa` program.

meain:
fixes #23 

screenshot

### DIFF
--- a/src/batch.rs
+++ b/src/batch.rs
@@ -57,6 +57,20 @@ impl Batch {
         for meta in &self.0 {
             let icon = icons.get(&meta.name);
 
+            let mut size_str;
+            if let Some(ref size) = meta.size {
+                size_str = size.render(colors, max_size_value_length, max_size_unit_length);
+            } else {
+                size_str = ANSIString::from(
+                    " ".repeat(max_size_value_length + max_size_unit_length) + "-",
+                );
+            }
+
+            let mut link_str = ANSIString::from("");
+            if let Some(ref symlink) = meta.symlink {
+                link_str = symlink.render(colors);
+            }
+
             let strings: &[ANSIString] = &[
                 meta.file_type.render(colors),
                 meta.permissions.render(colors),
@@ -65,8 +79,7 @@ impl Batch {
                 ANSIString::from("  "),
                 meta.owner.render_group(colors, max_group_length),
                 ANSIString::from("  "),
-                meta.size
-                    .render(colors, max_size_value_length, max_size_unit_length),
+                size_str,
                 ANSIString::from("  "),
                 meta.date.render(colors),
                 ANSIString::from("  "),
@@ -112,12 +125,14 @@ impl Batch {
         let mut max_unit_size: usize = 0;
 
         for meta in &self.0 {
-            if meta.size.render_value().len() > max_value_length {
-                max_value_length = meta.size.render_value().len();
-            }
+            if let Some(ref size) = meta.size {
+                if size.render_value().len() > max_value_length {
+                    max_value_length = size.render_value().len();
+                }
 
-            if meta.size.render_unit().len() > max_unit_size {
-                max_unit_size = meta.size.render_unit().len();
+                if size.render_unit().len() > max_unit_size {
+                    max_unit_size = size.render_unit().len();
+                }
             }
         }
 
--- a/src/batch.rs
+++ b/src/batch.rs
@@ -57,11 +57,6 @@ impl Batch {
         for meta in &self.0 {
             let icon = icons.get(&meta.name);
 
-            let mut link_str = ANSIString::from("");
-            if let Some(ref symlink) = meta.symlink {
-                link_str = symlink.render(colors);
-            }
-
             let strings: &[ANSIString] = &[
                 meta.file_type.render(colors),
                 meta.permissions.render(colors),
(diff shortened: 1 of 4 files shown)


src/batch.rs
```diff
@@ -50,6 +50,15 @@ impl Batch {
         let (max_size_value_length, max_size_unit_length) = self.detect_size_lenghts();
 
         for meta in &self.0 {
+            let mut size_str;
+            if let Some(ref size) = meta.size {
+                size_str = size.render(colors, max_size_value_length, max_size_unit_length);
+            } else {
+                size_str = ANSIString::from(
+                    " ".repeat(max_size_value_length + max_size_unit_length) + "-",
+                );
```
Peltoche: I'm not really enthusiastic about adding some Size related logic inside the Batch struct. The Symlink logic just above is a bad example and should be moved into the Symlink.render() method.

src/meta/mod.rs
```diff
@@ -25,7 +25,7 @@ pub struct Meta {
     pub date: Date,
     pub owner: Owner,
     pub file_type: FileType,
-    pub size: Size,
+    pub size: Option<Size>,
```
Peltoche: I think you can remove the Option here. Instead of calling `render()` if there is a Size or printing "-" you can move this logic inside the Size.render_value() which would return an empty String if the it's a directory or returning the actual size. Same for Size.render_unit().

src/meta/mod.rs
```diff
@@ -51,7 +51,7 @@ impl Meta {
         Some(Meta {
             path: path.to_path_buf(),
             symlink: SymLink::from_path(&path),
-            size: Size::from(&metadata),
+            size: Size::from_bytes(&metadata),
```
Peltoche: I think you will need a Size::new() method, like the Name struct with the file_type argument.

src/meta/size.rs
```diff
@@ -16,16 +16,16 @@ pub struct Size {
...
+        }
 
-        Size::from_bytes(len as i64)
+        Some(Size::get_file_size(meta.len() as i64))
     }
-}
 
-impl Size {
-    fn from_bytes(len: i64) -> Self {
+    fn get_file_size(len: i64) -> Self {
```
Peltoche: Why do you change the `from_bytes` into `get_file_size`? In my understanding the convention when you initialize a struct based on something the conventional name is :  `from_{the thing}`

meain: I did this to make the testing working, there still was `from_bytes`.
I separated the function which gets the file size separate to test it out otherwise I had to mock the `fs::Metadata` pass to `from_bytes`.

In any case the latest commit fixes this.

src/meta/size.rs
```diff
@@ -12,44 +12,51 @@ pub enum Unit {
 
 #[derive(Debug, PartialEq, Eq)]
 pub struct Size {
-    value: i64,
-    unit: Unit,
+    value: Option<i64>,
+    unit: Option<Unit>,
```
Peltoche: You can avoid those Option by adding a `None` value to the Unit value.

src/meta/size.rs
```diff
@@ -112,11 +127,14 @@ impl Size {
 
     pub fn render_unit(&self) -> String {
         match self.unit {
-            Unit::Byte => String::from("B"),
-            Unit::Kilo => String::from("KB"),
-            Unit::Mega => String::from("MB"),
-            Unit::Giga => String::from("GB"),
-            Unit::Tera => String::from("TB"),
+            Some(ref unit) => match unit {
```
Peltoche:
```suggestion
            Some(ref unit) => match unit {
                Une::None => String::from("-"),
```

src/meta/size.rs
```diff
@@ -12,44 +12,51 @@ pub enum Unit {
...
 
 impl<'a> From<&'a Metadata> for Size {
     fn from(meta: &Metadata) -> Self {
         let len = meta.len();
 
-        Size::from_bytes(len as i64)
+        if meta.is_file() {
+            Size::from_bytes(len as i64)
+        } else {
+            Size {
```
Peltoche: The Unit::None would allows you to have something like:
```
Size{
   value: 0,
   une: Unit::None,
}
```

meain:
looks good?

meain:
can we merge it?

Peltoche:
It seems that your continue to have some conflics. I think you need to resync your master with mine.
Just in case if you not familiar with the rebases:
```sh
git remote add mainstream git@github.com:Peltoche/lsd.git
git checkout no-size-for-folder
git pull --rebase mainstream master
# Here you will have some conflics to resolve
git push origin no-size-for-folder
```

PR #47, opened 2018-12-10
Color executable files green

LippyBoy:


src/meta/mod.rs
```diff
@@ -48,7 +48,8 @@ impl Meta {
         };
 
         let file_type = FileType::from(&metadata);
-        let name = Name::new(&path, file_type);
+        let permissions = Permissions::from(&metadata);
```
Peltoche: If you initialize permissions here, you should use the variable for the Meta initialization below.

l.58 -- `permissions: Permissions::from(&metadata),`
l.58 ++ `permissions,`

src/color.rs
```diff
@@ -35,6 +35,9 @@ pub enum Elem {
     FileLarge,
     FileMedium,
     FileSmall,
+
+    /// Misc
+    ExecutableFile,
```
Peltoche: You can move this into the` Node Type` part I think.

Peltoche:
Can you colorize the icon too? This can be done into the `icon.rs` file.

LippyBoy:
I added it, but I feel that the logic shouldn't be duplicated between name and icon, but I don't know where to put it.

Peltoche:
Totally agree for the duplicated logic. Maybe keeping the Icons struct, and moving the Icon.render() logic into Name.render() would make it better. 

So in practice it would remove the `Icon` struct and the `Name.render()` proto could looks like:
```rust
    pub fn render(&self, colors: &Colors, icons: &Icons) -> ColoredString {}
```
And Name.render() is charged to render the name and the icon based on the given icon list (Icons).

LippyBoy:
This would change how --tree works though, since it also uses name.render()

LippyBoy:
> This would change how --tree works though, since it also uses name.render()

I actually like it better now, with --tree also showing icons.

src/batch.rs
```diff
@@ -37,9 +37,7 @@ impl Batch {
         let mut res = Vec::with_capacity(self.0.len());
 
         for meta in &self.0 {
-            let icon = icons.get(&meta.name);
-
-            let strings: &[ANSIString] = &[icon.render(colors), meta.name.render(colors)];
+            let strings: &[ANSIString] = &[meta.name.render(colors, icons)];
```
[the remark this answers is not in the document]
LippyBoy: Is it still relevant with the indicator?

PR #67, opened 2018-12-13
Add option to show relative time

meain:
Might have to decide on what flag to pass and any other specifics.

Currently if you pass `--date relative` it shows relative date.

screenshot


src/app.rs
```diff
@@ -49,4 +49,12 @@ pub fn build_app() -> App<'static, 'static> {
                 .long("tree")
                 .help("Recurse into directories and present the result as a tree"),
         )
+        .arg(
+            Arg::with_name("date")
+                .long("date")
+                .possible_value("normal")
```
Peltoche: I would vote for replacing "normal" by "date". It make more logic to me to have date/relative instead of default/relative. Another proposition would be "classic".

What do you think of it?

src/batch.rs
```diff
@@ -68,7 +69,7 @@ impl Batch {
                 meta.size
                     .render(colors, max_size_value_length, max_size_unit_length),
                 ANSIString::from("  "),
-                meta.date.render(colors),
+                meta.date.render(colors, max_date_length, flags.date),
```
Peltoche: You should give the Flags struct directly instead of a bool. It would avoid to change the `Render()` signature each time need an another flag (Spoiler: it's coming soon).

src/batch.rs
```diff
@@ -107,6 +108,18 @@ impl Batch {
         max
     }
 
+    fn detect_date_lenght(&self, date_flag: DateFlag) -> usize {
```
Peltoche: Same here. All the variables inside the `Flags` structs will be soon private and replaced by getter functions used to wrap the logic around the flags like the priority between them.

src/meta/date.rs
```diff
@@ -36,7 +44,24 @@ impl Date {
...
+        colors.colorize(content, elem)
+    }
+
+    pub fn date_string(&self, date_flag: DateFlag) -> String {
+        match date_flag {
+            DateFlag::Normal => self.0.ctime().to_string(),
+            DateFlag::Relative => {
+                let now = time::now();
+                let ht = HumanTime::from(self.0 - now);
+                format!("{}", ht)
```
Peltoche: You can avoid the format by using: 
```rust
HumanTime::from(self.0 - now).to_text_en(Accuracy::Rough, Tense::Past)
```

src/app.rs
meain: `date` sounds good. `classic` can be a bit ambiguous.

PR #74, opened 2018-12-14
Accept the same flag several times

### LINKED ISSUE #73: Flags should be able to Accept Multiple Times
.. with the last value counted.

I aliased `lsd` with `lsd=lsd --icon=never`, then I want icons to print, but:

```
$ lsd --icon=auto
error: The argument '--icon <icon>' was provided more than once, but cannot be used multiple times
```

which is not expected. Consider to adopt this same behaviour as `ls`?

Peltoche:
Fix #73

### DIFF
--- a/src/flags.rs
+++ b/src/flags.rs
@@ -15,16 +15,21 @@ pub struct Flags {
 
 impl<'a> From<ArgMatches<'a>> for Flags {
     fn from(matches: ArgMatches) -> Self {
-        Flags {
+        let color_inputs: Vec<&str> = matches.values_of("color").unwrap().collect();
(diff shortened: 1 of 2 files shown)


src/flags.rs
```diff
@@ -54,6 +59,7 @@ pub enum WhenFlag {
 
 impl<'a> From<&'a str> for WhenFlag {
     fn from(when: &'a str) -> Self {
+        println!("foobar: {}", when);
```
mitnk: ???

Peltoche: arf... fixed.d

PR #78, opened 2018-12-14
Add options to sort by time and to reverse sorting order.

### LINKED ISSUE #4: sort by mod time
Just curious if a sort by mod time (ie, ls -t) is in the plan.....

boxdot:
Resolves #4.

### DIFF
--- a/src/batch.rs
+++ b/src/batch.rs
@@ -1,6 +1,6 @@
 use ansi_term::{ANSIString, ANSIStrings};
 use color::Colors;
-use flags::Flags;
+use flags::{Flags, SortFlag, SortOrder};
 use icon::Icons;
 use meta::FileType;
 use meta::Meta;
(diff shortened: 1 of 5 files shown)


src/batch.rs
```diff
@@ -30,8 +30,8 @@ impl Batch {
         self.0.len()
     }
 
-    pub fn sort(&mut self) {
-        self.0.sort_unstable_by(sort_by_meta);
+    pub fn sort(&mut self, sort: (SortFlag, SortOrder)) {
```
Peltoche: You should pass the `Flags` struct directly here. Soon the variables inside `Flags` will be private and only accessible via getters in order to keep the flag priority logic into the flags package.

src/flags.rs
```diff
@@ -8,6 +8,7 @@ pub struct Flags {
     pub display_tree: bool,
     pub display_indicators: bool,
     pub recursive: bool,
+    pub sort: (SortFlag, SortOrder),
```
Peltoche: I think you should split this tuple into two variable as it is two different flags and if you pass all the struct to the `sort()` method you would have no need of it.

boxdot:
Thanks for reviewing. Fixed your comments and added a unit test.

PR #89, opened 2018-12-23
WIP: Fix directory icon with extensions

cat12079801:
Hi @Peltoche
I fixed a bug probably, and added some tests for `Icons::get`.
Please see below image. Is this expected works? (at master branch d1fdee3)
Directory with extensions shows extension icon.

I have two questions.

First, my added test is failed because `yarn.lock` key in `Icons::get_default_icons_by_extension()`.
`yarn.lock` is duplicate with `Icons::get_default_icons_by_name()`.
Could I remove `yarn.lock` key from `Icons::get_default_icons_by_extension()`?

Second, I don't have confidence for my implementation. Because I'm a novice of Rust :(
Would you point out if there is a bad implementation?

Thank you for reading my poor English.

### DIFF
--- a/src/icon.rs
+++ b/src/icon.rs
@@ -36,13 +36,10 @@ impl Icons {
         let mut res = String::with_capacity(4 + ICON_SPACE.len()); // 4 == max icon size
 
         // Check directory.
-        match name.file_type() {
-            FileType::Directory => {
-                res += "\u{f115}"; // 
-                res += ICON_SPACE;
-                return res;
-            }
-            _ => {}
+        if name.file_type() == FileType::Directory {
+            res += "\u{f115}"; // 
+            res += ICON_SPACE;
+            return res;
         }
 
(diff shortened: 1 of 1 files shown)


Peltoche:
Hi @cat12079801,

Thanks for the PR and the tests!

For the folder issue, it's the expected behavior (for now at least). As you can see inside the `Icons::get_default_icons_by_name()` method there is icons for some specific folders (.git / .Trash / etc), this is the same behavior than the `colorls` project. This means that's not a bug. 

However, like you I'm not fan of the customs icons for the folders, I found the output harder to read, so let's merge it once it's finished.

For the `yarn.lock` issue your are right, it's not an extension and shouldn't be inside `Icons::get_default_icons_by_extension()`. So yep, you can remove it.

Peltoche:
LGTM, I rebase and merge

PR #115, opened 2019-01-30
Remove Cargo "homepage" manifest attribute

Enet4:
The [C-METADATA](https://rust-lang-nursery.github.io/api-guidelines/documentation.html#c-metadata) guideline suggests that the `homepage` field should only be defined when the crate has a dedicated website other than the repository or documentation sites. As such, it is best to leave it out for the time being.


Peltoche:
Hi @Enet4,

Thanks for the contribution!

PR #116, opened 2019-02-04
Add Windows support

danieldulaney:
I apologize for the size of this PR, but there are a bunch of changes that have to be made at the same time.

The fundamental issue is that Windows has a completely different model for permissions from Unix. Rather than using UIDs and GIDs, Windows uses a structure called a Security Identifier or [SID](https://docs.microsoft.com/en-us/windows/desktop/api/winnt/ns-winnt-_sid), which ~can~ should only be created or modified using WinAPI functions. And rather than the simple permission bits, Windows uses a structure called an Access Control List or [ACL](https://docs.microsoft.com/en-us/windows/desktop/api/winnt/ns-winnt-acl). An ACL can be compared to a SID to get that SID's access rights, which we can translate into Unix-style access bits. The owner and group SIDs and the file's ACL are contained in a structure called a [Security Descriptor](https://docs.microsoft.com/en-us/windows/desktop/api/winnt/ns-winnt-security_descriptor). So the general workflow is:

- Get the file's security descriptor, including the owner SID, group SID, and ACL
- Look up the account name and domain associated with the owner SID
- Look up the account name and domain associated with the group SID
- Create a SID that represents "everyone on the system" (the [WinWorldSid](https://docs.microsoft.com/en-us/windows/desktop/api/winnt/ne-winnt-well_known_sid_type))
- Evaluate the ACL with each of those SIDs to determine that SID's access rights
- Translate those access rights into Unix access bits

All of these (except the last one) rely on the Windows API. Unfortunately, there aren't any existing safe wrappers for this part of the WinAPI, so we have to fall back on the unsafe functions provided by the [winapi](https://docs.rs/winapi/*/x86_64-pc-windows-msvc/winapi/) crate. Therefore, I created a new module, `windows_utils`, which wraps all of the unsafe bits in a safe-to-call function that handles all of the unsafe API calls and memory management.

The other issue is that Windows doesn't allow you to easily separate getting the owner/group of a file from getting the permissions. Both of them use the same unsafe API call ([GetNamedSecurityInfoW](https://docs.microsoft.com/en-us/windows/desktop/api/aclapi/nf-aclapi-getnamedsecurityinfow)) to get the security descriptor. Therefore, I created a single function that generates both the Owner and the Permissions data structures. Every unsafe block has a comment listing the soundness assumptions it makes.

Additionally device files and pipes aren't in the main Windows file system tree, and symlinks are rare to see on Windows, so those aren't implemented for `FileType`.

Most of the rest of the changes fall into a few categories:
- Added an API for creating `Owner` structures directly from `String`s
- Porting tests to Windows (lots of them use `touch` and other non-Windows utilities which have Windows equivalents)
- Disabling a few tests on Windows (pipes, device files, and symlinks)

### DIFF
--- a/src/meta/filetype.rs
+++ b/src/meta/filetype.rs
@@ -3,6 +3,7 @@ use crate::meta::Permissions;
 use std::fs::Metadata;
 
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
+#[cfg_attr(windows, allow(dead_code))]
 pub enum FileType {
     BlockDevice,
     CharDevice,
--- a/src/meta/filetype.rs
+++ b/src/meta/filetype.rs
@@ -47,7 +47,22 @@ impl FileType {
 
     #[cfg(windows)]
     pub fn new(meta: &Metadata, permissions: &Permissions) -> Self {
-        unimplemented!()
+        let file_type = meta.file_type();
+
+        if file_type.is_file() {
+            FileType::File {
+                exec: permissions.is_executable(),
+                uid: permissions.setuid,
+            }
+        } else if file_type.is_dir() {
+            FileType::Directory {
+                uid: permissions.setuid,
+            }
+        } else if file_type.is_symlink() {
+            FileType::SymLink
+        } else {
+            FileType::Special
+        }
     }
 }
 
--- a/src/meta/filetype.rs
+++ b/src/meta/filetype.rs
@@ -1,7 +1,6 @@
 use crate::color::{ColoredString, Colors, Elem};
 use crate::meta::Permissions;
 use std::fs::Metadata;
-use std::os::unix::fs::FileTypeExt;
 
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
 pub enum FileType {
@@ -16,7 +15,10 @@ pub enum FileType {
 }
 
 impl FileType {
+    #[cfg(unix)]
     pub fn new(meta: &Metadata, permissions: &Permissions) -> Self {
+        use std::os::unix::fs::FileTypeExt;
+
         let file_type = meta.file_type();
 
         if file_type.is_file() {
@@ -41,6 +43,11 @@ impl FileType {
         } else {
             FileType::Special
         }
+    }
+
+    #[cfg(windows)]
+    pub fn new(meta: &Metadata, permissions: &Permissions) -> Self {
+        unimplemented!()
     }
 }
 
--- a/src/meta/filetype.rs
+++ b/src/meta/filetype.rs
@@ -70,7 +70,7 @@ impl FileType {
     }
 }
 
-#[cfg(test)]
+#[cfg(all(test, unix))]
 mod test {
     use super::FileType;
     use crate::color::{Colors, Theme};
--- a/src/meta/filetype.rs
+++ b/src/meta/filetype.rs
@@ -86,19 +86,26 @@ impl FileType {
     }
 }
 
-#[cfg(all(test, unix))]
+#[cfg(test)]
 mod test {
     use super::FileType;
     use crate::color::{Colors, Theme};
+    #[cfg(unix)]
     use crate::meta::Permissions;
+    use crate::meta::Meta;
     use ansi_term::Colour;
+    #[cfg(unix)]
     use std::fs::File;
+    #[cfg(unix)]
     use std::os::unix::fs::symlink;
+    #[cfg(unix)]
     use std::os::unix::net::UnixListener;
+    #[cfg(unix)]
     use std::process::Command;
     use tempdir::TempDir;
 
     #[test]
+    #[cfg(unix)] // Windows uses different default permissions
     fn test_file_type() {
         let tmp_dir = TempDir::new("test_file_type").expect("failed to create temp dir");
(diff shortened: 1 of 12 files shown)


gilescope:
Will give this a try on my windows box. Was just looking for windows support. For what its worth the PR tests pass and the command seems to still work fine on OSX.

danieldulaney:
Let me know if there are any bugs! We definitely need more tests of the Windows code, as well as turning on Travis.

BTW, the Cygwin documentation (https://www.cygwin.com/cygwin-ug-net/ntsec.html) is a useful reference. They have a much wider scope (mapping SIDs onto UIDs/GIDs, creating ACLs from Unix permissions bits), but they had to solve the same basic problem.

Peltoche:
Hello @danieldulaney ,

Thanks a lot for this PR!

I'm not sure how to approach this PR, it's a really big one and I can't test it locally because I don't have any Windows available. At first glance the code seems ok, even if it add a lot of unsafe code. Can you show us a screenshot? Just to see the result in a graphical point of view. I think you can also add the Windows target inside the .travis.yaml in order to check if everything compile.

@gilescope , do you have any feed-backs after using it?

BTW: Sorry for the slow response, I was busy during the last days.

jamiebrynes7:
I've had a chance to give this a quick test on my Windows machine - will paste a screenshot below. Can verify that it appears to work as expected :+1: 

image

danieldulaney:
Sorry for the delay. Here's it running on my C:\ drive (multiple groups, lots of permissions, graceful error handling)

image

Peltoche:
BTW, @danieldulaney do you want to be part of the active development team? If you want I will add you to the codeowners and you will be able to make the reviews. @meain is already part of it. 

The goal is to avoid the "one man show" project. Every PR should be reviewed by at least 2 reviewers, (even mine) and so I would stop being the only maintainer.

Peltoche:
LGTM, can you add the windows target inside the .travis.yml please?

After that I think we are good for the merge!

gim913:
Awesome, I wanted to try `exa`, but it has some weird deps, that make it *nix only.
Just checked this branch and it  compiles without problems. Hope it's gonna get merged soon.

src/meta/filetype.rs
```diff
@@ -42,6 +45,26 @@ impl FileType {
...
+    #[cfg(windows)]
+    pub fn new(meta: &Metadata, permissions: &Permissions) -> Self {
+        let file_type = meta.file_type();
+
+        if file_type.is_file() {
+            FileType::File {
+                exec: permissions.is_executable(),
+                uid: permissions.setuid,
+            }
+        } else if file_type.is_dir() {
```
humb1t: I would recommend avoiding if\else statements, especially if\elseif. 
In case there would be a new file type, you wouldn't catch it via a compiler, but you could catch it if you would use pattern matching. And you already used 'better' approach here https://github.com/Peltoche/lsd/blob/master/src/meta/mod.rs#L50

src/meta/windows_utils.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    if error_code != winerror::ERROR_SUCCESS {
+        return Err(std::io::Error::from_raw_os_error(error_code as i32))
+    }
+
+    // Assumptions:
+    // - owner_sid_ptr is valid
+    // - group_sid_ptr is valid
+    // (both OK because GetNamedSecurityInfoW returned success)
+    let (owner_name, owner_domain) = unsafe { lookup_account_sid(owner_sid_ptr) }?;
```
pandarrr: Propagating errors in this function will cause `sd_ptr` to be leaked. Safest option is to wrap `sd_ptr` in a struct that calls `LocalFree` in its drop method. You won't have to worry about making sure it's cleaned up in every path.

src/meta/windows_utils.rs
```diff
@@ -0,0 +1,326 @@
...
+    let (group_name, group_domain) = unsafe { lookup_account_sid(group_sid_ptr) }?;
```
pandarrr: Propagating errors in this function will cause `sd_ptr` to be leaked. Safest option is to wrap `sd_ptr` in a struct that calls `LocalFree` in its drop method. You won't have to worry about making sure it's cleaned up in every path.

Peltoche:
As this PR is open since a long time, I merge it inside the `windows` branch so other people can contribute on it. Once everyone will have done its changes and be happy with the feature we will merge it into master.

src/meta/windows_utils.rs
Peltoche: @pandarrr , thanks for the comment. I'm not really confident with all the unsafe stuff. Would it be possible for you to create a PR to the `windows` branch with a possible fix please?

src/meta/filetype.rs
Peltoche: Hi @humb1t , thanks for the comment.

I don't think it's possible because it's a `fs::Metadata::FileType` and not a `lsd::meta::filetype::Filetype` enum like your example.

src/meta/windows_utils.rs
pandarrr: @Peltoche can do. I should have time this week.

PR #125, opened 2019-02-12
switch to non-grid when no tty is available

### LINKED ISSUE #123: lsd crashes when output is not the tty
e.g. when it is a pipe:

```
$ lsd | less
thread 'main' panicked at 'failed to retrieve terminal size', src/display.rs:29:17
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

Similarly when its output is captured in a variable (it is true that one should not parse ls output, but this is a harsh lesson for such an edge case):

```
$ foo=$(lsd)
thread 'main' panicked at 'failed to retrieve terminal size', src/display.rs:29:17
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```
GNU ls would simply activate --oneline mode rather than try to calculate the size of a nonexistent tty.

meain:
fixed #123

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -26,8 +26,8 @@ pub fn one_line(metas: Vec<Meta>, flags: Flags, colors: &Colors, icons: &Icons) 
 
 pub fn grid(metas: Vec<Meta>, flags: Flags, colors: &Colors, icons: &Icons) -> String {
     let term_width = match terminal_size() {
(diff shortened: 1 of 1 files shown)


Peltoche:
Hi @meain ,

Thanks for the PR. This is not the first time this bug appened, I think we had some regression during the refactoring.

PR #138, opened 2019-03-04
Fix the confinement rule for the snapcraft package

### LINKED ISSUE #79: snap version gets "Permission denied"
This is on ubuntu 18.04.1

## Expected behavior
lsd to give a listing

## Actual behavior
```bash
$ lsd
cannot open directory'.': Permission denied (os error 13)
```

Peltoche:
Fix #79

### DIFF
--- a/.snapcraft.yaml
+++ b/.snapcraft.yaml
@@ -6,7 +6,7 @@ description: lsd is a ls command including pretty icons based on nerd-font.
 architectures:
   - amd64
 grade: stable
-confinement: classic
+confinement: strict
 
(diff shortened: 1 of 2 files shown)


wrapperband:
Just reporting, 18.04 still has the (--classic) install problem : reports version as :    lsd 0.13.0

Not up on how snaps are updated, i.e. fix may still be in flight ...

kumavis:
per #79 seems like `classic` confinement is correct, unless you can correctly set the permissions for `strict`

Issue #141, opened 2019-03-06
Relative date should print "now" instead of "now ago"


hongleong:
## Expected behavior

The relative date-time for current time should be "now" instead of "now ago".

```
$ lsd -la --date=relative .bash_history
.rw-------  user  user  783.8 KB  now  .bash_history
```

## Actual behavior

```
$ lsd -la --date=relative .bash_history
.rw-------  user  user  783.8 KB  now ago  .bash_history
```

The text "ago" should be removed in the output above.

Peltoche:
Hi @hongleong ,

Thanks for the report. 

It would make it easier to read indeed but I'm not sure that we have a hand on this at the moment because this format is handled by the [chrono-humanize](https://docs.rs/chrono-humanize/0.0.11/chrono_humanize/struct.HumanTime.html) package. 

I see two possibilities to fix this:
- Patch the chrono_humanize lib
- Parse handle the case in our code by matching and replacing the result.

I think we can start with the second solution and then propose a PR to chrono.

Do you want to make it?

PR #142, opened 2019-03-06
Do not pass tense explicitly

meain:
Fixes https://github.com/Peltoche/lsd/issues/141


meain:
Sure, I have added that.

PR #143, opened 2019-03-06
Fix wording

jneidel:
which makes it really faster -> which makes it much faster


Peltoche:
Hi @jneidel ,

Thanks for the PR! It's much appreciated because I'm not an native english speaker and I guess there is a lot of typos.

PR #144, opened 2019-03-07
Update windows utils

pandarrr:
- Added struct to maintain lifetime of pointers
- Moved logic that operates on pointers into struct

Side note: I worked on this using the x86_64-pc-windows-msvc target and it builds fine.


Peltoche:
Hi @pandarrr ,

Thanks for the PR!

PR #151, opened 2019-03-11
Handle all the color code format for the visible width calculation

### LINKED ISSUE #145: Weird grid alignment

Peltoche:
Should fix #145  and #149

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -278,21 +278,16 @@ fn get_long_output(
 fn get_visible_width(input: &str) -> usize {
     let mut nb_invisible_char = 0;
 
-    for (idx, _) in input.match_indices("\u{1b}[38;5;" /* "\e[38;5;" */) {
-        let color_code = input.chars().skip(idx + 7);
-        let mut code_size = 0;
-        color_code
-            .skip_while(|x| {
-                code_size += 1;
-                char::is_numeric(*x)
-            })
-            .count();
-        nb_invisible_char += 6 + code_size; /* "\e[38;5;" + color number + "m" */
-    }
-
-    if nb_invisible_char > 0 {
-        // If no color have been set, the is no reset character.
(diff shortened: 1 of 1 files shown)


meain:
Looks like all the tests are failing, not just the first one.

src/display.rs
```diff
@@ -278,21 +278,16 @@ fn get_long_output(
...
-    }
-
-    if nb_invisible_char > 0 {
-        // If no color have been set, the is no reset character.
-        nb_invisible_char += 3; /* "[0m" */
+    // Search for every instance of "\e[" indicating the start of a color code
+    // then count the number of charactere between the string and the first 'm'
+    // charactere indicating the end of the code color.
+    for (idx, _) in input.match_indices("\u{1b}[") {
+        for c in input.chars().skip(idx) {
```
meain: I feel like using skip will have issues with non unicode ones.
Unicode length might be more than just simple length.

meain:
@Peltoche  Something like this works:

```rust
    let matches: Vec<_> = input.match_indices("\u{1b}[").collect();
    if matches.len() > 0 {
        for c in input.chars().skip(matches[0].0) {
            nb_invisible_char += 1;
            println!("{:?} {}", c, nb_invisible_char);
            if c == 'm' {
                break;
            }
        }
        if nb_invisible_char > 0 {
            nb_invisible_char += 2;
        }
    }
```

meain:
I have made all my changes and made it as a pr to this branch on #152

PR #154, opened 2019-03-11
Add "manual" clap alias for -A

### LINKED ISSUE #147: Add the `-A` flag for compatibility purpose
(From #136 )

As of right now -a is actually doing the job of -A. Maybe we should just have -A as a alias for -a for now until we get a fix for #122

matthiasbeyer:
This is requested as a feature, but because clap does not support
"short" aliases, we have to create another clap argument for this.

Not the best possible way to implement this (because the best way would
be to patch clap itself to allow short aliases) but the fastest one.

---

Closes #147

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -13,6 +13,12 @@ pub fn build() -> App<'static, 'static> {
                 .help("Do not ignore entries starting with ."),
         )
         .arg(
+            Arg::with_name("almost-all")
+                .short("A")
+                .multiple(true)
+                .help("Alias of -a"),
+        )
+        .arg(
             Arg::with_name("color")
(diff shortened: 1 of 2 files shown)


src/app.rs
```diff
@@ -12,6 +12,12 @@ pub fn build() -> App<'static, 'static> {
                 .multiple(true)
                 .help("Do not ignore entries starting with ."),
         )
+        .arg(
+            Arg::with_name("all_big")
```
meain: Maybe name it as `almost-all`. That is what `ls` calls it.

matthiasbeyer:
Here you go. Sorry, I was at work the whole day.

PR #155, opened 2019-03-11
Build hashmaps at compiletime

matthiasbeyer:
This patch alters the icons code so that the big hashmaps are build at
compiletime using the 'phf' crate.

I did not performance-test this patch, although building these hashmaps
at compiletime should at least theoretically improve execution time.

---

It was fun to write this patch. If you want to apply it, I'd rather speed-test the difference before blindly applying it.

Unfortunately, I do not have the computing power to do meaningful speed tests right now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17,6 +17,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "autocfg"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "bitflags"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -53,6 +58,19 @@ dependencies = [
  "unicode-width 0.1.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "vec_map 0.8.1 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
+
+[[package]]
+name = "cloudabi"
+version = "0.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "bitflags 1.0.4 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "fuchsia-cprng"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "fuchsia-zircon"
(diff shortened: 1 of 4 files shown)


Peltoche:
Hi @matthiasbeyer ,

Thanks for this PR! If you want to make some fine grained traces to test the bottlenecks you can try [hawktracer](https://github.com/AlexEne/rust_hawktracer). I did some tests with it and the main bottleneck I found was curiously not the map but the creation of the string for for permissions (`meta::permission::Permissions.render` to be exact). You can also try with hyperfine, like the benchmarks inside the readme.

As the goal of this PR is to improve the perfs, I think we should take the time to make thorough performances tests in order to be sure that it will really improve them.

Peltoche:
My tests with hyperfine gives:
``` ~  Code  Perso  lsd  master  1+  %  ⎈ prodild  prod   hyperfine --warmup 10 './target/release/lsd -la /etc/*' '/tmp/lsd/target/release/lsd -la /etc/*'
Benchmark #1: ./target/release/lsd -la /etc/*
  Time (mean ± σ):      13.6 ms ±   0.9 ms    [User: 6.3 ms, System: 7.3 ms]
  Range (min … max):    12.2 ms …  17.5 ms    184 runs

Benchmark #2: /tmp/lsd/target/release/lsd -la /etc/*
  Time (mean ± σ):      13.6 ms ±   0.8 ms    [User: 5.8 ms, System: 7.8 ms]
  Range (min … max):    12.0 ms …  16.9 ms    186 runs

Summary
  '/tmp/lsd/target/release/lsd -la /etc/*' ran
    1.00 ± 0.09 times faster than './target/release/lsd -la /etc/*'
```

It seems that you solution improve a little the perfs indeed :)

matthiasbeyer:
I would love to do a more troughout testing/perfing before merging this. Not only because I never did perf tests before, but also because I'm curious how much I can improve this.

Though you have to know that I have no internet connection at home for my workstation machine, so I cannot do this _right now_ because I do not have the resources. I would love to continue playing with this, but it might take some time before I can say something. I would wait with merging, though.

So if you have time (in the range of weeks rather than days), ...

Peltoche:
> but it might take some time before I can say something

No worries, we have time. We can keep this PR open as long as we want.

matthiasbeyer:
Why are you pushing to my branch?

Peltoche:
> Why are you pushing to my branch?

I had rebase the branch with master via github because I was thinking to merge the branch. I like to keep the PR on top of the master in order to detect quickly the merge conflics.

If this bother you, you can override it and I will not keep you PR update to date with master anymore.

PR #152, opened 2019-03-11
fix spacing issue on colors

meain:

PR #157, opened 2019-03-12
Refactor width computation for colored strings

meain:
Has some tiny perf improvement as well, I guess.

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -278,18 +278,12 @@ fn get_long_output(
 fn get_visible_width(input: &str) -> usize {
     let mut nb_invisible_char = 0;
 
-    // Search for every instance of "\e[" indicating the start of a color code
-    // then count the number of charactere between the string and the first 'm'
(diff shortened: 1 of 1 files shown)


meain:
Oh, in that case I think we will have to look into the current implementation as well. In https://github.com/Peltoche/lsd/pull/152 I was assuming it has just one color string and it is only checking the first one.

The reason for this was that for non ascii names, the index returned from `match_indices` sometimes is greater than the total length.

Peltoche:
Rectification: You solution is correct

We can have only the simple output format with the grid layout and so we can have only one color at the moment.

```rust
fn get_short_output(meta: &Meta, colors: &Colors, icons: &Icons, flags: Flags) -> String {
    let strings: &[ANSIString] = &[
        meta.name.render(colors, icons),
        meta.indicator.render(flags),
    ];

    ANSIStrings(strings).to_string()
}
```

PR #165, opened 2019-03-22
show current and parent directory

hemreari:
This might solve issue#112 about showing current and parent directory.


src/meta/mod.rs
```diff
@@ -56,8 +56,14 @@ impl Meta {
             println!("cannot access '{}': {}", path.display(), err);
             return Ok(meta);
         }
+		let mut content = Vec::new();
+
+		let current_meta = Self::from_path(&PathBuf::from("."))?;
```
meain: We should probably add the path that is received as just checking for `.` will not take into consideration the path received 

as in if someone gives

```
lsd ~/Documents
```

we should be providing the `.` and `..` relative to `~/Documents` and not the folder the user in.

src/meta/mod.rs
```diff
@@ -56,8 +56,14 @@ impl Meta {
...
+		let parent_meta = Self::from_path(&PathBuf::from(".."))?;
+
+		content.push(current_meta);
```
meain: We should check for the flag `-a` ie `-all` to decide if we should be showing `.` and `..`

meain: also maybe the creation and insert into a single line and push based on the condition.

hemreari:
> I have suggested a few changes in the review, could you look into it?

Of course I can look into. I will review in 1-2 days.

hemreari:
I figure out the problem from the last commit. Now, I changed `Mod.Name` structs' `name` field to `pub`:
https://github.com/hemreari/lsd/blob/master/src/meta/name.rs#L9

So that I can change current and parent folder's name to `.` and `..` in where we get paths:
https://github.com/hemreari/lsd/blob/master/src/meta/mod.rs#L71-L74
If I don't do that, output is 
image 
instead of this 
image
(note: my `~/Documents` folder is empty.)


But, If you don't want to change name field to pub, I added two new functions under `mod.rs` which is `from_current_path` and `from_parent_path`. These functions seem like to me repeater of original `from_path` function so I commented them:
https://github.com/hemreari/lsd/blob/master/src/meta/mod.rs#L147-L195

If you haven't like both ways, I can think something else :smiley: 
Let me know what you think.

src/meta/mod.rs
```diff
@@ -56,8 +57,46 @@ impl Meta {
...
+		let mut content = Vec::new();
+
+		if list_hidden_files {
+			let mut current_meta;
+			let mut parent_meta;
+
+			let path_str = path.to_str().unwrap();
+
+			if path_str == ".." {
+				let temp_path = fs::canonicalize(&PathBuf::from("..")).unwrap();
```
meain: I think we might have to move this to `core.rs` line 70. Maybe something like 

```rust
match Meta::from_path_recursive(&fs::canonicalize(&path.to_path_buf()).unwrap(), depth, self.flags.display_all) {}
```

The issue is that we will not be handling for paths like `./../`. 

Once that is done, I think we should be able to completely remove the if else check here as we will always get the absolute path.

> There might also be other place in which we might be doing this check which will have to be taken care of later

src/meta/mod.rs
```diff
@@ -56,8 +57,46 @@ impl Meta {
             println!("cannot access '{}': {}", path.display(), err);
             return Ok(meta);
         }
+		let mut content = Vec::new();
+
+		if list_hidden_files {
```
meain: There needs to be some change to this as the check has to be `all` (a flag that will have to be added in). 

Right now `list_hidden_files` is set to true if `-a` or `-A` is passed as a flag. It should be just `-a`.
The `list_hidden_files` variable comes from `flags.rs` line 68. 

The current one will show `.` and `..` with both `all` and `almost-all`. We need it to be showing when  it is just `all`. You might have to change it there. This was a  placeholder which was set there just until we had this feature added. You might have to add another flag into the mix.

hemreari: > You might have to add another flag into the mix.

Did you mean we have to add a new flag which is 'almost-all' by the 'add another flag into the mix'?

Currently I remove `matches.is_present("almost-all")` on line 68.

meain: Right now the condition you are checking now `list_hidden_files` is set to true for both `-a` and `-A` flag, but we should be showing the current and parent directory only if `-a` is used not when `-A` flag is used.

What I meant was that you will have to add a new flag in `flags.rs` to separate them and pass the it to your code.

hemreari:
I added `almost-all` flag, no conflict between original ls as far as I know.

hemreari:
> One small addition.
> 
> In `app.rs` could you add `.overrides_with("display_all")` to the `-A` flag and `.overrides_with("display_almost_all")` to `-a`.
> 
> Once that is done I think we are pretty much good to go.

Done :+1:

meain:
@hemreari  Woops, my bad. I god the names wrong. I meant this

```rust
        .arg(
            Arg::with_name("all")
                .short("a")
                .overrides_with("almost-all")
                .long("all")
                .multiple(true)
                .help("Do not ignore entries starting with ."),
        )
        .arg(
            Arg::with_name("almost-all")
                .short("A")
                .overrides_with("all")
                .long("almost-all")
                .multiple(true)
                .help("Do not list implied . and .."),
        )
```

hemreari:
Okay, fixed.

src/flags.rs
```diff
@@ -3,6 +3,7 @@ use clap::{ArgMatches, Error, ErrorKind};
 #[derive(Clone, Debug, Copy)]
 pub struct Flags {
     pub display_all: bool,
+    pub display_almost_all: bool,
```
Peltoche: As it should have only one of the option display_all/display_almost_all/display_only_visible active I think it would be a good idea to refactor a bit and create a "Display" enum or something like this. 

This can be made in a future PR.

What do you think of it?

Peltoche:
Sorry for the late review,

I just see a little improvement possible. Except that, LGTM

src/flags.rs
meain: I  guess that will be a good addition, will reduce some confusion in the check in `mod.rs` line 84.

@hemreari , could you make that change too?

hemreari: > @hemreari , could you make that change too?

of course

hemreari:
Added display enum. Let me know what you think.

src/meta/mod.rs
```diff
@@ -56,25 +57,51 @@ impl Meta {
             println!("cannot access '{}': {}", path.display(), err);
             return Ok(meta);
         }
-
         let mut content = Vec::new();
+
+         match display {
```
Peltoche: You can remplace the match by something like:
```rust
if let Display::DisplayAll = display {
```

src/meta/mod.rs
```diff
@@ -56,25 +57,51 @@ impl Meta {
...
-            if !list_hidden_files
-                && path
-                    .file_name()
-                    .ok_or_else(|| Error::new(ErrorKind::InvalidInput, "invalid file name"))?
-                    .to_string_lossy()
-                    .starts_with('.')
-            {
-                continue;
-            }
+           match display {
```
Peltoche: same here

src/meta/mod.rs
```diff
@@ -56,25 +57,51 @@ impl Meta {
...
+
+                let parent_path = match path.parent() {
+                    None => PathBuf::from("/"),
+                    Some(path) => PathBuf::from(path),
+                };
+
+                current_meta = Self::from_path(&PathBuf::from(path))?;
+                current_meta.name.name = ".".to_string();
+
+                parent_meta = Self::from_path(&PathBuf::from(parent_path))?;
```
Peltoche: The linter says that you can remove the `PathBuf::from(` for this line

hemreari: I changed match to if.

Which linter program do you use? I use emacs's racer, it don't give such suggestions.

Peltoche: you can check the specific command into the .travis.yml file.

PR #166, opened 2019-03-22
sizesort

hjanuschka:
adds `--sizesort` and `-S`  to sort by Filesize.

changes: displays sizes for folders too, they have a size.

`ls -lShr` is now compatible:

# `ls -lShr`

Bildschirmfoto 2019-03-22 um 14 56 59

# `lsd -lShr` 💊 

Bildschirmfoto 2019-03-22 um 14 58 33


src/meta/size.rs
```diff
@@ -15,49 +15,50 @@ pub enum Unit {
...
     value: u64,
     unit: Unit,
+    bytes: u64,
 }
 
 impl<'a> From<&'a Metadata> for Size {
     fn from(meta: &Metadata) -> Self {
         let len = meta.len();
-
-        if meta.is_file() {
```
meain: Is there a reason why you removed this check?

src/sort.rs
```diff
@@ -13,7 +13,11 @@ pub fn by_meta(a: &Meta, b: &Meta, flags: Flags) -> Ordering {
             DirOrderFlag::None => by_name(a, b, flags),
             DirOrderFlag::Last => by_name_with_files_first(a, b, flags),
         },
-
+        SortFlag::Size => match flags.directory_order {
+            DirOrderFlag::First => by_size(a, b, flags),
```
meain: Could you also take into account the `DirOrderFlag`.

src/meta/size.rs
hjanuschka: it didnt' return a size for directories. and therefore it produced some weird result.
directories have a size, depending how many files are in them.
returning the size, produced the same result as one know's from `ls`

src/sort.rs
hjanuschka: i don't get the idea of this, i tried todo it like the `name`  sorting, but it produced, to me, unlogic results.

Peltoche: IMHO it would be curious to take into account the `DirOrderFlag` into the size sort. If the change doesn't bring any major improvement we should try to stick as much as possible to the classic `ls`.

Peltoche:
LGTM

@meain, as you have done some objections I let you decide when this PR can be merged. Fell free to continue the conversations if needed.

src/sort.rs
meain: I was thinking something like this
Imgur

as ls has this
Imgur


The idea is that we will be separately sorting the dirs and files and depending on the flag we will have dirs before or after files

meain: sort with dirs first will look something like this
```rust
fn by_size_with_dirs_first(a: &Meta, b: &Meta, flags: Flags) -> Ordering {
    match (a.file_type, b.file_type) {
        (FileType::Directory { .. }, FileType::Directory { .. }) => by_size(a, b, flags),
        (FileType::Directory { .. }, _) => Ordering::Less,
        (_, FileType::Directory { .. }) => Ordering::Greater,
        _ => by_size(a, b, flags),
    }
}
```

Peltoche: > as ls has this

I just learn something new

borats:
when can we expect this to land in master?

meain:
@Peltoche Since we have not heard back from @hjanuschka , shall we merge this in and handle the `DirOrder` flag in a later commit.

Peltoche:
@meain  I was about to write the exact same thing. Totally agree.

PR #171, opened 2019-03-27
Stick to user colors for permissions.

meain:
This will make sure that the colors will not be off from their setup. Also changed `NoAccess` and `NonFile` to a grey color.

Comparison:
comparison


meain:
I also wanted to know your opinion on changing the  padding between the different blocs to one space than 2 spaces.

It will look something like this

Peltoche:
> @Peltoche I also wanted to know your opinion on changing the padding between the different blocks to one space than 2 spaces. If you are cool with that I will add that into this PR.

Hi! I don't really have any opinion about it. It will make the output a little more compact which seems to be a good idea so let's go.

meain:
@Peltoche I have pushed the changes to change it to one space.

PR #172, opened 2019-03-27
bugfix: only one value for multi value options in clap

meain:
before `lsd --icon never ../` would give an error like

error


Peltoche:
Hi!
Thanks for the PR!

PR #175, opened 2019-03-29
Add a new way to display size (smaller)

meain:
What do you  think of an option like this?

### DIFF
--- a/src/flags.rs
+++ b/src/flags.rs
@@ -9,6 +9,7 @@ pub struct Flags {
     pub sort_by: SortFlag,
     pub sort_order: SortOrder,
     pub directory_order: DirOrderFlag,
+    pub size: SizeFlag,
     pub date: DateFlag,
     pub color: WhenFlag,
     pub icon: WhenFlag,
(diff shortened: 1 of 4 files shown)


src/flags.rs
```diff
@@ -117,6 +121,22 @@ impl Default for Flags {
...
+#[derive(Clone, Debug, Copy, PartialEq, Eq)]
+pub enum SizeFlag {
+    Default,
+    Small,
+}
+
+impl<'a> From<&'a str> for SizeFlag {
+    fn from(size: &'a str) -> Self {
+        match size {
+            "size" => SizeFlag::Default,
```
Peltoche: What do you think of replacing "size" by something more explicit like "classic" or even "default"?

Peltoche:
Good idea! This can be a good start for proposing a "compressed mode" used for some layout like the `exa -lG`.

src/flags.rs
meain: I think we will use `default` and maybe we could display raw number of bytes for `classic`?

meain:
@Peltoche I know this got merged but I feel like `short` might be a better flag than `small`. What do you think? Should we change it to short?

Peltoche:
> @Peltoche I know this got merged but I feel like `short` might be a better flag than `small`. What do you think? Should we change it to short?

It's true that "short" seems semantically more correct to me. As the feature have not been released yet, we can change the name without problems so let's go for it.

PR #178, opened 2019-03-30
Add --blocks flag as per #96

### LINKED ISSUE #96: Feature request: Choose blocks to display in long format
Add a feature which would let users choose the blocks to be displayed. The blocks could be a comma separated list.

Maybe something like:

```sh
lsd --blocks permission,size,name
```

This would just show these blocks. Maybe even show them in that specific order as well.

meain:
Closes #96 

Changes:
- Removed `Clone` derive on `Flags` struct 

TODO:

- [x] fix spacing of name

-  [ ] figure out a way to make second block argument override the the first one
`lsd -l --blocks name --blocks user,date` should be equal to `lsd -l --blocks user,date`. Currently it will result in `lsd -l --blocks name,user,date`. This helps to have a default in the `ls` alias.

Doubts:

- [x] seperately show `symlink` (indicator + symlink)? what to call it?

### Possible future addition

Add alignment direction. `:name` will left align. `name:` will right align. `:name:` will center align.
I took the idea from markdown table formatting.

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -58,7 +58,7 @@ fn inner_display_one_line(
             size: detect_size_lengths(&metas, &flags),
             date: detect_date_length(&metas, &flags),
             name: detect_name_length(&metas, &icons),
-            name_with_symlink: detect_name_with_symlink_length(&metas, &icons)
+            name_with_symlink: detect_name_with_symlink_length(&metas, &icons),
         })
     }
 
@@ -283,8 +283,10 @@ fn get_long_output(
     let mut strings: Vec<ANSIString> = Vec::new();
     for block in flags.blocks.iter() {
         match block {
-            Block::FileType => strings.push(meta.file_type.render(colors)),
-            Block::Permission => strings.push(meta.permissions.render(colors)),
+            Block::Permission => {
+                strings.push(meta.file_type.render(colors));
+                strings.push(meta.permissions.render(colors));
+            }
             Block::User => strings.push(meta.owner.render_user(colors, padding_rules.user)),
             Block::Group => strings.push(meta.owner.render_group(colors, padding_rules.group)),
             Block::Size => strings.push(meta.size.render(
--- a/src/display.rs
+++ b/src/display.rs
@@ -18,6 +18,7 @@ struct PaddingRules {
     group: usize,
     size: (usize, usize),
     date: usize,
+    name: usize,
 }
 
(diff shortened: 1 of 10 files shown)


meain:
Hi @Peltoche ,

This looks almost done. Any suggestions?

One thing I would like to know your opinion is about the `namewithoutsymlink` if any.

src/display.rs
```diff
@@ -251,28 +255,50 @@ fn get_long_output(
...
+                        colors,
+                        icons,
+                        Some(padding_rules.name_with_symlink + 3),
+                    ));
+                    strings.push(meta.indicator.render(&flags));
+                    strings.push(meta.symlink.render(colors, None));
+                }
+            },
+        };
+        strings.push(ANSIString::from(" ")); // TODO do not add this space to the end
```
Peltoche: This is still a TODO?

src/display.rs
```diff
@@ -251,28 +255,50 @@ fn get_long_output(
...
+                    ));
+                }
+                None => {
+                    strings.push(meta.name.render(
+                        colors,
+                        icons,
+                        Some(padding_rules.name_with_symlink + 3),
+                    ));
+                    strings.push(meta.indicator.render(&flags));
+                    strings.push(meta.symlink.render(colors, None));
```
Peltoche: I don't think you need to add it in this condition. It should always return an empty string.

Peltoche:
You proposition from https://github.com/Peltoche/lsd/issues/193#issuecomment-489319948 seems to be a very promising solution, it would be way more flexible.

src/display.rs
meain: Oh yeah, it still is.

src/display.rs
meain: I am not sure what you are referring to.

meain:
Yeah, I will add the `--no-symlink` flag.

src/display.rs
meain: I think I figured out what you where talking about. I have made it to not do padding inside.

Peltoche:
LGTM except the `TODO`. I'm not a big fan of keeping `TODO ` stuff outside a PR.

After I think we are good to go

meain:
Yeah sure, I will fix that. Just did not get around it.

meain:
@Peltoche Finally got around to completing this PR.

PR #182, opened 2019-04-06
Make the --tree flag respect the --group-dirs value

### LINKED ISSUE #104: tree display result not correct
```
$ cd lsd/src
$ lsd --tree
```

## Expected behavior
```
...
├── icon.rs
├── main.rs
└── meta
    ├── date.rs
    ├── filetype.rs
    ├── indicator.rs
    ├── mod.rs
    ├── name.rs
    ├── owner.rs
    ├── permissions.rs
    ├── size.rs
    └── symlink.rs
```

## Actual behavior

`meta` does not show as the parent of its children.

```
│  ├── date.rs
│  ├── filetype.rs
│  ├── indicator.rs
│  ├── mod.rs
│  ├── name.rs
│  ├── owner.rs
│  ├── permissions.rs
│  ├── size.rs
│  └── symlink.rs
├── meta
├── app.rs
├── batch.rs
...
```

### LINKED ISSUE #158: Display tree in alphabetical order
Is it possible to add a flag to sort alphabetically when using the tree feature? Currently files are first shown and only then inner directories. I would like to replicate `tree` default behaviour:

Screenshot_2019-03-12_14:37:13

I thought that maybe `lsd --tree --group-dirs=none` would achieve such behaviour but it didn't.

JD557:
Fix #158 

The fix is a bit weird though... It seems that this was a special case added on 0fa68098e2af46420a549f4a5d092dc882d1ecc3

I did some small tests and everything appears to be working though. Maybe this was just an oversight?

### DIFF
--- a/src/sort.rs
+++ b/src/sort.rs
@@ -1,12 +1,8 @@
-use crate::flags::{DirOrderFlag, Flags, Layout, SortFlag, SortOrder};
+use crate::flags::{DirOrderFlag, Flags, SortFlag, SortOrder};
 use crate::meta::{FileType, Meta};
 use std::cmp::Ordering;
 
 pub fn by_meta(a: &Meta, b: &Meta, flags: Flags) -> Ordering {
-    if flags.layout == Layout::Tree {
-        return by_name_with_files_first(a, b, flags);
-    }
-
     match flags.sort_by {
         SortFlag::Name => match flags.directory_order {
             DirOrderFlag::First => by_name_with_dirs_first(a, b, flags),


Peltoche:
> [](https://github.com/Peltoche/lsd/issues/158)
> 
> The fix is a bit weird though... It seems that this was a special case added on [0fa6809](https://github.com/Peltoche/lsd/commit/0fa68098e2af46420a549f4a5d092dc882d1ecc3)
> 
> I did some small tests and everything appears to be working though. Maybe this was just an oversight?

Hi @JD557 , thanks for the PR!

The first implementation of `--tree` was... "not perfect" and for what I remember, adding files after the folder was not possible. The display logic have been rewritten since and I would not be surprised if this is not more a requirement.

PR #183, opened 2019-04-07
Windows

Peltoche:


Peltoche:
I was hoping for some more feedback but I guess the best way to have them is to send it into production.

So I merge the branch into master and let's see what's append

PR #189, opened 2019-04-18
Sort files without considering leading dot

### LINKED ISSUE #187: Sort dot files without the leading dot
A directory with .b and a could be shown with lsd -A as:
a .b
instead of the current
.b a

bofh69:
This should fix #187

### DIFF
--- a/src/meta/name.rs
+++ b/src/meta/name.rs
@@ -74,7 +74,18 @@ impl Name {
 
 impl Ord for Name {
     fn cmp(&self, other: &Self) -> Ordering {
-        self.name.to_lowercase().cmp(&other.name.to_lowercase())
+        // Compare names without considering leading dot
+        let mut a = self.name.as_str();
+        if a.len() > 1 && a.starts_with(".") {
+            let (_, a_) = a.split_at(1);
+            a = &a_;
+        } 
+        let mut b = other.name.as_str();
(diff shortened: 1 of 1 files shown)


meain:
Hey @bofh69 , we should probably be sticking to the current format as it is what `ls` seems to be doing so that we can keep the behaviour somewhat similar.

bofh69:
Hi @meain

Different ls implementations does different things here. I just tried ls on UNIX v7 on a PDP11/70 and it works like you say, but GNU ls does:
```
$ ls --version
ls (GNU coreutils) 8.28
[...]

$ mkdir tmp; cd tmp; touch a .b; mkdir c .d; ls -a; ls -A
.  ..  a  .b  c  .d
a  .b  c  .d
```

Peltoche:
> GNU ls uses the locale for its sorting order:

Interesting. This can be indeed seen by executing the following commands: 
```
 LC_ALL=en_US.UTF-8 sort <<< $'a\nb\nA\nB\n.b\n..b\n'
 LC_ALL=C sort <<< $'a\nb\nA\nB\n.b\n'
```
The GNU ls seems to have the exact same behavior. 

In any case if we want to make the same behavior your PR will not solve this permanently. We need to find a sort which take into account the locale.

PR #196, opened 2019-05-08
Add -d/--directory-only option.

### LINKED ISSUE #190: No way to list directory itself rather than its contents
## Expected behavior

In "regular" `ls`, there is a `-d` flag:

```
       -d, --directory
              list directories themselves, not their contents
```

I often find this useful when checking ownership and permissions. I suspect it might be impossible to show file metadata for the root directory with `lsd` without this flag, i.e. `lsd -l /` shows the contents of `/` not the directory itself.

Here's the behaviour I would expect:

```
$ lsd -ld src
drwxr-xr-x  gavin  staff       -   Tue Apr 23 16:17:38 2019    src
```

Here's the behaviour of `/bin/ls` on Ubuntu 18.04:

```
$ /bin/ls -ld /
drwxr-xr-x 24 root root 4096 Apr  3 17:53 /
```

## Actual behavior

```
$ lsd -ld src
error: Found argument '-d' which wasn't expected, or isn't valid in this context

USAGE:
    lsd --long

For more information try --help
```

allenap:
Fixes #190, and:
- Moves the decision about which _metas_ to display out of the `display` module.
- Re-uses or clones `Meta` structs when possible instead of using `Meta::from_path` which incurs a system call or two.
- Avoids calling `read_dir()` twice for the same path.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -103,6 +103,15 @@ pub fn build() -> App<'static, 'static> {
                 .help("Stop recursing into directories after reaching specified depth"),
         )
         .arg(
+            Arg::with_name("directory_only")
+                .short("d")
+                .long("directory-only")
+                .conflicts_with("recursive")
+                .conflicts_with("tree")
+                .conflicts_with("depth")
+                .help("Display directories themselves, and not their contents"),
+        )
+        .arg(
             Arg::with_name("size")
                 .long("size")
                 .possible_value("default")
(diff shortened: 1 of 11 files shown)


allenap:
I'm really happy to pair-review this live with someone, if that would help. It's a larger change than you might have expected for this feature, but I think/hope I can justify it :-) Some of the diff is change created by `rustfmt` – I can unwind that too if you would prefer.

meain:
Hey @allenap , sorry that it took a little while to review the PR.

Looks good to me, could you rebase it with the master branch?

allenap:
> Looks good to me, could you rebase it with the master branch?

@meain Rebased and pushed, with some additional updates from `println!` to `eprintln!`. Thanks!

Issue #204, opened 2019-05-21
--long and --tree are mutually exclusive


windware-ono:
version 0.14.0

## Expected behavior
It should show file infromation from --long option under --tree as well.

## Actual behavior
Behaves as if --long was not specified.

Issue #185, opened 2019-04-12
Show symlink target with --tree option


sandorex:
Add flag to show symlink target with tree option

sandorex:
Im such a noob, if somebody could remove the bug label its not a bug...

Peltoche:
Hi @sandorex ,

Thanks for the issue. This would be a great idea indeed.

meain:
Hey @sandorex . PR #178  and PR #226 combined will help you to help you get what you need.

You can use something like `lsd --tree --depth 1 -l --blocks name` to get the desired output.

I don't think we have a release yet which has both the PR's but we should soon have one.
@Peltoche

PR #226, opened 2019-05-25
flags: allow to mix --long and --tree

Monkeypac:
This is a feature that I'd love to have in `lsd`. 

I just recently discovered `lsd` ~a few weeks ago.
I'm not a Rust developer, this is the first time I do some so feel free to criticize, I just want to suggest the solution.

Refs: #185
Closes: #204

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -64,8 +64,13 @@ impl Core {
 
     fn fetch(&self, paths: Vec<PathBuf>) -> Vec<Meta> {
         let mut meta_list = Vec::with_capacity(paths.len());
+        let tree_layout_true = Layout::Tree { long: true };
+        let tree_layout_false = Layout::Tree { long: false };
 
-        let depth = if self.flags.recursive || self.flags.layout == Layout::Tree {
+        let depth = if self.flags.recursive
+            || self.flags.layout == tree_layout_true
+            || self.flags.layout == tree_layout_false
+        {
             self.flags.recursion_depth
         } else {
             1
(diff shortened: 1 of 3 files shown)


src/core.rs
```diff
@@ -64,8 +64,13 @@ impl Core {
 
     fn fetch(&self, paths: Vec<PathBuf>) -> Vec<Meta> {
         let mut meta_list = Vec::with_capacity(paths.len());
+        let tree_layout_true = Layout::Tree { long: true };
+        let tree_layout_false = Layout::Tree { long: false };
 
-        let depth = if self.flags.recursive || self.flags.layout == Layout::Tree {
+        let depth = if self.flags.recursive
```
meain: I guess we could switch it out to this.

```rust
        let depth = match &self.flags.layout {
            &Layout::Tree { long } if long || !long || self.flags.recursive => {
                self.flags.recursion_depth
            }
            _ => 1,
        };
```

Monkeypac: Thx, I was looking for a better solution but couldn't get it ahah.

meain:
Hmm, looks like clippy is complaining about that code. This should be good.

```rust
        let depth = match self.flags.layout {
            Layout::Tree { .. } => self.flags.recursion_depth,
            _ if self.flags.recursive => self.flags.recursion_depth,
            _ => 1,
        };
```

Monkeypac:
oh yeah I didn't see your answer, thanks ahah

PR #232, opened 2019-06-07
remove gometalinter.json

Keylor42:
**Summary:**
Remove the .gometalinter.json file that is included at the root of the repository since the project is in Rust, not in Go.


Peltoche:
WUT??! How it finished here??! I guess that's my fault...

Thanks @Keylor42 !

Issue #180, opened 2019-04-04
Support multiple sorting options (e.g. size)


Philipp-M:
It would be nice to have support for something like exas `--sort` feature, which I often use.
exa supports the following:
`name, Name, size, extension, Extension, modified, accessed, created, inode, type, none`
I most often use size.
A killer feature would be to also include a sort size option that recursively includes the size of directories, like `du`, though this should probably be additional, as it could be quite slow (like --tree)

Peltoche:
Hi @Philipp-M,

Thanks for the report! It seems to be a good feature indeed and it shouldn't be too hard to implement it. Sadly at the moment I don't have much time for writing new features. 

If you want to make a PR it would be very welcomed :)

jhaungs:
There seems to be code in here to handle sorting by size, but for some reason the -S option is disallowed. I'm just learning rust, so I can sort of read it, but I had no luck understanding how the flag parsing works.

meain:
We have most of these now ;) Anyone landing here, feel free to open and issue for any specific sort method.

PR #235, opened 2019-06-16
Show total size of directories and a new flag for displaying the size (bytes)

Philipp-M:
This partially implements #180 (the `du`-like feature)
It provides a new `--total-size` flag that shows the total size of directories and its contents (recursively) and a new option `bytes` for `--size` which acts similar to the default `ls` size display.

At first I thought it might be necessary to refactor 'Size' to implement this feature. In hindsight it isn't necessary anymore, but I think it's cleaner (and probably faster) now anyway (especially 'render_value').
Any further suggestions like a short flag for `--total-size`?

### DIFF
--- a/src/meta/size.rs
+++ b/src/meta/size.rs
@@ -14,15 +14,13 @@ pub enum Unit {
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Size {
-    value: u64,
-    unit: Unit,
     bytes: u64,
 }
 
 impl<'a> From<&'a Metadata> for Size {
     fn from(meta: &Metadata) -> Self {
         let len = meta.len();
-        Self::from_bytes(len)
+        Self { bytes: len }
     }
 }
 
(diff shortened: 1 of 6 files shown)


src/meta/size.rs
```diff
@@ -73,79 +57,68 @@ impl Size {
...
-        self.paint(colors, content)
+        self.paint(&unit, colors, content)
     }
 
-    fn paint(&self, colors: &Colors, content: String) -> ColoredString {
-        if self.unit == Unit::None {
+    fn paint(&self, unit: &Unit, colors: &Colors, content: String) -> ColoredString {
+        if unit == &Unit::None {
             colors.colorize(content, &Elem::NonFile)
-        } else if self.unit == Unit::Byte || self.unit == Unit::Kilo {
```
meain: I think you removed this accidentally

Philipp-M: Oh how did that happen... Great catch, thanks.
Rebased and fixed the first commit.

meain:
@Peltoche any suggetion on the flag? `--total-size` looks good?

PR #236, opened 2019-06-17
fix name padding to include indicator's length

thammin:
Before:
Screen Shot 2019-06-17 at 12 11 51

After:
Screen Shot 2019-06-17 at 12 12 05


meain:
Hey @thammin , looks like this fix will break it for cases where we do not show the indicator.

thammin:
@meain You are right, fixed it by return 0 length when not showing indicator.

PR #238, opened 2019-06-18
Fix more spacing issues

meain:
A few more changes related to #236


meain:
@Peltoche 
I have added in a comment. I decided to use comment as it was getting used only once per function.

PR #249, opened 2019-07-21
Add indent flag

### LINKED ISSUE #22: Add indentation before filenames
`lsd` looks like this:

image

`colorls` looks like this:

image

It would be cool to be able to set the indentation somehow.

alichtman:
Fix #22. 

For some reason this breaks file icon rendering? I can't figure out why, but I also haven't ever written in Rust before this, so I might be missing something basic. If you have any idea how to make this work with file icons, I'd appreciate some help.

This isn't a fully fleshed out feature (see the Issues page), but it's enough for what I want it to do (aside from the file icon rendering issue).


src/app.rs
```diff
@@ -53,6 +53,13 @@ pub fn build() -> App<'static, 'static> {
                 .number_of_values(1)
                 .help("Whether to use fancy or unicode icons"),
         )
+        .arg(
+            Arg::with_name("indent")
+                .short("i")
```
Peltoche: This short flag is already taken by the original ls to show the inode number. I don't think it's a good idea to use the same flag for an another behavior. Can you remove it in order to keep only the long format?

src/app.rs
```diff
@@ -53,6 +53,13 @@ pub fn build() -> App<'static, 'static> {
...
+                .long("indent")
```
Peltoche: There is a lot of possible indentations. Specifying `prefix-indent` would avoid conflicts with some possible future features. It's longer but I guess that this flag will be only used inside the alias so it shouldn't be too inconvenient.

@meain what do you think of it?

alichtman: It is a little long. I'd prefer to find a shorter flag, but I can't think of any at the moment.

alichtman:
Note that this PR currently breaks file icon rendering. Do not merge.

src/app.rs
meain: We should probably reserve the short `i` flag, maybe for something like ignore which will be a more important flag.

I think we should go with `prefix-intent`. Yes, it is a bit long, but I feel like this will not be a common flag.

Peltoche:
Indeed... I dont think this bug is due to your code. The git bisect indicate that the bug come from dd10a6e5431beafab064e950ff577d2cd03a38f7. I will fix this in an another PR.

Peltoche:
https://github.com/Peltoche/lsd/pull/252 should fix it

alichtman:
Any update on this?

alichtman:
@Peltoche Bump

alichtman:
Awesome! Thanks!

PR #261, opened 2019-08-05
Can now replace symlink arrow with --symlink-arrow

zeevro:
I added an option to replace the string used for symlink arrows since my terminal font doesn't have ⇒


zeevro:
One of the AppVeyor jobs failed on a crate download and I couldn't find a way to ask it to try it again.

xSetech:
One whole flag for the symlink arrow seems silly when nerd-font and/or font-awesome are listed as prerequisites. A better flag would disable any non-ascii iconography. Or, pre-processor directives that produce font-independent builds.

Peltoche:
Hi @zeevro ,

Thanks for this PR and sorry for the delay. I was in holidays. 

I'm agree with @xSetech  for the flag. I think we should see this arrow like an icon and so handle like it. This should include to handle the flag `icon` and `classic`.

What do you think of it?

zeevro:
Sounds like a good idea to me.
I just wanted a quick fix for my own setup. :)

Peltoche:
After some reflection I think that any option around the arrow seems overkill. I don't want to add complexity for a silly arrow so I would like to propose the simplest solution: changing the arrow for the classic `->`.

This solution does the job, is simple and would solve your issue

meain:
Hey just a small suggetion if you are going with that approch.
We should probably use `=>` instead of `->` since `=` seems to center itself better and it will look more like an arrow.

Peltoche:
Following the discussion above I close the PR because the flags doesn't seems to be a viable solution.

PR #262, opened 2019-08-07
let lscolors apply default colors if LS_COLORS not defined

### LINKED ISSUE #90: Use terminal color scheme
## Expected behavior
`lsd` uses colors from my terminal color scheme, being consistent with all the other command line tools.

## Actual behavior
It appears `lsd` uses hard-coded colors, instead of the terminal color scheme. The hard coded colors don't look as nice. See screenshot below.

image

My terminal colors:
image

meain:
Fixes #90

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -77,7 +77,7 @@ impl Colors {
         };
         let lscolors = match theme {
             Theme::NoColor => None,
-            Theme::Default => LsColors::from_env(),
+            Theme::Default => Some(LsColors::from_env().unwrap_or_default()),
             Theme::NoLscolors => None,
         };
 


Kevin-Mok:
This doesn't seem to work for me. The compiled binary still uses the hardcoded colors.

meain:
@Kevin-Mok  Hmm, that is interesting.

We depend on a tool called `lscolors` to do the coloring.
Is it possible for you to check if [lscolors](https://github.com/sharkdp/lscolors) gives you the right colors.

You could do it by 

```sh
cargo install lscolors  # creates a binary
lscolors *
```

Kevin-Mok:
Ah, so `lscolors` doesn't work for me - it just hangs. Not sure why.

meain:
@Kevin-Mok 

Just running `lscolors` will wait for input, you have to run `lscolors *` or `lscolors` with some filename.

Kevin-Mok:
:man_facepalming: Right. So, `lscolors` does give me the correct colors.

Issue #264, opened 2019-08-08
v0.16.0 is adding extra lines in long listing mode


daiyam:
Hi,

Since v0.16.0, when long listing, lsd is adding extra lines when a single line is wider that the console's width.

## behavior of lsd v0.15.1
![Screenshot](https://user-images.githubusercontent.com/587742/62723027-02291280-ba10-11e9-894d-14870339dad3.png)

## behavior of lsd v0.16.0
![Screenshot](https://user-images.githubusercontent.com/587742/62723059-13721f00-ba10-11e9-932b-954ea387e3a3.png)

meain:
This has been fixed by https://github.com/Peltoche/lsd/pull/289
Will be available in the next release.

PR #265, opened 2019-08-09
do not apply padding to last block

meain:
It is not the most elegant solution. But I can't think of a simpler way to do it.

Fixes https://github.com/Peltoche/lsd/issues/264


meain:
Closing in favor of https://github.com/Peltoche/lsd/pull/289

PR #266, opened 2019-08-09
Selected icons for special Unix file types

xSetech:
/var/run:
var_run_example

/dev:
dev_example

/proc/.../fd:
proc_fd_example

### DIFF
--- a/src/icon.rs
+++ b/src/icon.rs
@@ -58,9 +58,35 @@ impl Icons {
 
         let mut res = String::with_capacity(4 + ICON_SPACE.len()); // 4 == max icon size
 
-        // Check directory.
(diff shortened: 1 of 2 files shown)


stgarf:
Nice PR but looks like build is failing due to some tests (either missing or now broken). Happily awaiting merge once fixed! :)

I am meaning to do similar to fix `.lock` files so they don't look like Ruby Gems (ewwwwww, gross!) and add an icon for both `.ko/.kext` and `.pid`.

xSetech:
Added name.rs tests in 6c055d3

meain:
Hopefully this will be the last master merge that I will have to do.

PR #268, opened 2019-08-26
Support to not display icons in 'short' display format

### LINKED ISSUE #267: Feature: only display icons when using --long format
Hi,
it will be great to add a new optional value for the --icon option to only display icon if the --long format is used.
It will allow not to clutter the display when not using the --long format.

A new value (like **long**) for the --icon flag will only display icons if --long format is used.

JrCs:
* Add a new 'long' optional value for the --icon flags. If set
  only 'long' display formats (long or tree) will show icons.

Close #267

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -37,6 +37,7 @@ pub fn build() -> App<'static, 'static> {
                 .long("icon")
                 .possible_value("always")
                 .possible_value("auto")
+                .possible_value("long")
                 .possible_value("never")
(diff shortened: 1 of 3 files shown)


Peltoche:
Hi @JrCs , thanks for this PR.

I think this feature is too custom and could be resolved with an alias like: 
```
alias ll='lsd -l --icon=never'
```

I clone this PR but don't hesitate to exchange on the subject on the issue #267

PR #282, opened 2019-09-14
Fix #275: Implement -I/--ignore-glob

### LINKED ISSUE #275: [Feature Request] Implement --ignore-glob
I regularly end up using `tree -a -I .git` which shows me a tree of everything including hidden files except the `.git` directories. This doesn't work with `lsd` as `-I/--ignore-glob` isn't supported.

sumitsahrawat:


src/flags.rs
```diff
@@ -31,6 +33,11 @@ impl Flags {
         let date_inputs: Vec<&str> = matches.values_of("date").unwrap().collect();
         let dir_order_inputs: Vec<&str> = matches.values_of("group-dirs").unwrap().collect();
         let blocks_inputs: Vec<&str> = matches.values_of("blocks").unwrap().collect();
+        let ignore_globs_inputs: Vec<&str> = if matches.is_present("ignore-glob") {
+            matches.values_of("ignore-glob").unwrap().collect()
```
Peltoche: I think you can keep the same format than the options above here and remove the if/else condition because the `.collect()` will return an empty array if no option is set.

Peltoche:
Hi @sumitsahrawat , thanks for the PR.

Looks almost good, there is only one little change I see in order to keep the same formatting everywhere

src/flags.rs
sumitsahrawat: For some reason removing the `if` clause results in an error:

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/libcore/option.rs:378:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

sumitsahrawat: The above error happens only when running with out `-I`/`--ignore-glob`.

sumitsahrawat: I was able to make it work without an `if` statement by providing a default value in `src/app.rs`.

PR #285, opened 2019-10-05
Basic integration tests, to demonstrate one possible approach.

allenap:
Hi. Last time – well, the first and only time, so far – I did anything significant in `lsd` I caused a regression. Sorry about that. Anyway, it's been on my mind to add some integration tests. This PR is a basic outline of what I had in mind. I think it'll get more complicated as soon as we start adding the `--long` flag, but I have some thoughts about that. I wanted to show you this to see if you like the approach. If so I can put some more time into it.


Peltoche:
Hello @allenap, thanks a lot for the PR! Some integration tests seems a really good idea indeed.

Don't worry for the regressions, it should had been catched during the review so shame on me.

PR #289, opened 2019-10-23
Use only blocks

Peltoche:


src/app.rs
```diff
@@ -184,8 +184,15 @@ pub fn build() -> App<'static, 'static> {
...
-                .possible_values(&["permission", "user", "group", "size", "date", "name"])
-                .default_value("permission,user,group,size,date,name")
+                .possible_values(&[
+                    "permission",
+                    "user",
+                    "group",
+                    "size",
+                    "date",
+                    "name",
+                    "name-with-symlink",
```
meain: don't think this is supposed to be here

Peltoche: "name-with-symlink" ? For now it sadly is because the padding rules and the formatting change between a `name` and a `name-with-symlink`.  I will try to re-merge them but it require a generic way to give parameters to blocks.

I should be able to remove in a next commit inside this PR I hope

meain:
Can we make use of `--no-symlink` flag? We had introduced this for a reason similar to this.
I could be wrong, I am not certain on the padding issue that you are talking about.

On Oct 24, Pierre Peltier wrote:
> Peltoche commented on this pull request.
> 
> 
> 
> > @@ -184,8 +184,15 @@ pub fn build() -> App<'static, 'static> {
>                  .multiple(true)
>                  .number_of_values(1)
>                  .require_delimiter(true)
> -                .possible_values(&["permission", "user", "group", "size", "date", "name"])
> -                .default_value("permission,user,group,size,date,name")
> +                .possible_values(&[
> +                    "permission",
> +                    "user",
> +                    "group",
> +                    "size",
> +                    "date",
> +                    "name",
> +                    "name-with-symlink",
> 
> "name-with-symlink" ? For now it sadly is because the padding rules and the formatting change between a `name` and a `name-with-symlink`.  I will try to re-merge them but it require a generic way to give parameters to blocks.
> 
> I should be able to remove in a next commit inside this PR I hope
> 
> -- 
> You are receiving this because your review was requested.
> Reply to this email directly or view it on GitHub:
> https://github.com/Peltoche/lsd/pull/289#discussion_r338434595

-- 
Regards:
Abin Simon (Developer)
meain.io - abinsimon10@gmail.com

Peltoche:
> Can we make use of `--no-symlink` flag? We had introduced this for a reason similar to this. I could be wrong, I am not certain on the padding issue that you are talking about.

I just have restored the `no-symlink`  flag like before and you can forget my previous response, I was on a false issue with an incorrect solution.

Peltoche:
This PR is bigger than usual but I don't see how to reduce the size while keeping the program working. Yet there is still a lot of stuff to do but this PR is already fare too big but they will be done in some other PR.

@meain what do you think of this PR? Do you think it's ok to merge it?

meain:
I have not had time to take a look at the PR yet. I will take a look at it this weekend.

Since it is, at the end of the day a single refactor around how things get rendered(as far as I can tell) I think it would be better to merge it as one single PR.

src/flags.rs
```diff
@@ -206,6 +214,8 @@ impl<'a> From<&'a str> for Block {
             "user" => Block::User,
             "group" => Block::Group,
             "size" => Block::Size,
+            "size_value" => Block::SizeValue,
```
meain: These are not added to `app.rs` in the cli arg parser.

build.rs
```diff
@@ -18,8 +18,8 @@ use std::process::exit;
 include!("src/app.rs");
 
 fn main() {
-    match version_check::is_min_version("1.31.0") {
-        Some((true, _)) => {}
+    match version_check::is_min_version("1.33.0") {
+        Some(true) => {}
         // rustc version too small or can't figure it out
         _ => {
             writeln!(&mut io::stderr(), "'lsd' requires rustc >= 1.31.0").unwrap();
```
zwpaper: should we also update the error msg to `1.33.0`?

src/flags.rs
```diff
@@ -95,15 +92,31 @@ impl Flags {
             }
             None => usize::max_value(),
         };
-        let no_symlink = matches.is_present("no-symlink");
-        let total_size = matches.is_present("total-size");
+
+        let blocks: Vec<Block> = if !blocks_inputs.is_empty() {
```
meain: `lsd --tree -l` will not work

meain:
@Peltoche The code looks good. But there seems to be mess when using `-l` with `--tree`. It seem to only respect `-l`

One more thing, Maybe we should reduce the spacing between the icon and the name of the file and increase the spacing before by one. Ideally two space before the icon and one after.

```diff
diff --git a/src/display.rs b/src/display.rs
index 5223306..401fcec 100644
--- a/src/display.rs
+++ b/src/display.rs
@@ -38,7 +38,7 @@ fn inner_display_grid(
 
     let padding_rules = get_padding_rules(&metas, flags);
     let mut grid = Grid::new(GridOptions {
-        filling: Filling::Spaces(1),
+        filling: Filling::Spaces(2),
         direction: if flags.layout == Layout::OneLine {
             Direction::LeftToRight
         } else {
diff --git a/src/icon.rs b/src/icon.rs
index 3d7b42f..f78a516 100644
--- a/src/icon.rs
+++ b/src/icon.rs
@@ -16,7 +16,7 @@ pub enum Theme {
     Unicode,
 }
 
-const ICON_SPACE: &str = "  ";
+const ICON_SPACE: &str = " ";
 
 // In order to add a new icon, write the unicode value like "\ue5fb" then
 // run the command below in vim:
```

|before|after|
|---|---|
||

build.rs
Peltoche: good catch! Fixed

src/flags.rs
Peltoche: At the moment, I need the `size_value` for the left padding calculation but I have removed the `unit_value`.

I currently work to add the left padding feature into the term_grid library. Once it's done we should be able to remove all the padding logic. I hope this will be my next PR

Peltoche: > @Peltoche The code looks good. But there seems to be mess when using `-l` with `--tree`. It seem to only respect `-l`


Fixed

Peltoche:
> @Peltoche The code looks good. But there seems to be mess when using `-l` with `--tree`. It seem to only respect `-l`


Fixed

Peltoche:
> One more thing, Maybe we should reduce the spacing between the icon and the name of the file and increase the spacing before by one. Ideally two space before the icon and one after.

src/display.rs
```diff
@@ -114,9 +36,14 @@ fn inner_display_grid(
 ) -> String {
     let mut output = String::new();
 
+    let padding_rules = get_padding_rules(&metas, flags);
     let mut grid = Grid::new(GridOptions {
         filling: Filling::Spaces(2),
```
meain: This seems to be how it is handled everywhere else. I think we should go with this too.

```suggestion
         filling: if flags.layout == Layout::OneLine {
            Filling::Spaces(1)
         } else {
            Filling::Spaces(2)
         },
```

meain:
The way `--size short` is displayed has changed. There is now an extra space between the value and the unit.

Peltoche:
> The way `--size short` is displayed has changed. There is now an extra space between the value and the unit.

All the spacing between blocks have changed from one space to two. This is due to the fix for your previous comment about the two spaces before the icons.

I revert the two spaces separator

meain:
> The way `--size short` is displayed has changed. There is now an extra space between the value and the unit.

|berfore|after|
|---|---|
||

I am talking about the space between `3` and `K` when displaying size.

About the grid spacing, we should have two space for normal `lsd` and one space when running with `-l` flag. That is how everyone else seems to do it.

meain:
@Peltoche I think we should go with two spaces for grid when not having `-l` and for long format, we could change to 1 space. That seems to be what gnu `ls` does.

That is the only change I think we need for this and we could merge this in.

meain:
@Peltoche Sorry if I starting to sound like a broken record about the whole spacing stuff. What I was saying was that we could use two different sizes based on the layout.

Something like this:
```rust
let mut grid = match flags.layout {
    Layout::OneLine => Grid::new(GridOptions {
        filling: Filling::Spaces(1),
        direction: Direction::LeftToRight,
    }),
    _ => Grid::new(GridOptions {
        filling: Filling::Spaces(2),
        direction: Direction::TopToBottom,
    }),
};
```

Issue #276, opened 2019-09-04
Request: Implement -i (--inode) to include inode numbers


AndydeCleyre:
## Expected behavior
```bash
touch README.rst
ls -i README.rst
```
`53850413 README.rst`
```bash
busybox ls -i README.rst
```
`53850413 README.rst`
```bash
lsd -i README.rst
```
`53850413  README.rst`
```bash
ls -li README.rst
```
`53850413 -rw-r--r-- 1 andy andy 0 Sep  4 15:56 README.rst`
```bash
busybox ls -li README.rst
```
`53850413 -rw-r--r--    1 andy     andy             0 Sep  4 15:56 README.rst`
```bash
lsd -li README.rst
```
`53850413 .rw-r--r-- andy andy 0 B Wed Sep  4 15:56:36 2019   README.rst`

## Actual behavior
`-i` not implemented.

Both GNU Coreutils and BusyBox support `-i` to show inode numbers, and Coreutils additionally supports the equivalent `--inode`. `-i` should be sufficient, and handy for identifying hard links.

Peltoche:
Hi @AndydeCleyre,

This look a great feature indeed. I guess it can be integrated as a block. I don't have time to work on this feature but if you want to make a PR we would be happy to review it.

AndydeCleyre:
If it were implemented as a block, would that preclude the following behavior?

```bash
lsd -i
```
```
57011337   app.rs    57011339   core.rs     57011341   flags.rs  57011343   main.rs  57011354   sort.rs
57011338   color.rs  57011340   display.rs  57011342   icon.rs   57094940   meta
```

In other words, would we have to use the `--long`/table view to access it? In this case and in general I prefer keeping behavior close to other `ls`(-like) implementations.

Peltoche:
I don't think it would preclude anything. It would just facilitate the `-i` integration as is only require to run an ls with the blocks `name` and `inode`.  

So `lsd -i` == `lsd --blocks inode,name`

meain:
This is now possible with https://github.com/Peltoche/lsd/pull/296

PR #296, opened 2019-11-02
 impl showing inode by setting blocks

zwpaper:
working on https://github.com/Peltoche/lsd/issues/276

but I noticed https://github.com/Peltoche/lsd/pull/289, should I wait until it merged?

windows do not have the `inode`, should we just report errors and exit when specified `inode` on windows?

BTW, I did a `cargo fmt`, and updated some file format.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -187,7 +187,6 @@ pub fn build() -> App<'static, 'static> {
                     "name",
                     "inode",
                 ])
-                .default_value("permission,user,group,size,date,name")
                 .help("Specify the blocks that will be displayed and in what order"),
         )
         .arg(
--- a/src/app.rs
+++ b/src/app.rs
@@ -210,6 +210,14 @@ pub fn build() -> App<'static, 'static> {
                 .default_value("")
                 .help("Do not display files/directories with names matching the glob pattern(s)"),
         )
+        .arg(
+            Arg::with_name("inode")
+                .short("i")
+                .long("inode")
+                .multiple(true)
+                .conflicts_with("blocks")
+                .help("Display inode and file names"),
+        )
 }
 
 fn validate_date_argument(arg: String) -> Result<(), String> {
--- a/src/app.rs
+++ b/src/app.rs
@@ -178,7 +178,16 @@ pub fn build() -> App<'static, 'static> {
                 .multiple(true)
                 .number_of_values(1)
                 .require_delimiter(true)
-                .possible_values(&["permission", "user", "group", "size", "date", "name"])
+                .possible_values(&[
+                    "permission",
+                    "user",
+                    "group",
+                    "size",
+                    "date",
+                    "name",
+                    "inode",
+                ])
+                .default_value("permission,user,group,size,date,name")
                 .help("Specify the blocks that will be displayed and in what order"),
         )
         .arg(
--- a/src/app.rs
+++ b/src/app.rs
@@ -215,7 +215,7 @@ pub fn build() -> App<'static, 'static> {
                 .short("i")
                 .long("inode")
                 .multiple(true)
-                .help("Display inode(Linux only)"),
+                .help("Display the index number of each file"),
         )
 }
 
--- a/src/app.rs
+++ b/src/app.rs
@@ -215,8 +215,7 @@ pub fn build() -> App<'static, 'static> {
                 .short("i")
                 .long("inode")
                 .multiple(true)
-                .conflicts_with("blocks")
-                .help("Display inode and file names"),
+                .help("Display inode(Linux only)"),
         )
 }
 
--- a/src/display.rs
+++ b/src/display.rs
@@ -221,6 +221,9 @@ fn get_output<'a>(
     let mut strings: Vec<ANSIString> = Vec::new();
     for block in flags.blocks.iter() {
         match block {
+            Block::INode => {
+                strings.push(meta.inode.render(colors));
+            }
             Block::Permission => {
                 let s: &[ColoredString] = &[
                     meta.file_type.render(colors),
--- a/src/display.rs
+++ b/src/display.rs
@@ -221,9 +221,7 @@ fn get_output<'a>(
     let mut strings: Vec<ANSIString> = Vec::new();
     for block in flags.blocks.iter() {
         match block {
-            Block::INode => {
-                strings.push(meta.inode.render(colors));
-            }
+            Block::INode => strings.push(meta.inode.render(colors)),
             Block::Permission => {
                 let s: &[ColoredString] = &[
                     meta.file_type.render(colors),
--- a/src/flags.rs
+++ b/src/flags.rs
@@ -205,6 +205,7 @@ pub enum Block {
     SizeValue,
     Date,
     Name,
+    INode,
 }
 impl<'a> From<&'a str> for Block {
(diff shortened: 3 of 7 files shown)


meain:
Hey @zwpaper .

Yeah, i think it would be better if you could write this on top of that once that is merged as that is a major rewrite on a lot of code that this PR modifies.

meain:
@zwpaper Sorry that it took a little while, but that PR has been merged!

zwpaper:
@meain conflict fixed, I return 0 for windows currently.

Do you have any better ideas for inode in windows?

src/app.rs
```diff
@@ -179,7 +179,16 @@ pub fn build() -> App<'static, 'static> {
...
+                .possible_values(&[
+                    "permission",
+                    "user",
+                    "group",
+                    "size",
+                    "date",
+                    "name",
+                    "inode",
+                ])
+                .default_value("permission,user,group,size,date,name")
```
meain: We should not have a default value for this since the new rewrite or it will always resort to displaying as if for `-l`.

meain:
@zwpaper for windows maybe we could use `-` instead of using `0`

src/app.rs
zwpaper: fixed by deleting the default value

zwpaper:
@meain fixed the two things you pointed out, check it again, please

src/display.rs
```diff
@@ -221,6 +221,9 @@ fn get_output<'a>(
     let mut strings: Vec<ANSIString> = Vec::new();
     for block in flags.blocks.iter() {
         match block {
+            Block::INode => {
```
meain: Can't we just write this in one line, keep it consistent.

src/meta/inode.rs
```diff
@@ -0,0 +1,66 @@
+use crate::color::{ColoredString, Colors, Elem};
+use std::fs::Metadata;
+
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+pub struct INode {
+    index: u64,
+    valide: bool,
```
meain: I'm guessing this is supposed to be `valid`.

In any case, wouldn't it be better to just use `Option` for `index` than have a separate flag?

src/meta/inode.rs
```diff
@@ -0,0 +1,66 @@
...
+            index: 0,
+            valide: false,
+        }
+    }
+}
+
+impl INode {
+    pub fn render(&self, colors: &Colors) -> ColoredString {
+        if !self.valide {
+            return colors.colorize(String::from("-"), &Elem::SymLink);
```
meain: We could probably use `NonFile` for color. A faded color will probably look better.

src/meta/inode.rs
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        colors.colorize(self.index.to_string(), &Elem::SymLink)
```
meain: It won't be totally bad idea to create a new color maybe? You could just copy the color from `SymLink` or we could probably check with @Peltoche and assign something colorful.

src/meta/inode.rs
zwpaper: I try added two colors for the inode, as lsd is trying to provide a colorful ls, `NonFile` looks a little bit boring for `inode,name`.

Maybe @Peltoche could help assign a better color for this?

Currenttly impl:
colorful for unix:
image

faded for windows:
image

src/meta/inode.rs
zwpaper: added a new color for `inode`, will continue to discuss in the previous comment

src/meta/inode.rs
zwpaper: updated to 

```rust
pub struct INode {
    index: Option<u64>,
}
```

zwpaper:
@meain fixed all the requested except the inode color, could you please review it again and discuss about the color.

PS:
Great thanks for the review, this really makes this my good first rust issue.

xSetech:
I object to the `-l` flag not affecting output. IMO if I mix `-i` with `-l` it means I want all stats _and_ the inode, not just ignoring `-l`.

Here's the behavior of this branch with and without `-l`:

Screen Shot 2019-12-15 at 2 53 20 PM

And here's the behavior of coretools' ls with and without the `-l` flag:

Screen Shot 2019-12-15 at 2 55 54 PM

src/meta/inode.rs
meain: My bad, I meant the faded color only for windows.

meain:
Hey @zwpaper , I think it would be a better idea to take care of `-i -l` before we merge.

zwpaper:
@meain sorry for the delay, `-i` and `-l` fixed

src/app.rs
```diff
@@ -215,8 +215,7 @@ pub fn build() -> App<'static, 'static> {
                 .short("i")
                 .long("inode")
                 .multiple(true)
-                .conflicts_with("blocks")
-                .help("Display inode and file names"),
+                .help("Display inode(Linux only)"),
```
meain: Use something like:

`print the index number of each file`

src/flags.rs
```diff
@@ -112,6 +111,11 @@ impl Flags {
             _ => vec![Block::Name],
         };
 
+        // Add inode as first column if with inode flag
```
meain: If `inode` is in blocks then do not change its position.

We just have to check if we have `-l` and `-i` without `--blocks` and in that case have inode infront.
The change should be going somewhere here: https://github.com/Peltoche/lsd/blob/87e87d87cdf91d7e8a3f385b0a5beb96ddbd8824/src/flags.rs#L99

zwpaper: I have checked is the `blocks` containing the `inode` below this.

I update the blocks below the creation is because there are too many cases if combining into one `if else`

- `-l -i`
- `-i --blocks name,inode`
- `-i --blocks name`
- `-l -i --blocks name,inode`

So I check is `blocks` containing inode, and is `inode flag` set, and update `blocks` if needed

src/app.rs
zwpaper: I updated the help msg, but using the `Display`,  for the consistent of other help msgs.

src/flags.rs
meain: My bad, I thought the code did something else.
This looks good.

zwpaper: Sorry for interrupting your vacation, have fun for it!😁

src/flags.rs
```diff
@@ -94,21 +98,24 @@ impl Flags {
             None => usize::max_value(),
         };
 
-        let blocks: Vec<Block> = if !blocks_inputs.is_empty() {
```
meain: Any reason why you switched this from if to match?

zwpaper: I was trying to add `inode` to this `if else` chain at first, but found that a `switch case` may be more suitable.

after googling, `match` like this is how `switch` do in `rust`.

Or could you leave some advice? Rust is all new to me

meain: For `inode` I think the current way in which you have done would be the best option. As you said, it would be too many ifs or nested matches, both not ideal.

I just think that if is a more readable than match here. This match is more like "match all but if (condition)" which is essentially if. Could you change this back and. Once that is done, I think we are good to go.

PR #297, opened 2019-11-03
Make windows fast again (on domain connected machines)

### LINKED ISSUE #200: lsd is slow Windows
Performance is very slow - about 50 seconds to list a few files.

The long delay appears to be in calling `lsd::meta::windows_utils::get_acl_access_mask`.

Possibly relevant details - I'm testing on Win10 1903 on a domain joined machine. I suspect the issue is contacting the domain controller, because when I disconnect the machine from the network, `lsd.exe` gives me this error:

```
cannot access '.': The specified domain either does not exist or could not be contacted. (os error 1355)
```

I also tested on a non-domain joined machine (Server 2016) and it seemed plenty fast.

omerbenamram:
Hi @Peltoche, thanks for making `lsd`!

I've looked into #200 and did some profiling, the indeed almost all of the time spent is in `GetEffectiveRightsFromAclW` when trying to query one of the "Well known" sids.

The problem is that on domain joined machines, windows will try to contact the DC for every file (you can see the calls in the flamegraph to `GetDomainInfo`), which could take a few hundread miliseconds per file, this PR changes `lsd` to only query user and group information, which return instantly.

image

There isn't a 1:1 mapping between windows file security conecpts and linux's, and querying this information for an entire folder takes 20-50 seconds, which makes `lsd` difficult to use.
Personally I find the information for User/Group more than sufficient :)

I could also feature-gate this change if you feel more comfortable with it.

Closes #200.

### DIFF
--- a/src/meta/windows_utils.rs
+++ b/src/meta/windows_utils.rs
@@ -85,49 +85,11 @@ pub fn get_file_data(path: &PathBuf) -> Result<(Owner, Permissions), io::Error> 
     // This structure will be returned
     let owner = Owner::new(owner, group);
 
-    // Get the size and allocate bytes for a 1-sub-authority SID
-    // 1 sub-authority because the Windows World SID is always S-1-1-0, with
-    // only a single sub-authority.
-    //
-    // Assumptions: None
-    // "This function cannot fail"
-    //     -- Windows Dev Center docs
-    let mut world_sid_len: u32 = unsafe { winapi::um::securitybaseapi::GetSidLengthRequired(1) };
-    let mut world_sid = vec![0u8; world_sid_len as usize];
-
-    // Assumptions:
-    // - world_sid_len is no larger than the number of bytes available at
-    //   world_sid
-    // - world_sid is appropriately aligned (if there are strange crashes this
-    //   might be why)
-    let result = unsafe {
-        winapi::um::securitybaseapi::CreateWellKnownSid(
-            winnt::WinWorldSid,
-            null_mut(),
-            world_sid.as_mut_ptr() as *mut _,
-            &mut world_sid_len,
-        )
-    };
-
-    if result == 0 {
-        // Failed to create the SID
-        // Assumptions: Same as the other identical calls
-        unsafe {
-            winapi::um::winbase::LocalFree(sd_ptr);
-        }
-
-        // Assumptions: None (GetLastError shouldn't ever fail)
-        return Err(io::Error::from_raw_os_error(unsafe {
-            winapi::um::errhandlingapi::GetLastError()
-        } as i32));
-    }
-
     // Assumptions:
     // - xxxxx_sid_ptr are valid pointers to SIDs
     // - xxxxx_trustee is only valid as long as its SID pointer is
     let mut owner_trustee = unsafe { trustee_from_sid(owner_sid_ptr) };
     let mut group_trustee = unsafe { trustee_from_sid(group_sid_ptr) };
-    let mut world_trustee = unsafe { trustee_from_sid(world_sid.as_mut_ptr() as *mut _) };
 
     // Assumptions:
     // - xxxxx_trustee are still valid (including underlying SID)
@@ -135,8 +97,6 @@ pub fn get_file_data(path: &PathBuf) -> Result<(Owner, Permissions), io::Error> 
     let owner_access_mask = unsafe { get_acl_access_mask(dacl_ptr as *mut _, &mut owner_trustee) }?;
 
     let group_access_mask = unsafe { get_acl_access_mask(dacl_ptr as *mut _, &mut group_trustee) }?;
-
-    let world_access_mask = unsafe { get_acl_access_mask(dacl_ptr as *mut _, &mut world_trustee) }?;
 
     let has_bit = |field: u32, bit: u32| field & bit != 0;
 
@@ -149,9 +109,9 @@ pub fn get_file_data(path: &PathBuf) -> Result<(Owner, Permissions), io::Error> 
         group_write: has_bit(group_access_mask, winnt::FILE_GENERIC_WRITE),
         group_execute: has_bit(group_access_mask, winnt::FILE_GENERIC_EXECUTE),
 
-        other_read: has_bit(world_access_mask, winnt::FILE_GENERIC_READ),
-        other_write: has_bit(world_access_mask, winnt::FILE_GENERIC_WRITE),
-        other_execute: has_bit(world_access_mask, winnt::FILE_GENERIC_EXECUTE),
+        other_read: false,
+        other_write: false,
+        other_execute: false,
 
         sticky: false,
         setuid: false,


lzybkr:
Instead of calling `GetEffectiveRightsFromAclW` for every file, can we instead call it once per unique sid?

And if that still feels too slow - maybe persist this cache in TEMP and refresh it at most once a day?

bgianfo:
Any chance of progress being made here?

bgianfo:
Any chance of progress being made on this PR?

AsafMah:
Why isn't this merged already?

meain:
Hey, sorry that this PR did not get much attention till now. I would like to reopen the PR, but I don't really have any knowledge of how permissions in Windows works.

From what I could understand, "world" is kinda like "everyone else". I would like to know how important of a metric this permission is and how much of an effect that this has.

AsafMah:
I've tried making a version that caches results to "get_acl_access_mask", and while it does help a little, especially with world that only need to be done once,
It is still very very slow compared to a normal dir command.

meain:
I don't think we should cache it. For a tool like this showing stale/incorrect data might be worse that not showing any.

sql-sith:
I hope this PR or something like it can be released as soon as is feasible. I just installed the latest and greatest `lsd` on a domain-joined Windows laptop. Like many others, I am working from home. If I am not on my company VPN, `lsd` fails with `lsd: .\soapui-settings.xml: The specified domain either does not exist or could not be contacted. (os error 1355).` It works fine in WSL but not in pwsh or cmd. See the attached screencap for a couple examples of this.

image

bgianfo:
I think not querying this data and not displaying the hard coded data on windows is the best option. For every day use it makes more sense to make the tool usable, before supporting features like this on windows.

sql-sith:
By the way, I can enumerate the NTFS access permissions of all 866 objects in a directory on my laptop in 10.67 seconds. This is on a work laptop that is a domain member but not attached to the domain, and it is probably not using the fastest available method to do the enumeration.

@meain , you said you don't understand Windows permissions very well, but if you just imagine a Linux system where all the files show a `+` at the end of their mode/permission block in `ls`, that's at least a good start. They also have the concept of explicit denial of permissions, which overrides granted permissions, no matter what. And they have the concept of inheritance, which if enabled allows you to change the permissions of an entire directory tree with one modification to the parent.

Significantly, while filesystem objects in NTFS have owners, they do not have any specific group associated with them, and they might or might not have any permissions relating to something similar to "world."

To give some context, this scheme is really flexible, but it is complex enough that Windows doesn't even try to show a summary of permissions. The closest their standard CLI tools get is `dir /q` from `cmd.exe`, which displays the owner of each object. `Powershell`'s Get-ChildItem displays a `Mode` column, but those are object attributes like read-only, archive, hidden, etc., and not Linux-style modes (like 644). `Get-ChileItem` also has a parameter to list the owner, but there is no way to list permissions with it. What you _can_ do with `Get-ChildItem` is pipe it to `Get-NTFSAccess`, which will show you the actual underlying permissions. Here's an example of that for a typical file and a typical directory:

image

So TL;DR - I say if MS doesn't show permissions, why should `lsd`? Don't include them now if they are causing so much trouble. Later on, maybe somebody (like me?) can circle back and put in a more efficient way to get permissions from Windows. Alternatively, maybe we could consider adding an owner column and an "attributes" block (read-only, etc) like MS does.

meain:
@sql-sith , thanks for the details and screenshots.

Just hard coding `false` as the value for "world" permission feels like a hack to me personally. Here is what I think we could do instead.

`lsd` has a concept of blocks which lets to enable or disable certain blocks of data like `date` or `permission`. As of now, we try to load all the metadata for a file when listing even if we don't have to display it. I think we could change that behavior to instead loading only the metadata needed to display the blocks that the user requested for.
Once this is implemented, anyone on a domain connect machine could just disable the permissions block(in the config file). This should help keep the "world" permission for non domain connect machines which does not seem to have this issue. As an added benefit, this might also speed up listings which don't make use of all the blocks.

In future we could probably add an "attributes" column on Windows like @sql-sith suggested.

meain:
From what I could understand, 2 blocks don't translate over (permissions, group) as windows might have multiple groups and there is no `other`.

Regarding disabling `permission` block(which if we were to do, we should disable both `group` and `permission` blocks) on Windows, even thought it does not exactly translate over, I think we will keep the current default blocks **for now** so as to not change anything for existing users as this specific issue only seems to affect domain connected machines. That said, we could add info about this in the README, and in any error messages when trying to retrieve permissions on Windows machines.

sql-sith:
> From what I could understand, 2 blocks don't translate over (permissions, group) as windows might have multiple groups and there is no `other`.

@meain, that is exactly right. Also, I think that what you are proposing is perfect for now. Keeping the output looking similar is a good thing. I say go for it. :-) I can't help right now because work is too busy but might be able to later ... although maybe I should check what language this is before volunteering.

AsafMah:
Maybe add a flag to not retrieve this data? 
Then people with domain machines can just put it in their config files

meain:
@AsafMah Yes, that is the plan. The idea is the we will only retrieve only the info needed to display the output. So if the user specifies, `block` in the config to not have permissions the code will skip trying to fetch that information.

vvuk:
This affects domain-connected machines the most, but it's still slow even on machines not part of a domain.  In a directory with ~100 files (mostly .dll/.pdb files), lsd `master` (today's) takes 0.65s to produce a listing.  With this patch, it takes 0.44s, which is a significant improvement.

Note that 0.44s is still extremely slow; I haven't done any profiling yet to figure out why.  Running `bash -c ls` (running ls through WSL) takes 0.07s.  Running powershell `Show-ChildItemColumns` takes 0.01s.  lsd should be able to hit at least WSL speeds; half a second is painful enough that it's not really usable as a replacement on windows.

vvuk:
It's `GetEffectiveRightsFromAclW` causing the slowdown.  From a trace:
image

There was a comment earlier about not querying for data that isn't going to be displayed; that sounds like the right approach to solving this, though in the interim it might be helpful to just have a way to not query any of this on windows.

vvuk:
I just added a quick and dirty way to disable the ownership/permissions checks on windows in #484.  With that feature, lsd becomes as fast as any other tool (down to 0.04s for the dir above -- so twice as fast as WSL, and within range of the speed-of-light  no-process-spawn powershell cmdlet)

sql-sith:
> I just added a quick and dirty way to disable the ownership/permissions checks on windows in #484. With that feature, lsd becomes as fast as any other tool (down to 0.04s for the dir above -- so twice as fast as WSL, and within range of the speed-of-light no-process-spawn powershell cmdlet)

This is great news! I came out to see if anyone had worked on this yet. I'll give it a spin. Feel free to hit me up directly if/when you want to find fast ways to enumerate ACLs in Windows (not that you'll need my help, but just in case).

sql-sith:
@vvuk , I am not a Rustacean, so I have a really basic question. After building the executable from your PR, when I use it on a domain-connected computer it is still very slow, unfortunately. Is there anything I need to do to enable this feature? Thanks a lot for working on this.

vvuk:
Yep, you need to build with `cargo build --release --features no-windows-permissions`.  The feature is not enabled by default.

sql-sith:
> Yep, you need to build with `cargo build --release --features no-windows-permissions`. The feature is not enabled by default.

See? I know nothing about rust. Or rather, now I know one thing. :) Thanks!

AsafMah:
Also see this PR: https://github.com/Peltoche/lsd/pull/475
Where I added not getting permissions if the block isn't requested.

meain:
This is what I have in mind so far. For now, as a stopgap solution we can have permission fetching disabled in windows under a feature flag (https://github.com/Peltoche/lsd/pull/484). The issue with https://github.com/Peltoche/lsd/pull/475 is that we need the permissions to color file names even if they are not explicitly requested.
In the long term, the idea is to either speed up the actual fetch (ref : https://github.com/Peltoche/lsd/pull/475#issuecomment-777101864) or show permission on windows in a more windowsy way (ref: https://github.com/Peltoche/lsd/pull/297#issuecomment-758370154) for just the one user (which I guess should be relatively faster to fetch).

meain:
Closing this. https://github.com/Peltoche/lsd/pull/484 seems to be better for now.

Issue #298, opened 2019-11-04
Argument '--depth <num>' cannot be used multiple times


nathanph:
I often set what I consider to be sane defaults in my aliases.
E.g., `alias lt="lsd --tree --depth 2"`

So that when I run I run a command, `lt` for example, I will more often than not get the output that I want.

When I want to override my defaults, such as `depth`, I will just append them to my aliased command, `# lt --depth 5`.

## Expected behavior

My expectation is that the most recent argument overrides the oldest.

In this case: `lsd --tree --depth 2 --depth 5`, I would expect a tree of depth 5, *not 2*, to be produced.

## Actual behavior

An argument parsing error:

```
error: The argument '--depth <num>' was provided more than once, but cannot be used multiple times

USAGE:
    lsd --blocks <blocks>... --color <color>... --date <date>... --depth <num> --group-dirs <group-dirs>... --icon <icon>... --icon-theme <icon-theme>... --classify --long --size <size>... --tree

For more information try --help
```

Peltoche:
Hi @nathanph, thanks for the report.

This seems to be a fair request. Some of the flags seems to already have this feature. I guess we will have to align all the other flags on it.

PR #307, opened 2019-12-01
Allow usage of multiple 'depth' arguments

abazylewicz:
Fixes behavior mentioned in #298


abazylewicz:
in the mentioned issue @Peltoche mentioned that
> I guess we will have to align all the other flags on it.

I checked the other flags and it looks like all the ones that take values already allow overriding them

meain:
It seems to be breaking in rust `1.31`

abazylewicz:
Yes, it looks like Rust allows references to by-move bindings in match guards only from 1.39 version. I'll try to rework PR.

PR #309, opened 2019-12-05
maint/cicd ~ always run all GHA:Style testing steps

rivy:
- always runs all style testing steps regardless of failure of the prior step
- with this change, `clippy` tests will run regardless of `fmt` test success/failure

.# [why]

The GHA:Style testing steps are essentially independent of each other and give independent / complementary information about style issues. So, running all of them despite errors can increase code iteration efficiency.

Note: as intended, the GHA:Style job will still fail if any individual step fails.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -13,11 +13,11 @@ on: [push, pull_request]
 jobs:
   style:
     name: Style
-    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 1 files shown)


rivy:
I just noticed that `clippy` tests weren't running if `fmt` tests were failing.
This fixes it so both run regardless.

Peltoche:
Hi @rivy,

It seems that the commit have been already merged with #310 so this PR is not more required.

PR #320, opened 2020-01-17
added csharp and visual studio sln icons

jpda:


meain:
Hi @jpda , thanks for the PR. Would you mind squashing the commits?

jpda:
I ended up in a bad state so doing it in a single commit under #321

Issue #319, opened 2020-01-17
Directory wildcard problem


jtyr:
`lsd` 0.16.0 is omitting directory names when trying to target a specific file across multiple directories via wildcard:

```shell
mkdir a b c
touch {a,b,c}/test.txt
# Normal ls output
ls */test.txt
a/test.txt  b/test.txt  c/test.txt
# lsd output
lsd --icon never */test.txt
test.txt  test.txt  test.txt
```

dvvvvvv:
@Peltoche @meain 
https://github.com/dvvvvvv/lsd/tree/relative-path

I've been working on this issue for a few days. The above branch contains the following modifications:
1. meta::Name includes PathBuf instead of String

2. Name::render function accepts DisplayOption enum value
DisplayOption::FileName: draw file name(using when drawing tree)
DisplayOption::Relative{ base_path }: draw relative path to base_path
DisplayOption::None: draw path it self

3. the file path passed directly has None DisplayOption. so it can draw file paths like ls

$ lsd ../Cargo.*
 ../Cargo.lock   ../Cargo.toml

$ lsd meta/*.rs 
 meta/date.rs        meta/inode.rs   meta/owner.rs         meta/symlink.rs
 meta/filetype.rs    meta/mod.rs     meta/permissions.rs   meta/windows_utils.rs
 meta/indicator.rs   meta/name.rs    meta/size.rs

$ lsd /home/dvvv/src/lsd/src/*.rs
 /home/dvvv/src/lsd/src/app.rs       /home/dvvv/src/lsd/src/flags.rs
 /home/dvvv/src/lsd/src/color.rs     /home/dvvv/src/lsd/src/icon.rs
 /home/dvvv/src/lsd/src/core.rs      /home/dvvv/src/lsd/src/main.rs
 /home/dvvv/src/lsd/src/display.rs   /home/dvvv/src/lsd/src/sort.rs

4. I realized while writing. The relative path option does not seem to be necessary. wow

but This doesn't seem to be rusty code. The Name structure must always have a PathBuf: This is a duplicate of the Meta structure.
I've tried changing it to a Path reference that refers to Meta's PathBuf, but that's not possible because move occurs when the Meta structure is created(as far as I know) .
I've tried changing the Name structure to a Path structure, but that doesn't fit the module named Meta.
I would like to hear you guys opinion.

jtyr:
@dvvvvvv Would you mind to create a PR with your change so we can review it?

[...]

PR #328, opened 2020-02-08
Relative path

dvvvvvv:
This pull request contains following changes

1. meta::Name includes PathBuf instead of String

2. Name::render function accepts DisplayOption enum value
    DisplayOption::FileName: draw file name(using when drawing tree)
    DisplayOption::Relative{ base_path }: draw relative path to base_path
    DisplayOption::None: draw path it self

3. the file path passed directly has None DisplayOption. so it can draw file paths like ls
    $ lsd ../Cargo.*
     ../Cargo.lock  ../Cargo.toml
    
    $ lsd meta/.rs
     meta/date.rs  meta/inode.rs  meta/owner.rs  meta/symlink.rs
     meta/filetype.rs  meta/mod.rs  meta/permissions.rs  meta/windows_utils.rs
     meta/indicator.rs  meta/name.rs  meta/size.rs
    
    $ lsd /home/dvvv/src/lsd/src/.rs
     /home/dvvv/src/lsd/src/app.rs  /home/dvvv/src/lsd/src/flags.rs
     /home/dvvv/src/lsd/src/color.rs  /home/dvvv/src/lsd/src/icon.rs
     /home/dvvv/src/lsd/src/core.rs  /home/dvvv/src/lsd/src/main.rs
     /home/dvvv/src/lsd/src/display.rs  /home/dvvv/src/lsd/src/sort.rs
    
4. also, recursive directory listing has Relative DisplayOption, so it looks like ls and previous lsd
    $ lsd src/*
     src/app.rs   src/color.rs   src/core.rs   src/display.rs   src/flags.rs   src/icon.rs   src/main.rs   src/sort.rs

    src/meta:
     date.rs       indicator.rs   mod.rs    owner.rs         size.rs      windows_utils.rs
     filetype.rs   inode.rs       name.rs   permissions.rs   symlink.rs

i don't think this pull request looks rusty. The Name structure must always have a PathBuf: This is a duplicate of the Meta structure.
I've tried changing it to a Path reference that refers to Meta's PathBuf, but that's not possible because move occurs when the Meta structure is created(as far as I know) .
I've tried changing the Name structure to a Path structure, but that doesn't fit the module named Meta.
I would like to hear you guys opinion.

### DIFF
--- a/src/meta/name.rs
+++ b/src/meta/name.rs
@@ -45,7 +45,10 @@ impl Name {
     }
 
     pub fn file_name(&self) -> &str {
-        self.path.file_name().and_then(OsStr::to_str).unwrap_or(&self.name)
+        self.path
+            .file_name()
+            .and_then(OsStr::to_str)
+            .unwrap_or(&self.name)
     }
 
     fn relative_path<T: AsRef<Path> + Clone>(&self, base_path: T) -> PathBuf {
--- a/src/meta/name.rs
+++ b/src/meta/name.rs
@@ -13,13 +13,36 @@ pub enum DisplayOption<'a> {
 
 #[derive(Clone, Debug, Eq)]
 pub struct Name {
+    pub name: String,
     path: PathBuf,
     extension: Option<String>,
-    pub display_name: String,
     file_type: FileType,
 }
 
 impl Name {
+    pub fn new(path: &Path, file_type: FileType) -> Self {
+        let name = match path.file_name() {
+            Some(name) => name.to_string_lossy().to_string(),
+            None => path.to_string_lossy().to_string(),
+        };
+
+        let mut extension = None;
+        if let Some(res) = path.extension() {
+            extension = Some(
+                res.to_str()
+                    .expect("failed to encode file name")
+                    .to_string(),
+            );
+        }
+
+        Self {
+            name,
+            path: PathBuf::from(path),
+            extension,
+            file_type,
+        }
+    }
+
     pub fn file_name(&self) -> &str {
         self.path.file_name().and_then(OsStr::to_str).unwrap()
     }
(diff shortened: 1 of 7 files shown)


dvvvvvv:
This PR is fix for #319

src/meta/name.rs
```diff
@@ -111,7 +111,7 @@ impl Name {
     }
 
     pub fn extension(&self) -> Option<&str> {
-        self.extension.as_ref().map(|string| string.as_str())
+        self.extension.as_deref()
```
meain, line 114: This seems to fail on 1.33

dvvvvvv: reverted. This commit comes from the warning provided by clippy. The Rust compiler included in the github action should also be downgraded to 1.33.

meain: Hmm, let me look into it.

src/meta/name.rs
```diff
@@ -256,7 +304,7 @@ mod test {
 
     #[test]
     fn test_order_impl_is_case_insensitive() {
-        let path_1 = Path::new("AAAA");
+        let path_1 = Path::new("/AAAA");
```
meain, line 307: Any reason why you changed this?

dvvvvvv: There is no reason. I was concerned that this path is a relative path unlike other paths.

src/meta/name.rs
meain: Hey, sorry for the late reply. I am pretty busy with work recently.
I think we will the min supported version to 1.34 as lscolors already did it https://github.com/sharkdp/lscolors/commit/936e518d6c3a26a1ddc812d8d40e542b916888b1 

Could you revert to the code and just bump the rust version in travis to 1.34
https://github.com/Peltoche/lsd/blob/88e4133e16eee821cb3c9cc2c84aa7f995e22ed2/.travis.yml#L44

dvvvvvv: as_deref() is supported since 1.40. it doesn't need to be revert. The Trevis Rust version bumpup will be conducted as a separate pull request.

meain: Oh, my bad. I was hoping we could get rid o f the clippy warnings.

meain:
@dvvvvvv , looks like there is some conflicts with master.

Issue #72, opened 2018-12-13
Broken symbolic links


kuviman:
Consider a directory with broken symlink:

```
$ mkdir temp && cd temp
$ ln -s `pwd`/a `pwd`/b
```

## Expected behavior

Classic ls shows symlink, and, if you enable color, it will show that it is broken.

```
$ /bin/ls -la
total 8
drwxr-xr-x 2 kuviman kuviman 4096 Dec 13 22:52 .
drwxr-xr-x 9 kuviman kuviman 4096 Dec 13 22:51 ..
lrwxrwxrwx 1 kuviman kuviman   25 Dec 13 22:52 b -> /home/kuviman/temp/a
```

## Actual behavior

But `lsd` just throws an error

```
$ lsd -la
cannot access './b': No such file or directory (os error 2)
```

It doesn't crash as before(#2), but still would be better if broken symlinks were highlighted red like in `ls`

Peltoche:
Hi @kuviman,

Thanks for the report. It's curious because I don't reproduce the error on my laptop (Arch Linux x86).
```sh
mkdir temp && cd temp
 tmp  temp  %  ⎈ prodild  prod  ln -s `pwd`/a `pwd`/b
 tmp  temp  %  ⎈ prodild  prod  l
lrwxrwxrwx  peltoche  users   -  now ago    b ⇒ /tmp/temp/a
```
Can you give the result of `lsd --version` please?

alaymari:
Could not reproduce @kuviman's error on 0.8.0 version.

Agree with him/her that the symlink display color is misleading. On my terminal, the link shows up in cyan color (hope that is the correct one) when `lsd` is used, leading one to presume that the link is correct. `lsd -l` shows the target in red, indicating that it is dangling.

kuviman:
Right, I was using old version, sorry, its fine on 0.9.0

AndydeCleyre:
I am experiencing this on Arch with `lsd 0.16.0`. It's not a problem when listing the current directory contents with no files specified in the command, but is if any files (or files via patterns interpreted by the shell) are provided.

First, GNU `ls` behavior:
```bash
$ ls -l
total 0
lrwxrwxrwx 1 andy andy 19 Nov 12 16:37 anotherlink -> 'no file here either'
lrwxrwxrwx 1 andy andy 16 Nov 12 16:38 rellink -> ./nofilerelative
[...]

PR #341, opened 2020-04-05
delete fs canonicalize for showing broken softlink without error

zwpaper:
fix https://github.com/Peltoche/lsd/issues/72

I have checked:
1.  it could report file not existed for not existed files without the `canonicalize`
2. deleting `canonicalize` not broken any tests

And if we add the `canonicalize`, https://github.com/Peltoche/lsd/issues/72 happened

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -27,7 +27,7 @@ fn main() {
         }
     }
 
-    let var = std::env::var_os("SHELL_COMPLETIONS_DIR").or(std::env::var_os("OUT_DIR"));
+    let var = std::env::var_os("SHELL_COMPLETIONS_DIR").or_else(|| std::env::var_os("OUT_DIR"));
     let outdir = match var {
         None => return,
         Some(outdir) => outdir,
--- a/src/app.rs
+++ b/src/app.rs
@@ -220,9 +220,8 @@ pub fn build() -> App<'static, 'static> {
 }
 
 fn validate_date_argument(arg: String) -> Result<(), String> {
-    use std::error::Error;
     if arg.starts_with('+') {
-        validate_time_format(&arg).map_err(|err| err.description().to_string())
+        validate_time_format(&arg).map_err(|err| err.to_string())
     } else if &arg == "date" || &arg == "relative" {
         Result::Ok(())
     } else {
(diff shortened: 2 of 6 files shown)


zwpaper:
BTW, I found `print_error` not appending newline,

maybe this should also be fixed?

image

meain:
@zwpaper Yeah, I think it would be better for the error to have a newline.

zwpaper:
@meain I was trying to fix the `print_error`, but macro is beyond my ability, 
could you give me some advice on the improvement?

```
            let res = handle.write_all(std::format!($($arg)*).as_bytes());
```

this printed without newline, how could I fix this?

meain:
@zwpaper I would go about just adding a `\n` in `print_error` call. That would be the better way in my opinion. Because this macro could be reused even when we don't need `\n`. 
```rust
print_error!("cannot access '{}': {}\n", path.display(), err);
```

----

Just so that you know, all that is happening in that is that we are expanding everything that was inside `print_error!` call into `std.format`. So if we had to change and add a `\n` we will have to either capture the first arg separately and add `\n` to it. Another option would be to have the `std.format` captured in a variable and use another `std.format` to append `\n` to it.

zwpaper:
@meain I added a `newline` for each print_error, there should be a `newline` for the most cases, If there is any duplicated found later, I will fix it.
And I have changed the error message to match the `ls` command like
image

---
BTW, Maybe an icon for the error message should be more `lsd style`?
I will fire another PR and discuss for it.

meain:
@zwpaper I personally don't think `lsd: ` is a good idea. It does not add anything useful. I think we will stick with the current format or at least `lsd: cannot access '{}': ...`

Looks like there is not much consistency in how anyone does this.




One thing I did notice is that `lsd` is the only one that expands the symlink without `-l`. Maybe we should change that behavior.

zwpaper:
IMO, the error message explained the error, making the `cannot access` a little bit duplicated.
just like there is only one of the other three commands printing the `cannot access`

agree with the symlink expanding part.

meain:
Hmm, I guess that makes sense. We will stick to `lsd: ` then.

meain:
Looks like travis messed up something, I have restarted the build. Once that passes I will merge the PR.

PR #346, opened 2020-04-11
Do not expand sylink in grid display

meain:
Symlinked file was getting shown without `-l` flag, which does not seem to be how it is generally done.


meain:
For the old behavior you could use `lsd -l --blocks name`

Issue #192, opened 2019-04-26
Support hyperlink rendering


pothos:
The `--hyperlink` option for `ls`¹ is allows to open a document/folder with the default (GUI) application by using Ctrl+click in various terminals.
It would be nice to have this in lsd.
See here for more info: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

¹ http://git.savannah.gnu.org/cgit/coreutils.git/commit/?id=799bac0d06cfabe9491498727308df8d1aca6d98

exa issue: https://github.com/ogham/exa/issues/396

Peltoche:
Hi @pothos

Thanks for the proposition. Personally I don't see the when this feature could be used (a filename with a link ??) but if someone needs it let's go for it. At the moment I don't have the time to write new features but I would review a PR with pleasure it you want to make one.

xenopeek:
Running `ls --hyperlink` in any VTI based terminal lets you ctrl+click any file in the `ls` output to open it with its default filetype handler (or open the link through context menu). i.e., ctrl+click a text file opens the file in your text editor, ctrl+click a directory opens the directory in your file manager and so on. Quite handy, try it out.

KSXGitHub:
I have made [a pull request](https://github.com/Peltoche/lsd/pull/354) for initial support. But I am way too tired to fix any remaining issue. <sub>Seriously, reading this codebase is like traversing a labyrinth</sub>

ppwwyyxx:
The implementation of hyperlink support in #642 does not match the recommendation in the OSC8 spec here https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda#file-uris-and-the-hostname that suggests hostname to be included in the hyperlink. This is what `ls --hyperlink` does as well.

Issue #350, opened 2020-04-17
Windows builds failing


meain:
Windows builds seem to fail after d366c169874e266b82a517249ae2479f3c7a0387

dvvvvvv:
terminal-size downgrade result
https://github.com/dvvvvvv/lsd/actions/runs/81436854
terminal-size windows test result
https://github.com/dvvvvvv/terminal-size/actions/runs/81444672

This seems to be a problem caused by terminal_size crate.
terminal_size crate affects the behavior of lsd in core.rs: 34,35.
```
#[cfg(target_os = "windows")]
let tty_available = terminal_size().is_some(); // terminal_size allows us to know if the stdout is a tty or not.
```
This is a syntax related to whether not to display the icon when there is no terminal. And as you can see from the link I wrote above, It appears that the github action has a console buffer of size 1,1. However, looking at the code, this seems to be a bug with terminal_size crate. 

Anyway, the test shows the following error message.
```
thread 'test_list_almost_all_populated_directory' panicked at 'Unexpected stdout, failed var.is_match(one
two
$)
└── var as str:  one
 two
```
This shows that pattern matching has failed because of the icon.

There are several solutions.

1. downgrade terminal-size crate.
2. change args in tests.

meain:
Well, I guess technically `1,1` passes for `some()`.

Since we cannot really control flags, I think we can force `classic` in the tests.
I guess we should also have a test with `--icons always` and `--color always` in there too.

dvvvvvv:
I suspect that the latest version of terminal-size will work completely opposite to 0.1.8.  So first of all, it would be a good idea to check if it works normally in Windows. If it doesn't work as I thought, lowering the version would be a better option.  If not, it would be better to do it according to @meain

dvvvvvv:
I checked it. terminal-size 0.1.11 doesn't seem to work as I thought.
I'll leave the results for each version below.

master build with terminal-size 0.1.8:
```
PS C:\Users\smlim2\src\lsd> .\target\release\lsd.exe C:\Users\smlim2\AppData\
 Local   LocalLow   Roaming
```
master build with terminal-size 0.1.11:
```
PS C:\Users\smlim2\src\lsd> .\target\release\lsd.exe C:\Users\smlim2\AppData\
Local
LocalLow
Roaming
```
It would be better to downgrade the terminal-sized version until the terminal-sized behavior is properly modified.
[...]

PR #354, opened 2020-04-18
Hyperlink

KSXGitHub:
Fix https://github.com/Peltoche/lsd/issues/192

## Preview

preview: `lsd --hyperlink=always -1`

## Remaining Issues

It works well with `-1` flag, but it prints a [broken list](https://user-images.githubusercontent.com/11488886/79633337-f725a500-818e-11ea-923b-ef747b5b2fcf.png) without it. I can guess this is due to width of filenames not being calculated correctly: It did not strip away ANSI symbols first. I will not try fixing it as reading the source code alone had sucked the energy out of me.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -202,6 +202,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "idna"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "matches 0.1.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "unicode-bidi 0.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "unicode-normalization 0.1.12 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "ignore"
 version = "0.4.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


KSXGitHub:
I don't know why the tests fail on Windows, or if it was because of my doing, I don't have Windows, I can't fix it by myself.

meain:
Don't worry about the windows builds for now, it is a known issue and is being tracked in https://github.com/Peltoche/lsd/issues/350

meain:
The actual length needs to be computed https://github.com/Peltoche/lsd/blob/d77fe699b5d610fbb9a9d4838c230ae3a6df04f4/src/display.rs#L266 here and we need to make sure the url length does not count in.

KSXGitHub:
I think it is easier to get invisible width than to get visible one.

meain:
@KSXGitHub yeah, the function is pretty much computing the invisible width and subtracting it from the total one.

KSXGitHub:
@meain Actually, why compute visible width at all? Why don't we use length of filenames directly?

meain:
This function is actually used to compute the length of all the items inside lsd, not just for filenames.

Actually, it is an option that we can pass along the "true length" of each item along with the coloured value.

KSXGitHub:
@meain Perhaps instead of trying to fix `get_visible_width`, I think we should invest our time in implementing "true length" for less error-prone and faster code. I may work on this when I have the time, but if someone else wants to do it, then it would be great.

PR #355, opened 2020-04-18
Using rust version on style job

dvvvvvv:
clippy corrects to use the new rust std API as the rust version goes up, so we need to apply a minimal rust version to the style job as well.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -23,7 +23,7 @@ jobs:
     - name: Install `rust` toolchain
       uses: actions-rs/toolchain@v1
       with:
-        toolchain: stable
+        toolchain: ${{ env.RUST_MIN_SRV }}
         override: true
         profile: minimal # minimal component installation (ie, no documentation)
         components: rustfmt, clippy
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -6,7 +6,7 @@ env:
   PROJECT_NAME: lsd
   PROJECT_DESC: "An ls command with a lot of pretty colors."
   PROJECT_AUTH: "Peltoche <peltoche@halium.fr>"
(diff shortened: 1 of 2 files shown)


meain:
Hey, I don't think we should bump down the version of clippy. I think we will just add something so that clippy will ignore that specific error on that line on https://github.com/Peltoche/lsd/pull/328.

dvvvvvv:
@meain I disagree with that opinion.

1. Factors outside the repository determine whether or not the test was successful.
2. Whenever we update the targeted Rust version, we should review the ignored warnings.

meain:
Hey, I was thinking about it and I was wondering as to why we would have to support older version of rust to begin with. At the end of the day, this is a user facing tool, not really a library, so why not just build it on top of latest stable rust?
@Peltoche

Peltoche:
> Hey, I was thinking about it and I was wondering as to why we would have to support older version of rust to begin with. At the end of the day, this is a user facing tool, not really a library, so why not just build it on top of latest stable rust?
> @Peltoche any thoghts?

In one side the rust language is secure enough to failed at build time and avoid any surprises at runtime if the rustc version is incorrect so I'm kind of agree with you. On the other side I think it would make it easier for the developers to have and explicit error if they build with the invalid rust version. 

So for me we should keep the minimum rusts versions but keep it really short, like forcing the current "default" rustc version and that's it.

meain:
@Peltoche That sounds great. The issue came up from https://github.com/Peltoche/lsd/pull/328#discussion_r407042836 as `clippy` was complaining about a piece of code. `clippy` was expecting current stable but we had a min supported of `1.34`.

From what @dvvvvvv said, we need at least `1.40` for the code that `clippy` suggests to work. So we could probably bump min version to at least `1.40` here or even `1.43` (latest). Maybe we could just trail `rustc` a few versions for behind for min rust unless we get into something like this.

PR #358, opened 2020-04-22
Some more per file icons

### LINKED ISSUE #357: Some of my icons are wrong
Some of my icons are wrong, and I'm not sure if lsd is mapping incorrectly, or if my patched font has something off (sorry, I'm not sure how to troubleshoot further).

FWIW, most icons work fine (directories, python files, plain files, markdown, images, tarballs). The ones described below are the only exceptions I've found so far.

Screenshot:
2020-04-21T10:29:51,727920152+02:00

Copy-paste (so you know exactly which unicode points `lsd` is showing):
```
➜ lsd docker-compose.yml poetry.lock uwsgi.ini
  docker-compose.yml    poetry.lock    uwsgi.ini
```

The middle one is obviously broken.
The other two just don't match (OpenSUSE icon for docker, and Windows 10 logo for a config file).

I'm using Operator Mono, patched with nerd fonts.

WhyNotHugo:
In case icons don't render for you on the browser, the [cheatsheet](https://www.nerdfonts.com/cheat-sheet) is handy.

Fixes #357

### DIFF
--- a/src/icon.rs
+++ b/src/icon.rs
@@ -236,7 +236,7 @@ impl Icons {
         m.insert("ico", "\u{f1c5}"); // ""
         m.insert("image", "\u{f1c5}"); // ""
         m.insert("iml", "\u{e7b5}"); // ""
-        m.insert("ini", "\u{f17a}"); // ""
+        m.insert("ini", "\u{e615}"); // ""
         m.insert("ipynb", "\u{e606}"); // ""
         m.insert("jar", "\u{e204}"); // ""
         m.insert("java", "\u{e204}"); // ""
--- a/src/icon.rs
+++ b/src/icon.rs
@@ -152,6 +152,7 @@ impl Icons {
         m.insert("rubydoc", "\u{e73b}"); // ""
         m.insert("yarn.lock", "\u{e718}"); // ""
         m.insert("Cargo.lock", "\u{e7a8}"); // ""
+        m.insert("poetry.lock", "\u{f023}"); // ""
 
         m
     }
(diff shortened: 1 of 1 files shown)


meain:
@WhyNotHugo Hey, I think for `poetry.lock` maybe we should go with python icon than a lock icon as `Cargo.tom` and `yarn.lock` seems to have rust and js icons.

WhyNotHugo:
My initial thought was going for the Python icon. However, since this file is generally in a file full of _actual_ python files, having a different icon made it stand out easier. It's also not actually python but toml.

I can change it to python if you feel they must align though.

Peltoche:
Hi @WhyNotHugo , thanks for this PR.

For what I have understood, this file is a configuration file so I'm for the "fe615" icon too.

meain:
@WhyNotHugo I personally don't use icons a lot, so I may not be the right person to decide, but I think it would be better to keep it consistent with the current set of icons. 

Also repos that add icons for nerdtree and ranger seems to have language icons for their package manager lockfiles.

meain:
@Peltoche I was talking about the `poetry.lock` file. Since we have a rust icons for `Cargo.lock` and node icon for `yarn.lock` I was thinking we should go with python icons for `poetry.lock`. 

[Poetry](https://github.com/python-poetry/poetry) is a kind of a package manager for python. A bit more than that, but essentially a package manager.

Peltoche:
Erf... for this one I have no ideas but I'm agree with you @meain about the consistency.

WhyNotHugo:
It feels weird to have the python icon for the lockfile because:
- It just makes it harder to spot between python files (the idea of different icons based on types is to spot them easier).
- It's not really python, rather toml.

Given that consistency is key here, how would you feel about changing the lock file icons for cargo/yarn to follow suit?

0jdxt:
I think `.lock` files should be the lock icon. Any project itself should have files specific to the language anyway and so the lock would indicate "not to be touched" across projects/languages

WhyNotHugo:
@0jdxt Done!

WhyNotHugo:
Thanks!

PR #361, opened 2020-04-28
Update terminal_size dependency to v0.1.12

eminence:
This fixes an embarassing bug in terminal_size where it wouldn't work at
all on Windows.

For more details see https://github.com/eminence/terminal-size/pull/19


eminence:
The clippy failures were not related to this change, but I pushed a commit anyway that cleans that up

PR #371, opened 2020-05-12
doc: Add FreeBSD installation instructions

andoriyu:
I've recently submitted port to FreeBSD's ports tree and my PR was merged today. That means from now on this application can be installed from binaries on FreeBSD.

On the side note: I can setup CI for FreeBSD if you wan. Let me know and i will submit a PR.


meain:
Thanks a lot.

Yeah, CI would be nice. If you could submit that as another PR, I would be happy to merge that in.

PR #379, opened 2020-05-27
added music icon for opus filetype

nabakolu:
Really bothered me, that mp3 has music icon but opus doesnt.
Added music icon for opus filetype.


src/icon.rs
```diff
@@ -261,6 +261,7 @@ impl Icons {
         m.insert("mp4", "\u{f03d}"); // ""
         m.insert("mustache", "\u{e60f}"); // ""
         m.insert("npmignore", "\u{e71e}"); // ""
+		m.insert("opus", "\u{f001}"); // ""
```
meain, line 264: Could you fix the spacing, after that we are good to go.

nabakolu: Ah sorry, changed it to spaces now

PR #383, opened 2020-06-09
Add case-insensitive matching of known filenames and extensions

poita66:
Although this achieves what I set out to do, I fully expect that there's a more idiomatic way of doing it, so I'm expecting changes to be requested

The main reason for this change is that I had a dockerfile that had a capital 'd' and therefore it didn't match anything in the known filename hash set. I figured that the same logic should apply to extensions, too

Feel free to reject this, too, as I know it's an edge case and some would argue that a dockerfile with a capital 'd' doesn't deserve to have a nice icon haha

### DIFF
--- a/src/icon.rs
+++ b/src/icon.rs
@@ -82,13 +82,13 @@ impl Icons {
             .icons_by_name
             .get(String::from(name.file_name()).to_lowercase().as_str())
         {
-            // Use the known names.
+            // Use the known lower-case names.
             icon
         } else if let Some(icon) = self
             .icons_by_name
             .get(String::from(name.file_name()).to_uppercase().as_str())
         {
-            // Use the known names.
+            // Use the known upper-case names.
             icon
         } else if let Some(icon) = name
             .extension()
@@ -100,13 +100,13 @@ impl Icons {
             self.icons_by_extension
                 .get(String::from(extension).to_lowercase().as_str())
         }) {
-            // Use the known extensions.
+            // Use the known lower-case extensions.
             icon
         } else if let Some(icon) = name.extension().and_then(|extension| {
             self.icons_by_extension
                 .get(String::from(extension).to_uppercase().as_str())
         }) {
-            // Use the known extensions.
+            // Use the known upper-case extensions.
             icon
         } else {
             // Use the default icons.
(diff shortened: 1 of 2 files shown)


meain:
@poita66 I am not sure if this is a good idea. For stuff like `Dockerfile`, it kinda makes sense, but I am not sure if ones like `NODE_MODULES` or `.GIT` should be treated same as their lower case counterpart.

It makes sense for extensions I guess.

poita66:
@meain 
If it was for anything more important than choosing an icon, I'd totally agree. However, because that's all it affects, I figure that a NODE_MODULES folder should get a nice icon too.
It would also then mimic the behavior of VS Code (and I expect most editors), which puts an icon on files no matter the case-ness.

meain:
Hey, I think we could probably merge this in. As of now there is only `.Trash` in the actual list that has a capital letter.
So I think you could change that to `.trash` and then only check with `input.tolower()`.

This should also fix https://github.com/Peltoche/lsd/issues/420

poita66:
Hey! Will do. Sorry about the late reply, I must've missed the notification email.

meain:
Hey, there seems to be some extra files that have been commited like `.Trash`, `.trash`, and `readme.md`.
Could you remove those. Also, could you rebase your work on top of current master.

poita66:
Done!

src/icon.rs
```diff
@@ -75,13 +75,16 @@ impl Icons {
...
+        {
             // Use the known names.
             icon
-        } else if let Some(icon) = name
-            .extension()
-            .and_then(|extension| self.icons_by_extension.get(extension))
-        {
+        } else if let Some(icon) = name.extension().and_then(|extension| {
+            self.icons_by_extension
+                .get(String::from(extension).to_lowercase().as_str())
```
meain, line 86:
```suggestion
                .get(extension.to_lowercase().as_str())
```

meain (changes requested):
@poita66 Could you also add a test to ensure both upper and lowercae filenames match.

poita66:
@meain Tests have been added and your suggested changes have been incorporated and tested. Thanks for your help and sorry about the delay :sweat_smile:

meain (approved):
LGTM, thanks :)

Issue #335, opened 2020-03-11
output of filenames with special characters (\n, \t ...) is not escaped


ChristianS99:
## Pre
touch tab$'\t'tab
touch newline$'\n'newline

## Expected behavior
not sure, ls displays this like `'newline'$'\n''newline'  'tab'$'\t''tab'`, but that's not very nice, too.

## Actual behavior
🗋  newline
newline  🗋  tab tab
displaying the tab/newline character as actuall tab/nl is not nice. it should be somehow escaped, maybe take ls as example, but something nicer is also nice :)

meain:
The simplest option would be to have something like

```diff
diff --git a/src/meta/name.rs b/src/meta/name.rs
index f734e6b..58b2a5f 100644
--- a/src/meta/name.rs
+++ b/src/meta/name.rs
@@ -43,7 +43,7 @@ impl Name {
         let mut content = String::with_capacity(icon.len() + self.name.len() + 3 /* spaces */);
 
         content += icon.as_str();
-        content += &self.name;
+        content += &self.name.escape_default().collect::<String>();
         content
     }
 ```

This will just print out:

```
newline\nnewline tab\ttab
```

One possible solution would be to color `\n` but doing that with the current code is not that simple but maybe we can go with this for now?

ChristianS99:
would be a start, definitely better than current.

meain:
I would agree.

meain:
Closing this as the primary issue is now fixed.

Issue #325, opened 2020-02-04
"Icon?" file on MacOS hides some files


ho-oto:
On MacOS, some applications (e.g., Dropbox, GoogleDrive) make hidden "Icon?" file to change the folder icon seen from Finder.app. These hidden files hides some files from the result of `lsd`.

I show the result on Catalina 10.15.3.

```
% lsd -V
lsd 0.16.0
% ls
Icon?   aaa     zzz
% ls -l
total 808
-rw-r--r--@ 1 hootooo  staff  0  2  4 17:42 Icon?
-rw-r--r--  1 hootooo  staff  0  2  4 17:22 aaa
-rw-r--r--  1 hootooo  staff  0  2  4 17:22 zzz
% lsd
    zzz  Icon
% lsd -l
.rw-r--r-- hootooo staff 0 B Tue Feb  4 17:22:10 2020   aaa
.rw-r--r-- hootooo staff 0 B Tue Feb  4 17:42:36 2020   Icon
.rw-r--r-- hootooo staff 0 B Tue Feb  4 17:22:12 2020   zzz
```

meain:
This should be fixed now.
Feel free to reopen if this issue still persists.

See https://github.com/Peltoche/lsd/issues/335 for any further improvement on how it is handled.

PR #384, opened 2020-06-13
Escape special chars in filename like `\t` and `\n`

meain:
- Basic fix for https://github.com/Peltoche/lsd/issues/335
- Might also fix this https://github.com/Peltoche/lsd/issues/325

Before it would put a tab instead of `\t` in the filename. Same for `\n`.

### DIFF
--- a/src/meta/name.rs
+++ b/src/meta/name.rs
@@ -80,6 +80,28 @@ impl Name {
             .collect()
     }
 
+    pub fn escape(&self, string: &str) -> String {
+        if string
+            .chars()
+            .all(|c| c >= 0x20 as char && c != 0x7f as char)
+        {
+            string.to_string()
+        } else {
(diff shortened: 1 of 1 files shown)


meain:
Looks like this actually breaks how emoji and other weird stuff gets rendered.

These stuff fail to render. Seems like git just prints escaped stuff, but gnu ls prints the actual chars.

```
("Ｈｅｌｌｏ,ｗｏｒｌｄ!", 24),
("ASCII1234-_", 13),
("File with space", 17),
("制作样本。", 12),
("日本語", 8),
("샘플은 무료로 드리겠습니다", 28),
("👩🐩", 6),
("🔬", 4),
```

meain:
`exa` seemed to have a pretty good solution for this, so I kinda stole it and adapted for our use case.

Issue #392, opened 2020-07-18
Permissions Error with opt folder


joesan:
## Expected behavior
Doing a ls -la should list me all the files in the directory, in this case the /opt directory

## Actual behavior
```
joesan@joesan-InfinityBook-S-14-v5:/opt/softwares$ ls -l
cannot access '.': Permission denied (os error 13)
```

I guess there is an issue already relating to this. I could not infer the proposed solution to that issue.

https://github.com/Peltoche/lsdcannot%20access%20'.':%20Permission%20denied%20(os%20error%2013)/issues/79

Should I uninstall and install a newer version to fix this? If so which package is that for Ubuntu 20.04?

zwpaper:
Hi @joesan 

May I ask how do u install `lsd`, if snap, maybe u should uninstall it and try a deb from
https://github.com/Peltoche/lsd#on-ubuntu
or a binary from
https://github.com/Peltoche/lsd#from-binaries

the problem may be caused by `snap` and it's not easy to fix

Peltoche:
hello @joesan,

The `snap` packages are not more maintained due to too much issues with the jail system. I try without success to delete the available binaries since quite some time. As proposed by @zwpaper you should install it for the .deb.

joesan:
How do I uninstall from the snap package and reinstall it from debian? Any guide or Wiki on how to do this would be helpful!

PR #401, opened 2020-08-15
:art: delete install via snap, add binary guide, add badges

zwpaper:
snap is not more maintained:
https://github.com/Peltoche/lsd/issues/79
https://github.com/Peltoche/lsd/issues/392

badges preview
image

I changed the license to blue from orange, as I checked some trending repos and blue is a more common choice.


meain:
@zwpaper What do you think about having a script that we can host in this repo which they can curl and pipe to bash which would pick the latest version and correct architecture and pull down the file and install it?
Not sure how this would work on Windows though.

zwpaper:
@meain happy that we have the same idea.
actually I have done some research for this

In my previous idea, we could just use the github assets.
GitHub offered a static link for the latest release, we could omit the version in assets name, and make latest static. 

For example 

```sh
curl -sSL github.com/Peltoche/lsd/releases/latest/download/lsd-$OS-$ARCH.tar.gz | tar zxvf - -C /usr/local/bin/
```

zwpaper:
BTW, I nearly not use Windows for dev at all, Windows may be a tough problem for me

meain:
I am not really sure about the Windows situation either. Don't have access to a Windows machine.

Btw, we probably should be using `uname -s` and `uname -m` as that should work pretty well on mac as well.

README.md
```diff
@@ -118,7 +118,12 @@ cargo install --git https://github.com/Peltoche/lsd.git --branch master
 
 ### From Binaries
 
-The [release page](https://github.com/Peltoche/lsd/releases) includes precompiled binaries for Linux and macOS.
+The [release page](https://github.com/Peltoche/lsd/releases) includes precompiled binaries for Linux and macOS, you can download the binary install lsd via:
+
+``` sh
+ tar zxvf lsd-0.17.0-x86_64-apple-darwin.tar.gz # adapt version number and architecture
+ mv lsd-0.17.0-x86_64-apple-darwin/lsd /usr/local/bin/
```
meain, line 125: I think it is better to keep it how it used to be.
For those who know what to do, we probably should not add anything and for those who don't know what they are doing this might only cause confusion if they are using anything other than macOS.

zwpaper: I will take your `install.sh` advice, and add one here.

this is how I will do:

1. change GitHub workflow to release assets with name ``lsd-`uname -s`-`uname-m`-`libc`.tar.gz`` and dir ``lsd-`uname -s`-`uname-m`-`libc``` (remove the version)
2. with `1.` we have the static link for latest release: ``https://github.com/Peltoche/lsd/releases/download/latest/lsd-`uname -s`-`uname -m`-`libc`.tar.gz``
3. add `install.sh` to detect libc, download the latest binary and move lsd to `/usr/local/bin`

what do u think about it?

meain: I am not sure if changing the release asset names might cause issues with scripts that might be in place for package manager submissions. Not sure how that might work out.
I think it would be better to just handle it in the `install.sh` script.

zwpaper: We have to update the download link in `install.sh` after every release if there is a version in asset's name

meain: I am guessing we should be able to grep it out of `https://api.github.com/repos/Peltoche/lsd/releases/latest`.

I have been using https://github.com/meain/homebrew-meain/blob/master/update for some time in one of my repos and it seem to work pretty fine.

meain:
We could probably drop https://github.com/Peltoche/lsd/blob/master/.snapcraft.yaml since we are not having `snap` anymore.

README.md
zwpaper: sounds reasonable and keeps a minimal change, let's do it this way

zwpaper:
sorry for the delay, I was overloaded last week...
this PR should be ok to go now

meain:
No worries, thanks.

Issue #394, opened 2020-07-26
Add support/parameter for "natural sort"


thomas-merz:
## Expected behavior
The "lsd" command should have a nice parameter wich naturally sorts numbers within text.

## Actual behavior
The "ls" command has a nice parameter, but "lsd" hasn't:  
-v     natural sort of (version) numbers within text

zwpaper:
I could help to implement this, 
but first I would like to make sure how this work because I was confused about the description `(version) numbers within text`

what's the meaning of the `(version)`?

AFAIK, `-v` should sort by file name in natural order, like 
1. numbers before letters
2. number order in the natural order, like 1 2 11 21

thomas-merz:
Hi @zwpaper , great, that you could help :-)
Normal "ls" has "-v" option which ends up in what you said in both points:
(but I don't know why a capital letter should be sorted before a lower case and I won't mind/care about this)
```
root@ubuntu-cx11-02:~/temp/1# command ls -lv #h #--time-style=+%Y-%m-%d_%H:%M:%S
total 0
-rw-r--r-- 1 root root 0 Aug 12 09:12 1
-rw-r--r-- 1 root root 0 Aug 12 09:12 2
-rw-r--r-- 1 root root 0 Aug 12 09:12 3
-rw-r--r-- 1 root root 0 Aug 12 09:12 4
-rw-r--r-- 1 root root 0 Aug 12 09:12 10
-rw-r--r-- 1 root root 0 Aug 12 09:12 11
-rw-r--r-- 1 root root 0 Aug 12 09:12 12
-rw-r--r-- 1 root root 0 Aug 12 09:12 20
-rw-r--r-- 1 root root 0 Aug 12 09:12 21
-rw-r--r-- 1 root root 0 Aug 12 09:12 A
-rw-r--r-- 1 root root 0 Aug 12 09:12 B
-rw-r--r-- 1 root root 0 Aug 12 09:12 C
-rw-r--r-- 1 root root 0 Aug 12 09:12 a
-rw-r--r-- 1 root root 0 Aug 12 09:12 b
-rw-r--r-- 1 root root 0 Aug 12 09:12 c
```
instead of this without "-v":
```
root@ubuntu-cx11-02:~/temp/1# command ls -l
total 0
-rw-r--r-- 1 root root 0 Aug 12 09:12 1
-rw-r--r-- 1 root root 0 Aug 12 09:12 10
-rw-r--r-- 1 root root 0 Aug 12 09:12 11
[...]

PR #414, opened 2020-09-18
sort: :sparkles: add version sort

zwpaper:
fix https://github.com/Peltoche/lsd/issues/394

### DIFF
--- a/src/sort.rs
+++ b/src/sort.rs
@@ -1,5 +1,6 @@
 use crate::flags::{DirOrderFlag, Flags, SortFlag, SortOrder};
 use crate::meta::Meta;
+use human_sort::compare;
 use std::cmp::Ordering;
 
 pub type SortFn = fn(&Meta, &Meta) -> Ordering;
@@ -19,6 +20,7 @@ pub fn assemble_sorters(flags: &Flags) -> Vec<(SortOrder, SortFn)> {
         SortFlag::Name => by_name,
         SortFlag::Size => by_size,
         SortFlag::Time => by_date,
+        SortFlag::Version => by_version,
         SortFlag::Extension => by_extension,
     };
     sorters.push((flags.sort_order, other_sort));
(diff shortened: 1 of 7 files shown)


src/sort.rs
```diff
@@ -56,6 +58,10 @@ fn by_date(a: &Meta, b: &Meta) -> Ordering {
     b.date.cmp(&a.date).then(a.name.cmp(&b.name))
 }
 
+fn by_version(a: &Meta, b: &Meta) -> Ordering {
+    compare(&a.name.name, &b.name.name).then(compare(&b.name.name, &a.name.name))
```
meain, line 62: Any specific reason why you have a `.then` chained onto it?

tests/integration.rs
```diff
@@ -260,6 +260,50 @@ fn test_show_folder_of_symlink_for_long_multi() {
         .stdout(predicate::str::contains("link/:"));
 }
 
+#[test]
+fn test_version_sort() {
```
meain, line 264: I think `sort.rs` would be a better place for this test than in `integration.rs`.
I guess we will keep this test here anyways, should help with an integration test for sort. But could you have a similar test in `sort.rs`?

meain: Maybe also add one with items like this:
```
0.3.7  0.11.5
```
Probably not necessary, but just to be safe.

src/sort.rs
zwpaper: my bad, this is a mistake. updated

tests/integration.rs
zwpaper: added, IMO, `integration.rs` is where we did the app functional testing, `lsd` as the entry.

of cause, some more package unit tests would always be better.

zwpaper:
I read the comment in ur email, it's a good idea to add some negative case, so I added some more file names as the test.

origin ls:

```
[vagrant@arch ~]$ ls -l -v
total 0
-rw-r--r-- 1 vagrant vagrant   0 Sep 19 09:30 0.2
-rw-r--r-- 1 vagrant vagrant   0 Sep 19 09:30 0.3.7
-rw-r--r-- 1 vagrant vagrant   0 Sep 19 09:30 0.11
-rw-r--r-- 1 vagrant vagrant   0 Sep 19 09:30 0.11.5
-rw-r--r-- 1 vagrant vagrant   0 Sep 19 09:30 1
-rw-r--r-- 1 vagrant vagrant   0 Sep 19 09:30 2
-rw-r--r-- 1 vagrant vagrant   0 Sep 19 09:30 11
-rw-r--r-- 1 vagrant vagrant   0 Sep 19 09:30 11a
-rw-r--r-- 1 vagrant vagrant   0 Sep 19 09:30 22
```

PR #416, opened 2020-09-25
Add a minimal emoji theme

damz:
Add a minimal emoji-based theme. Many terminal implementations support color emojis out of the box nowadays. This, for example, is on a Gnome Terminal on Fedora 32, with no configuration whatsoever:

Screenshot from 2020-09-25 15-30-23

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -48,10 +48,11 @@ pub fn build() -> App<'static, 'static> {
                 .long("icon-theme")
                 .possible_value("fancy")
                 .possible_value("unicode")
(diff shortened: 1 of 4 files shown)


meain:
Hi @damz , neat idea.
We have a [pr for reading config from a file](https://github.com/Peltoche/lsd/pull/317) coming up and I think it would be better to ship more color and icons themes as config options. This PR in itself will not let you setup extra colors or icons but we will be adding that in the future. Thanks a lot for opening the PR though.

PR #418, opened 2020-09-28
Fix #408

### LINKED ISSUE #408: FR: show fraction in file sizes only for values less than 100
From a certain size on the fraction is not that important anymore when viewed in `default` mode (which is the equivalent to ls' `human-readable`).

So maybe values less than `100` should be displayed with a fraction, equal/larger values as integer.

But those sizes should be displayed always as a number with fraction:
```
.rw-r--r-- me user 2020-04-04 12:36:29 5.9 MB  IMG_20200308_162055.jpg
.rw-r--r-- me user 2020-04-04 12:36:33   6 MB  IMG_20200308_162058.jpg
```
-> better to see `6.0 MB` instead of integer value.

LHY-iS-Learning:
Create a macro to handle trailing zero, only add `.0` when file size unit is KB or above and numeric size is smaller than 100.0;

### DIFF
--- a/src/meta/size.rs
+++ b/src/meta/size.rs
@@ -25,6 +25,12 @@ impl<'a> From<&'a Metadata> for Size {
     }
 }
 
+macro_rules! trailing_zero {
+    ($number: expr) => {
+        format!("{0:.1$}", $number, if $number < 100.0 { 1 } else { 0 })
+    };
+}
+
 impl Size {
     pub fn new(bytes: u64) -> Self {
         Self { bytes }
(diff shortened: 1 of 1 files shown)


meain:
Hey, I am really sorry. I completely overlooked the fact that this would break consistency with `ls` which even though we are not completely adhering to is something that we would like to keep to and extend.
So sorry for taking up your time. :(

zwpaper:
Hi @LHY-iS-Learning , the issue is updated, sorry for the mistaken closed, can you update the threshold to 10 and then we can try merging this again!

meain:
@zwpaper Thanks for looking into it. I rushed through it and missed it.

meain:
@LHY-iS-Learning , thanks a lot for the PR.

We probably don't need to use a macro for this. A function should be more than enough in this case and will be much more readable. Also, IMO something like `format_size` would be a better name for the function.

meain (approved):
LGTM, thank you! :)

PR #424, opened 2020-10-01
Adds sort flag

### LINKED ISSUE #415: Add --sort flag
Now that we have all the sort options, it would be good to have `--sort` option as well.

From gnu ls:

```
      --sort=WORD            sort by WORD instead of name: none (-U), size (-S),
                             time (-t), version (-v), extension (-X)
```

tumdum:
Fixes #415

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -152,6 +152,7 @@ pub fn build() -> App<'static, 'static> {
                 .overrides_with("sizesort")
                 .overrides_with("extensionsort")
                 .overrides_with("versionsort")
+                .overrides_with("sort")
                 .multiple(true)
                 .help("Sort by time modified"),
         )
@@ -162,6 +163,7 @@ pub fn build() -> App<'static, 'static> {
                 .overrides_with("timesort")
                 .overrides_with("extensionsort")
                 .overrides_with("versionsort")
+                .overrides_with("sort")
                 .multiple(true)
                 .help("Sort by size"),
         )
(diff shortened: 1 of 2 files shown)


src/app.rs
```diff
@@ -177,8 +180,22 @@ pub fn build() -> App<'static, 'static> {
...
+            Arg::with_name("sort")
+                .long("sort")
+                .multiple(true)
+                .possible_values(&["size", "time", "version", "extension"])
+                .takes_value(true)
+                .value_name("ORDER")
+                .overrides_with("timesort")
+                .overrides_with("sizesort")
+                .overrides_with("extensionsort")
+                .overrides_with("sort")
```
meain, line 196:
```suggestion
                .overrides_with("versionsort")
```

src/app.rs
```diff
@@ -177,8 +180,22 @@ pub fn build() -> App<'static, 'static> {
...
                 .help("Natural sort of (version) numbers within text"),
         )
+        .arg(
+            Arg::with_name("sort")
+                .long("sort")
+                .multiple(true)
+                .possible_values(&["size", "time", "version", "extension"])
+                .takes_value(true)
+                .value_name("ORDER")
```
meain, line 192: I think `FIELD` would be a better name here. Or we could use `WORD` which is what gnu ls uses.

meain (changes requested):
Hi, thanks for the PR.
Just had few minor changes.

src/app.rs
tumdum: Whichever you prefer :)

meain: I think we will go with `WORD`. Now that I think about it, `version` does not really fit into `FIELD`.

tumdum:
Looks to me like a random failure of ci but I have no way to retrigger.

meain:
I have triggered a rerun.

Issue #387, opened 2020-06-24
lsd panics


MaT1g3R:
## Expected behavior

ls output:

```shell
❯ /bin/ls -la
total 4408
drwx------  32 umi umi    4096 Jun 23 19:14  .
drwxr-xr-x 112 umi umi    4096 Jun 24 12:32  ..
-rwxr-xr-x   1 umi umi  735908 Sep 16  2019 '5a35.ico'$'\247\375''='$'\356\324\373\206\362\300''^m'$'\330'
-rwxr-xr-x   1 umi umi  141853 Sep  4  2019  684e
-rwxr-xr-x   1 umi umi  106170 Jun 23  2019  8465
drwxr-xr-x   5 umi umi    4096 Jun 24 00:44  appcache
drwxr-xr-x   4 umi umi    4096 Apr 28  2018  bin
-rwxr-xr-x   1 umi umi   11384 Feb 10  2014  bin_steamdeps.py
-rw-r--r--   1 umi umi 2899552 Jun 23 19:14  bootstrap.tar.xz
drwxr-xr-x   6 umi umi    4096 Jun  3 18:57  clientui
drwxr-xr-x   5 umi umi    4096 May 29 19:36  compatibilitytools.d
drwxr-xr-x   5 umi umi    4096 Jun 24 00:23  config
drwxr-xr-x   4 umi umi    4096 Apr 29 18:32  controller_base
-rw-r--r--   1 umi umi       0 Jun 23 19:14  .crash
drwxr-xr-x   2 umi umi   32768 Jun 24 00:23  depotcache
drwxr-xr-x   2 umi umi    4096 Dec  8  2019  fontconfig
-rwxr-xr-x   1 umi umi     146 Jun  6  2019  fossilize_engine_filters.json
drwxr-xr-x   2 umi umi   16384 Jun  6 14:23  friends
drwxr-xr-x   3 umi umi   69632 Dec 16  2019  graphics
drwxr-xr-x   2 umi umi    4096 Jul 18  2019  html5app
-rw-r--r--   1 umi umi     736 Jan 19 19:53  installscriptevalutor_log.txt
drwxr-xr-x   2 umi umi    4096 Jun 23 19:14  legacycompat
drwx------   2 umi umi    4096 Jun  6 14:23  linux32
drwxr-xr-x   2 umi umi    4096 Jun  6 14:23  linux64
drwxr-xr-x   2 umi umi    4096 Jun 23 20:44  logs
[...]

PR #425, opened 2020-10-02
skip extension if no ext found

zwpaper:
fix https://github.com/Peltoche/lsd/issues/387

I also found the output is a little bit different like ls, as `lsd` will display the UTF-8 in file names, It seems like work as expect IMO. If necessary we can discuss this in a new issue. @meain 

ls:
```
[root@b78708e2fc75 tmp]# \ls
'5a35.ico'$'\247\375''='$'\356\324\373\206\362\300''^m'$'\330'
```

lsd:

```
[root@b78708e2fc75 tmp]# lsd
 5a35.ico��=������^m�
```

FYI:
the utf8 code in the filename is reserved for utf-16(https://en.wikipedia.org/wiki/UTF-16#U+D800_to_U+DFFF), and rust will check this by default.

Following the rust way, we display the reserved chars as �

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -362,6 +362,42 @@ fn test_version_sort_overwrite_by_sizesort() {
         .stdout(predicate::str::is_match("11\n2\n$").unwrap());
 }
 
+#[test]
+#[cfg(target_os = "linux")]
+fn test_bad_utf_8_extension() {
+    let tmp = tempdir();
+    Command::new("python3")
+        .arg("-c")
+        .arg(format!(
+            r#"import pathlib; pathlib.Path('{}/bad.extension\udca7\udcfd').touch()"#,
+            tmp.path().to_str().unwrap()
+        ))
+        .output()
+        .expect("failed to create file");
+    cmd()
+        .arg(tmp.path())
+        .assert()
+        .stdout(predicate::str::is_match("bad.extension\u{fffd}\u{fffd}\n$").unwrap());
+}
+
+#[test]
+#[cfg(target_os = "linux")]
+fn test_bad_utf_8_name() {
+    let tmp = tempdir();
+    Command::new("python3")
+        .arg("-c")
+        .arg(format!(
+            r#"import pathlib; pathlib.Path('{}/bad-name\udca7\udcfd.ext').touch()"#,
+            tmp.path().to_str().unwrap()
+        ))
+        .output()
+        .expect("failed to create file");
+    cmd()
+        .arg(tmp.path())
+        .assert()
+        .stdout(predicate::str::is_match("bad-name\u{fffd}\u{fffd}.ext\n$").unwrap());
+}
+
 fn cmd() -> Command {
     Command::cargo_bin(env!("CARGO_PKG_NAME")).unwrap()
 }
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -364,46 +364,41 @@ fn test_version_sort_overwrite_by_sizesort() {
 
 #[cfg(test)]
 #[cfg(target_os = "linux")]
-macro_rules! bad_utf8 {
-    ($tmp:expr, $e:expr) => {{
-        let mut fname = format!($e, $tmp.path().to_str().unwrap())
-            .as_bytes()
-            .to_vec();
-        fname.push(0xa7);
-        fname.push(0xfd);
-        fname
-    }};
-    ($tmp:expr, $e:expr, $v:expr) => {{
-        let mut fname = bad_utf8!($tmp, $e);
-        fname.append($v.as_bytes().to_vec().as_mut());
-        fname
-    }};
+fn bad_utf8(tmp: &std::path::Path, pre: &str, suf: &str) -> String {
+    let mut fname = format!("{}/{}", tmp.display(), pre).into_bytes();
+    fname.reserve(2 + suf.len());
+    fname.push(0xa7);
+    fname.push(0xfd);
+    fname.extend(suf.as_bytes());
+    unsafe { String::from_utf8_unchecked(fname) }
 }
 
 #[test]
 #[cfg(target_os = "linux")]
 fn test_bad_utf_8_extension() {
+    use std::fs::File;
     let tmp = tempdir();
-    let fname = unsafe { String::from_utf8_unchecked(bad_utf8!(tmp, "{}/bad.extension")) };
-    std::fs::File::create(fname).expect("failed to create file");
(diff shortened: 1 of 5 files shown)


zwpaper:
BTW, I have tried to find a way to escape the invalid utf-8 code in the filename, but no luck.

`path.file_name()` returns an `Option<&OsStr>`, If invalid utf-8 contained, it was `None`, I can not find a way to get the name, can u give me some tips for getting it? even we did not need to print it in lsd, just out of curiosity.

Peltoche:
Hello @zwpaper, thanks for this PR.

Can you add a test with the filename from the issue? This would allow us to confirm that the code work as expected and to avoid any future regression.

zwpaper:
Hi @Peltoche , glad that you have time for lsd again 😃

I have tried to add the test for this, but this happened only if there is an invalid utf-8 char,
and rust keeps preventing me create the invalid utf-8 char.

`unicode escape must not be a surrogate` always happens after invalid utf-8 chars added

tests/integration.rs
```diff
@@ -362,6 +362,25 @@ fn test_version_sort_overwrite_by_sizesort() {
...
+    Command::new("sh")
+        .arg("-c")
+        .arg("mkdir -p /tmp/bad.extension/")
+        .output()
+        .expect("failed to create dir");
+    Command::new("python3")
+        .arg("-c")
+        .arg(r#"import pathlib; pathlib.Path('/tmp/bad.extension/bad.extension\udca7\udcfd').touch()"#)
+        .output()
+        .expect("failed to create file");
```
0jdxt, line 377: Hey man it may be best to avoid having python as a dependency, this can be done in sh with
```shell
$ sh -c "touch $(printf 'bad.extension\xdca7\xdcfd')"
```
```suggestion
    Command::new("sh")
    	.arg("-c")
    	.arg("touch $(printf '/tmp/bad.extension/bad.extension\xdca7\xdcfd')")
    	.output()
    	.expect("failed to create file");
```

0jdxt (changes requested):

tests/integration.rs
0jdxt: and this would also fix the CI since it's not set up to install python

zwpaper: I don't want the python dep either, but I can not find a way to create the invalid utf-8 name even use your code.🤣

I have tried the solution like this, The output of the `sh -c touch` is iliteral `\xdca7`, not an utf-8 code.

invalid utf-8 code will be treated as error if using double quote.

```
---- test_bad_utf_8_extension stdout ----
thread 'test_bad_utf_8_extension' panicked at 'Unexpected stdout, failed var.is_match(bad.extension��
$)
└── var as str: bad.extension\xdca7\xdcfd


command=`"/root/lsd/target/debug/lsd" "/tmp/.tmpRDHlAb"`
code=0
stdout=```bad.extension\xdca7\xdcfd
```
stderr=``````
',
```

Comment (zwpaper) [reply tests/integration.rs]: I was still trying to find a way to create the invalid name by `shell`, but no luck, python is my last choice...

File: tests/integration.rs [hunk 2]
```diff
@@ -362,6 +362,42 @@ fn test_version_sort_overwrite_by_sizesort() {
...
+fn test_bad_utf_8_extension() {
+    let tmp = tempdir();
+    Command::new("python3")
+        .arg("-c")
+        .arg(format!(
+            r#"import pathlib; pathlib.Path('{}/bad.extension\udca7\udcfd').touch()"#,
+            tmp.path().to_str().unwrap()
+        ))
+        .output()
+        .expect("failed to create file");
```
0jdxt, line 376: try with double slashes
```suggestion
    Command::new("sh")
        .arg("-c")
        .arg("touch $(printf 'bad.extension\\xdca7\\xdcfd')")
        .output()
        .unwrap();
```

zwpaper: have your try to run the code?

double backslashes fix the rust error report, but created a file named `'bad.extension\xdca7\xdcfd'` not the utf 8 code.

I have done several tries before using python

0jdxt: Yeah if i did a project with main.rs:
```rust
use std::process::Command;

fn main() {
    Command::new("sh")
        .arg("-c")
        .arg("touch $(printf 'bad.extension\\xdca7\\xdcfd')")
        .output()
        .unwrap();
}
```
and when i run `ls` i get `'bad.extension'$'\334''a7'$'\334''fd'`.

zwpaper: that's really weird, I can't reproduce your result
image

meain: @zwpaper Are you running from within a linux docker image on a mac?

Is it possible that that could affect the result?

zwpaper: that's also my concern, so I was setting up a VM in fusion, big sur broke my VirtualBox

zwpaper: the same result on debian 10 in fusion vm.

image

@0jdxt could you help me to reproduce this?

0jdxt: I'll submit a PR just to see if it works with the CI.

zwpaper:
@0jdxt thanks for the improvement

@meain This should be ready to go

0jdxt:
I submitted i small pr to tidy my code a bit

zwpaper:
@meain  this is ready to go, what's your idea?

meain:
Hey, I don't think there is a need to explicitly mention this in the readme under description as this is a really small edge case and an implementation detail.

zwpaper:
@meain I added because this affects the user experience.

Users will get unexpected � when an invalid utf-8 appeared in the file name, so we should tell them about this.

meain:
Yes, but the chances of user having an invalid utf-8 char in the filename is very unlikely.
Considering that, having this message in the project description does not seem right.

Plus, � seems like something that the user would be able to figure out to be a "non-standard" entry.
I am expecting only a few handful of users to ever run into this and if it ever comes to that, I think we could probably add it to the FAQ section of the readme.

zwpaper:
@meain moved to FAQ

0jdxt:
hey, you somehow got rid of some of my commits? the tests now have the old macro. Did you do a rebase?

zwpaper:
@0jdxt oh, it was my bad! I forgot to pull your work, so lucky that you found it!

0jdxt:
Btw don't merge my or its wrong atm

zwpaper:
@meain sorry for the accident, this should be ready to go

meain:
Thanks @zwpaper and @0jdxt . Also thanks for filling in the CHANGELOG.md file.

Issue #295, opened 2019-11-01
Request: Enable --tree to be used with --directory-only (-d)


AndydeCleyre:
lsd `0.16.0`

## Expected behavior
```bash
# "real" tree:
tree -d
```
```
.
├── ci
└── src
    └── meta

3 directories
```
```bash
lsd --tree -d
```
```
.
├──   ci
└──   src
   └──   meta
```

## Actual behavior
```bash
lsd --tree -d
```
```
error: The argument '--tree' cannot be used with '--directory-only'

USAGE:
    lsd --blocks <blocks>... --color <color>... --date <date>... --directory-only --group-dirs <group-dirs>... --icon <icon>... --icon-theme <icon-theme>... --size <size>... --tree

For more information try --help
```
With an exit code `1`

Peltoche:
Hi @AndydeCleyre, thanks for the report.

At first glance this should be easily fixed. I look into it now.

elliebike:
I'm not sure if the expected behaviour here is valid

For instance, with `--directory-only` I get

<img width="118" alt="Screenshot 2020-03-29 at 00 15 55" src="https://user-images.githubusercontent.com/53315310/77836914-9c31fc80-7152-11ea-8c39-bbc3983e5508.png">

Going by the help string for `-d`: 

`-d, --directory-only    Display directories themselves, and not their contents`

So I guess the name of the flag is a little ambiguous, if anything. It could be taken to mean

1. Only list directories, not files
2. Only list the directory specified (which, imo, seems a little pointless?)

If 1, then it seems like `--directory-only` might need fixing (as per my example above - I do have a lot inside of `~/src`!)

Otherwise I don't think this issue is valid, though maybe the docs could do with an adjustment.

meain:
@elliebike I am not sure why this is a thing, but `-d` is supposed to display the directory that is provided. The behavior seems to be consistent with ls.

I don't think `-d` in tree has the same meaning as `-d` in ls.

AndydeCleyre:
Alright, this means I can't alias `tree` to `lsd --tree` directly. FWIW, for anyone else looking to achieve something like that, I'm now using this in my Zsh profile:

```bash
if (( $+commands[lsd] )); then
    alias ls="lsd"
    tree () {
        local depth dirsonly
        while [[ $1 == -L || $1 == -d ]]; do
            if [[ $1 == -L ]]; then
                depth=(--depth $2)
                shift 2
            fi
[...]

PR #426, opened 2020-10-02
implement --tree -d

0jdxt:
Was looking at #295 and decided to have a crack

### DIFF
--- a/src/meta/mod.rs
+++ b/src/meta/mod.rs
@@ -116,6 +116,11 @@ impl Meta {
                 }
             };
 
+            // skip files for --tree -d
+            if flags.display == Display::TreeD && !entry_meta.file_type.is_dirlike() {
+                continue;
+            }
+
             match entry_meta.recurse_into(depth - 1, &flags) {
                 Ok(content) => entry_meta.content = content,
                 Err(err) => {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for `--extensionsort` `-X` from [aldhsu](https://github.com/aldhsu)
 - Add support for `--versionsort` `-v` from [zwpaper](https://github.com/zwpaper)
 - Add support for config symlink arrow from [zwpaper](https://github.com/zwpaper) [#409](https://github.com/Peltoche/lsd/issues/409)
+- Implement `--tree -d`, analogous to `tree -d` from [0jdxt](https://github.com/0jdxt)
 ### Changed
 - Use last sort flag for sort field from [meain](https://github.com/meain)
 ### Fixed
--- a/src/app.rs
+++ b/src/app.rs
@@ -116,7 +116,6 @@ pub fn build() -> App<'static, 'static> {
                 .conflicts_with("almost-all")
                 .conflicts_with("depth")
                 .conflicts_with("recursive")
-                .conflicts_with("tree")
                 .help("Display directories themselves, and not their contents"),
         )
         .arg(
--- a/src/flags/display.rs
+++ b/src/flags/display.rs
@@ -15,6 +15,7 @@ pub enum Display {
     AlmostAll,
     DirectoryItself,
     DisplayOnlyVisible,
+    TreeD,
 }
 
 impl Display {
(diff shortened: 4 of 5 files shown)


0jdxt:
there is probably a cleaner way to do it and have the logic together, ill have a look

0jdxt:
I'm not sure why I get the clippy errors, I dont get them locally

meain:
Hey, there has been a relatively big change in how flags are managed due to the introduction of config file. Could you rebase this on top of current master.

0jdxt:
i can only reproduce the clippy error on 1.43.1, not 1.46

0jdxt:
@meain this is ready if you're happy with it

zwpaper:
maybe we should do a rebase or squash to remove the unnecessary commits before merging this?

it would be more complicated for others to rebase on top of many small commit and got an unclear commit history

0jdxt:
Yeah just squash it

src/core.rs
```diff
@@ -132,7 +132,7 @@ impl Core {
     }
 
     fn display(&self, metas: &[Meta]) {
-        let output = if self.flags.layout == Layout::Tree {
+        let output = if let Layout::Tree { .. } = self.flags.layout {
```
meain, line 135: Something like this would be better:
```rust
matches!(self.flags.layout, Layout::Tree {..})
```

meain (changes requested):
Hey, I think there is a small issue with the implementation.
If we have `tree` as layout in the config but pass `-d` flag as cli arg, it is not getting treated as `--tree -d`.

I think it would be better to not touch how `Layout::Tree` or `dir_only` is parsed, but just do the check for if `dir_only` is set when displaying in tree mode. As in no need to add `dir_only` to `Layout::Tree`.

0jdxt:
done!

src/meta/mod.rs
```diff
@@ -54,14 +54,17 @@ impl Meta {
...
         }
 
-        if flags.display == Display::DirectoryItself {
+        if flags.display == Display::DirectoryItself && flags.layout != Layout::Tree {
             return Ok(None);
         }
 
         match self.file_type {
             FileType::Directory { .. } => (),
             FileType::SymLink { is_dir: true } => {
```
meain, line 63: Hey, I just realized that this might cause issues if we have something like below:
```
mkdir one
ln -s one one/two
```

This will go on a endless loop if we have dereference enabled. I am not really certain if we have to handle it though. I took a look at gnu ls and even thought it does not have `--tree` or `-R` it does not seem to handle the case were we symlink two itmes to each other and just error out with os error.
```
ln -s one two
ln -s two one
```

Do you have any other suggestion? If not, I think we can stick to this as of now hoping that users would be aware of `-L` use with `-R` or `--tree`. Also, when I tested it, it seemed to stop at few levels deep. I am not sure if this is an OS level restriction, but at least `lsd` does not go bonkers and eat up all the cpu if this happens.

meain (commented):
Could you also add a few more tests, just to avoid any future regressions.

- for --tree in config and -d in cli arg
- for --tree -l -d
- for case were symlink dir (do not show content without `-L`)

src/meta/mod.rs
meain: Just realized that gnu ls has `-R`.

```
/tmp/tree-test/one:
two
gls: /tmp/tree-test/one/two: not listing already-listed directory
```

This is the error that we get. Maybe we should do this check.

0jdxt: Yes I think we should take from `ls` in general so ideally people could replace `ls` with `lsd` and behaviour is as expected maybe with some exceptions.

I will look into some way of "remembering" displayed files, and at how `ls` does it. Also, `tree` gives the following which I will try to recreate. Also perhaps we could add a "X directories, Y files" line too for `lsd --tree`

```bash
$ tree
.
└── one
    └── two -> ../one  [recursive, not followed]

2 directories, 0 files
```

meain: Btw, as of now it looks like `--tree` follows the symlink, but `tree` does not seem to follow symlinks by default. I think it would be better for us to do it that way too and only follow symlinks if there is `-L` with tree.

meain: "recursive" might not be the proper term. Like gnu ls, something like "already-listed" would be better as we probably need to make it so that all previously visited items are not listed, not just recursive (tracing if something is recursive might be unnecessary work).

meain: Hey, the changes to incorporate "remembering" visited paths, I guess, can be a separate  PR if that works for you. I can merge this in for now if you could add the rest of the tests as well.

meain:
Closing in favor of https://github.com/Peltoche/lsd/pull/456 . Thanks @0jdxt for getting it started.

PR #428, opened 2020-10-03
Refactor: Convert dirorder flag to sort flag.

aldhsu:
This refactor will make it easier to add multisorting eg. something like `--sort=-directory,name,extension,-size`.
This refactor removes special case dirorder flags and combines sort order and sort field into one property.


aldhsu:
Woops didn't realise there has been a big refactor for on the sorting flags will have to rework this significantly.

Issue #409, opened 2020-09-08
Feature request: allow symlink arrow to be "->" instead of "⇒"


planet36:
On my system, the rightwards double arrow character is not visually distinctive from a bold hyphen.

Please allow (maybe with an option or env var) the symlink arrow to be "->" instead of "⇒".

zwpaper:
Add a option for rendering this would be an easy solution, what do u think? @meain

meain:
Yeah, I have a few concerns though:

- does the flag only affect `⇒` or all future non ascii chars if any
- what should the flag be (--only-ascii ?)
- should the flag be `=>` or `->` as some fonts have `-` rendered below middle

zwpaper:
@meain actually, there is a more complicated solution in my mind, as u mentioned maybe the flag may effect all chars.

How about we add themes for all those customized requests, like this issue, https://github.com/Peltoche/lsd/pull/333, and the description in README.

we can use a theme file to describe the `=>`, icons, also colors.

meain:
Yeah, that sounds good. Config would be a better place for this.

planet36:
It works great.  Thanks!

PR #436, opened 2020-10-24
create config for symlink arrow

zwpaper:
fix https://github.com/Peltoche/lsd/issues/409

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for using a config file [kmoschcau](https://github.com/kmoschcau)
 - Add support for `--extensionsort` `-X` from [aldhsu](https://github.com/aldhsu)
 - Add support for `--versionsort` `-v` from [zwpaper](https://github.com/zwpaper)
+- Add support for config symlink arrow from [zwpaper](https://github.com/zwpaper) [#409](https://github.com/Peltoche/lsd/issues/409)
 ### Changed
 - Use last sort flag for sort field from [meain](https://github.com/meain)
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -260,11 +260,11 @@ no-symlink: false
 # Possible values: false, true
 total-size: false
 
-# == Styles ==
+# == Styling ==
 # How lsd looks like
 #
 # symlink-arrow: chars both ascii and utf8
-styles:
+styling:
   symlink-arrow: ⇒
 ```
 
--- a/README.md
+++ b/README.md
@@ -260,12 +260,9 @@ no-symlink: false
 # Possible values: false, true
 total-size: false
 
-# == Styling ==
-# How lsd looks like
-#
-# symlink-arrow: chars both ascii and utf8
-styling:
-  symlink-arrow: ⇒
+# == Symlink arrow ==
+# Specifies how the symlink arrow display, chars in both ascii and utf8
+symlink-arrow: ⇒
 ```
 
 ## External Configurations
--- a/README.md
+++ b/README.md
@@ -259,6 +259,13 @@ no-symlink: false
 # Whether to display the total size of directories.
 # Possible values: false, true
 total-size: false
+
+# == Styles ==
+# How lsd looks like
+#
+# symlink-arrow: chars both ascii and utf8
+styles:
+  symlink-arrow: ⇒
 ```
 
 ## External Configurations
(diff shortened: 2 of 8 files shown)


meain:
The code part looks good. @kmoschcau do you have any other suggestion for the name of the group that will contain this(`styles`).
I am assuming in future we could add things like space between icon and name here as well.

kmoschcau:
@meain Yeah I don't think "styles" is a good way of naming it. At style to me is something like a set of settings, that you switch as a whole. It might be easier to comprehend, if you just keep this key to setting up icons/symbols and name it either "symbols" or "icons". That way you can reuse it for file type icons as well. Spacing can then be under its own key.

Also in terms of code organization, yes I understand to some degree that one might want to put this one setting into the symlinks module, but for future proofing it would be better to put it into its own icons (or similar) module.

meain:
I am also not sure if `icons` should be what this should go under, as that right now it has a relation to icons that are used for display with filenames. `icons` in itself would have 3 categories. `general` (ones for folder, generic file, and symlink), `by_name`, `by_extension`.

What I see here are mostly a bunch of unrelated configs (symlink-arrow, icon-spacing). What are you guys' opinion on dropping it at root level?

zwpaper:
IMO, root-level options would make it simpler, but as the number of options grows, the config file may be a little bit messy, this is I would like to group it up like your previous comment, adding the number of space between icon and filename here.

`styles` may not be the best name, what I am trying to express is: this is a group of settings about the output styles of lsd

zwpaper:
as code organization part, @kmoschcau you are right, I should have put it to a new mod.

when I first implement the symlink arrow, I treat it as an option for symlink, but later the `styles` option came up but I did not move the code.

If we reach an agreement, the `styles(maybe a better name)` module can be used to put those options.

kmoschcau:
@meain but why not group this together with the filename icons? To me this sound pretty sensible. Just to be clear, I would of course *not* put the spacing options in there as well. Those would go somewhere else.

@zwpaper after thinking some more about this, especially when putting options like spacing in this category, what about naming it "layout"?

meain:
@kmoschcau 

Are you suggesting something like this? I am assuming we will need some classification for general, filename and filetype anyways. Or should `symlink-arrow` be under some other heading?

```yml
icons:
    symlink-arrow: <>
    general:
        - directory: <>
        - symlink: <>
        - broken-symlink: <>
        - file: <>
    filename:
        - Dockerfile: <>
        - node_modules: <>
    extension:
        - rs: <>
        - py: <>
```

zwpaper:
The symlink arrow is a little bit special, it can be UTF-8 chars like the `icon` we talk about, but it serves not like the icon in front of filenames indicating the file types.
That's why I am not adding it to the `icon` set but a `styles`, it's something like the `space between icon and filename` IMO.

But as you guys said, it also can be an icon sitting between the filename and target file, I am also ok with this, but prefer leaving the `icon` be `the icon before filename`

meain:
@zwpaper Yeah, that was what I was thinking too. `icons` would better be reserved for file icons. But not sure if `styles` is the best "group" for it though.

zwpaper:
style is a little bit too general, but I can't figure out a better one.

If we can, I would also like to replace it.

`layout` fits the space option, but not the arrow one

kmoschcau:
How about a root key called "styling" and then have the "icons", "color-theme" (or similar), "layout" and all other miscellaneous keys like the "symlink-arrow" under it?

zwpaper:
@meain changed according to discuss, good to go now

meain:
Hey @zwpaper , sorry that this taking a while. I have been super busy with work stuff recently. Just wanted to make sure we get the config right as that will stick forever.

Here is my argument for having this in the root than under a heading.
From what I can tell, pretty much everything else other that what we have now comes under `styling`, and when that is the only heading, it seems unnecessary. There are two options which we can take.
Either separate it out a bit more, also have the currently supported items under `general` or `basic`. 
Another option would be to have most thing directly in root. We will sill have nested when it is needed, but things like this can go into the root.
Btw, since this is `yaml`, users can have comments if they wish to separate out parts of the config.

Just an example:

```yaml
# general settings
layout: grid
color:
  when: always
icons:
  when: never

# display colors
color-theme:
  file: #fafafa
  directory: #fce3e3
  symlink: #dddddd
  borken-symlink: #ff0000

# custom icons
icons:
  filetype:
    rs: "<>"
    go: "<>"

# other settings
symlink-arrow: "->"
icon-spacing: 2
```

This to me actually looks better than nesting everything other than the top items in `styling`.
What do you think?

zwpaper:
> everything else other that what we have now comes under `styling`

this convinced me，looks like root level make sense too.

PS: never mind the delay 😁

kmoschcau:
@meain you are making a good point, I agree.

zwpaper:
@meain should be ready to go now!

PR #438, opened 2020-10-25
Fix date test bug on Windows

0jdxt:
had a localisation issue with creating dates for testing since they are created in powershell using "MM/DD/YYYY" format and my machine is localised to "DD/MM/YYY" so keeping them ISO8601 will solve this i.e. "YYYY-MM-DD"

### DIFF
--- a/tests/integration.rs
+++ b/tests/integration.rs
@@ -363,7 +363,7 @@ fn test_version_sort_overwrite_by_sizesort() {
 }
 
 #[cfg(test)]
-#[cfg(not(target_os = "mac"))]
+#[cfg(target_os = "linux")]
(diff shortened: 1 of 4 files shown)


tests/integration.rs
```diff
@@ -362,7 +359,6 @@ fn test_version_sort_overwrite_by_sizesort() {
         .stdout(predicate::str::is_match("11\n2\n$").unwrap());
 }
 
-#[cfg(test)]
```
meain, line 365: Any reason why you removed this?

0jdxt: According to the Rust book, anything in the `tests` directory is only compiled for tests anyway so effectively the whole file is `#[cfg(test)]` and makes any use of it in the `tests` directory redundant.

PR #441, opened 2020-10-27
Speed up

0jdxt:
Whilst working on my other PR I found `--tree` to be quite slow compared to `tree` and so analysed `cargo flamegraph` results resulting in reducing unnecessary allocations and system calls, producing a significant speed gain for the `--tree` and `-R` options.

<details>
<summary>Original flamegraph: (--tree)</summary>

flamegraph-slow

</details>

<details>
<summary>After optimisations: (--tree)</summary>

flamegraph-fast

</details>

<details>
<summary>After optimisations: (-R)</summary>

flamegraph-r

</details>

Originally, the main bottleneck was system calls in order to retrieve `uid` and `gid` properties and the following processing into strings when the `-R` and `--tree` options do not need this information so now on unix, the information is lazy loaded.

Now, in the optimised `--tree` graph, we see the main bottleneck now is creating the rest of the `Meta` struct and that the sorting and display isn't too shabby. In the optimised `-R` graph, we see the displaying and sorting of information into a grid is the biggest bottleneck for this option. This would indicate for future optimisations, some information and processing may need to be lazy loaded for the `Meta` struct or perhaps simply the algorithms need optimising, on top of improving the grid/tree display performance.

Nevertheless, overall this PR has managed to make the following speed gains, tested only on my linux x86_64 machine with hyperfine, compared with native equivalents `ls -R` and `tree`:

~ 5 000 directories, 56 000 files, 27G
| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `lsd --tree` | 2.374 ± 0.016 | 2.350 | 2.400 | 6.89 ± 0.15 |
| `lsd --tree` (opt.) | 0.537 ± 0.007 | 0.528 | 0.550 | 1.56 ± 0.04 |
| `tree` | 0.345 ± 0.007 | 0.338 | 0.360 | 1.00 |

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `lsd -R` | 2.495 ± 0.027 | 2.458 | 2.535 | 19.77 ± 0.33 |
| `lsd -R` (opt.) | 0.649 ± 0.011 | 0.639 | 0.668 | 5.14 ± 0.11 |
| `ls -R` | 0.126 ± 0.002 | 0.123 | 0.130 | 1.00 |


src/meta/name.rs
```diff
@@ -22,14 +21,18 @@ pub struct Name {
 
 impl Name {
     pub fn new(path: &Path, file_type: FileType) -> Self {
-        let name = match path.file_name() {
-            Some(name) => name.to_string_lossy().to_string(),
-            None => path.to_string_lossy().to_string(),
+        let full = path.to_string_lossy().to_string();
+
+        let name = if let Some(idx) = full.rfind('/') {
```
meain, line 26: This will cause issues in Windows. Even if it did not, this would be a bad idea.

Cargo.toml
```diff
@@ -35,6 +35,7 @@ wild = "2.0.*"
 globset = "0.4.*"
 xdg = "2.1.*"
 yaml-rust = "0.4.*"
+chrono = "0.4.19"
```
meain, line 38: Does `chrono` help all that much with speed up? Otherwise it would be better to not add an extra dependency.

src/meta/name.rs
```diff
@@ -22,14 +21,18 @@ pub struct Name {
...
+            full.split_at(idx + 1).1.into()
+        } else {
+            full
         };
 
-        let extension = path
-            .extension()
-            .map(|ext| ext.to_string_lossy().to_string());
+        let extension = match name.rfind('.') {
```
meain, line 32: This too, probably not a good idea. Better let the language do this.
We will not be aware of all the edge cases related to extensions and it is really easy to mess things up.

src/meta/symlink.rs
```diff
@@ -1,54 +1,30 @@
...
-            };
-        }
-
-        Self {
-            target: None,
-            valid: false,
-        }
+impl From<&Path> for SymLink {
+    fn from(path: &Path) -> Self {
+        path.read_link()
```
meain, line 13: This would break us being able to display symlink targets for broken symlinks.

meain (changes requested):
Thanks a lot for the PR. I was not able to get this big a difference but a huge improvement however.

src/meta/symlink.rs
0jdxt: this is fixed now

0jdxt (commented):
Hey man! you're right these changes weren't necessary, I reverted some changes and tidied some expressions in some places. This commit now speeds things up in 2 ways:

String allocation
---
- Use `Cow` instead of an owned `String` in the `Name` struct, so allocating a new `String` only happens  if and when needed.
- Use `to_ascii_lowercase` instead of `to_lowercase` since it is faster and the keys we have in `Icons` only contain, and probably will always only contain, ascii letters.

Syscalls
---
On linux fetching the file info for all the files was quite slow, especially when the information isnt needed (like `-R` and `--tree`). The syscalls have been deferred to when needed in rendering the `Meta` entry.

Cargo.toml
0jdxt: Regarding this, `chrono` is already a dependency due to `chrono_humanize` so this way, we get fir of the extra `time` dependency

data-man:
What do you think about [jwalk](https://lib.rs/crates/jwalk)?

0jdxt:
> What do you think about [jwalk](https://lib.rs/crates/jwalk)?

I tried to re-implement it with `walkdir` and it is ideal for walking directory trees whereas in `lsd` we are mostly listing the contents of a single given directory, not the tree unless `-R` or `--tree`. When I benchmarked, the original `lsd` implementation was faster which I assume is due to the overhead of setting up  threads and preparing to walk the directories in `walkdir` even though typically we only need to get the entries in 1 directory.

If `jwalk` is anything like this I would presume it would have the same effect - just adding overhead. At the moment, the walking and filtering is rather efficient since we only descend into directories if needed. I think the most useful speed gains now would be the general workings of the `Meta` struct and its values (trying to avoid allocations), and the displaying of entries into a grid/tree.

0jdxt:
> What do you think about [jwalk](https://lib.rs/crates/jwalk)?

Upon looking into it I will see if it helps specifically for these recursive descents I'll try it

SuperSandro2000:
@0jdxt Do you have planed to finish this PR and resolve the merge conflicts?

0jdxt:
> @0jdxt Do you have planed to finish this PR and resolve the merge conflicts?

Hi! I've been busy so feel free. It's just such a mess I feel i should be responsible for cleaning it up. Alas, this PR could be merged soon, I'm just a bit concerned on how I've changed the flow of the program a bit and the structure of the program is getting quite complex. Either way the repo needs a tidy up at least with some comments describing the flow of the program so its easier to follow, especially for those who want to contribute to the repo and is reading the code for the first time.

meain:
Hey @0jdxt , you will have to rebase on master instead of merging from master.

0jdxt:
summary: Some code is merely tidied. Replaced `String` of `Name` with `Cow`, reducing unnecessary allocs. `Meta` keeps raw `Metadata` to avoid second syscall for the metadata during colouring/rendering. Keep `Owner` as `u64` until rendered. Change icons to `char`s and use the faster `FxHashMap`, delegates separator flag to `Name::render`. `Name::escape` function was O(2n), reduced to O(n). 'Name::Ord' now uses `to_ascii_lowercase` instead, increasing speed of comparisons.

src/display.rs
```diff
@@ -352,18 +339,21 @@ mod tests {
     use crate::meta::{FileType, Name};
     use std::path::Path;
 
+    // TODO: tempdir
```
meain, line 342: Ahh, need this TODO to be taken care of.

src/icon.rs
```diff
@@ -1,353 +1,377 @@
...
+                    }?;
+                    self.icons_by_shebang(command)
+                })
+                .cloned()
+                .or(Some(self.default_file_icon)),
+        }
+    }
 
-        format!("{}{}", icon, self.icon_separator)
+    fn icons_by_shebang(&self, cmd: &str) -> Option<&char> {
```
meain, line 103: This is a great addition.

src/meta/name.rs
```diff
@@ -14,108 +15,96 @@ pub enum DisplayOption<'a> {
 
 #[derive(Clone, Debug, Eq)]
 pub struct Name {
-    pub name: String,
+    name: Option<String>,
```
meain, line 18: Probably drop a comment on the use of `name` here. Might get confusion later otherwise.

meain (changes requested):
Thanks a lot, this has a great deal of code cleanup as well.
Just took a quick look at the changes.

tests/integration.rs
```diff
@@ -425,6 +425,8 @@ fn test_tree() {
     cmd()
         .arg(tmp.path())
         .arg("--tree")
+        .arg("--color")
+        .arg("never")
```
0jdxt, line 429: @meain could you have a look at this? The tests failed because of the colours so i added the option but apparently these tests were absolutely fine before. If you could perhaps try running the tests without the changes here and see if they work or not? It may be the case that I've changed something inadvertently but I can't figure it out from reading the code.

src/icon.rs
0jdxt: thanks! i  got tired of my bin folders being blank files :laughing:

meain:
Hey @0jdxt , please don't squash and force push every change. It makes it harder to review just the changes.

src/meta/name.rs
meain: From what I understand `name` is only used for `.` and `..`. I think we should probably mention it here.

tests/integration.rs
meain: I removed these and tried out in my local. It seems to work even without `--color never` in my local. It is possible that this was introduced because of the change in the lib that gives us term width. If `lsd` detects that it is not an interactive terminal, we disable colors automatically. This is done by checking the term width.

I was actually able to run this without these and it seems to be fine. (https://github.com/meain/lsd/actions/runs/685425620). Do you remember which platform it was failing for?

SuperSandro2000:
> Hey @0jdxt , please don't squash and force push every change. It makes it harder to review just the changes.

You can click on force pushed to see an interdiff.

https://github.com/Peltoche/lsd/compare/04bb7141df9226a774cb1cba91977514d20b3625..0f3eb3e459267d78207e3cfa8eca8b23b81a3fb0

0jdxt:
> Hey @0jdxt , please don't squash and force push every change. It makes it harder to review just the changes.

Hi sorry still not used to this PR stuff

tests/integration.rs
0jdxt: Strange, it's only failing locally for me. Even when reverted to `terminal_size`

0jdxt: Solved it! lsd was reading my config file which had `--color always`. Check my recent push.

meain: I think we should probably add `--ignore-config` in there.

meain:
@0jdxt Had to merge in https://github.com/Peltoche/lsd/pull/489 . You will have to rebase your current PR on top of master.
This looks mostly good to go otherwise.

0jdxt:
> @0jdxt Had to merge in #489 . You will have to rebase your PR on top of current master.
> This looks mostly good to go otherwise.

Rebased! should be good to go

src/display.rs
```diff
@@ -1,61 +1,65 @@
...
-        if let Block::Name = block {
-            index = i;
-            break;
-        }
-    }
+
+    let index = match flags.blocks.0.iter().position(|&b| b == Block::Name) {
+        Some(i) => i,
+        None => 0,
+    };
```
meain, line 43:
```suggestion
    let index = flags.blocks.0.iter().position(|&b| b == Block::Name).unwrap_or(0);
```

meain (changes requested):

meain:
I'm closing this PR for now as it has deviated quite a bit from master, but let me know if you are still interested in this and we can open it back up.

PR #443, opened 2020-11-02
using serde yaml to parse config file with type check

zwpaper:
Hi @meain , recently I was working on the symlink and theme features, finding that we were using raw rust_yaml to parse the config file.
I have tried to refactor it using `serde_yaml` so that we can parse the yaml file into a Rust struct, with args check, type check, and some other advanced. also we can isolate the config file logic inside the config_file modules.

for example, the config struct:

```rust
pub struct Config {
    pub classic: Option<bool>,
    pub blocks: Option<Sequence>,
    pub color: Option<Color>,
    pub date: Option<String>, // enum?
    pub dereference: Option<bool>,
    pub display: Option<String>, // enum?
    pub icons: Option<Icons>,
    pub ignore_globs: Option<Sequence>,
    pub indicators: Option<bool>,
    pub layout: Option<String>, // enum?
    pub recursion: Option<Recursion>,
    pub size: Option<String>, // enum?
    pub sorting: Option<Sorting>,
    pub no_symlink: Option<bool>,
    pub total_size: Option<bool>,
    pub styling: Option<Styling>,
}
```

### DIFF
--- a/src/config_file.rs
+++ b/src/config_file.rs
@@ -1,181 +1,171 @@
+use serde::Deserialize;
+use serde_yaml::Sequence;
+use std::error::Error;
 ///! This module provides methods to handle the program's config files and operations related to
 ///! this.
-use std::fs::File;
-use std::io::prelude::*;
+use std::fs;
+use std::io::BufReader;
 use std::path::PathBuf;
+use xdg::BaseDirectories;
 
 use crate::print_error;
-
-#[cfg(not(windows))]
-use xdg::BaseDirectories;
-use yaml_rust::{Yaml, YamlLoader};
 
 const CONF_DIR: &str = "lsd";
 const CONF_FILE_NAME: &str = "config";
 const YAML_LONG_EXT: &str = "yaml";
-const YAML_SHORT_EXT: &str = "yml";
 
 /// A struct to hold an optional file path [String] and an optional [Yaml], and provides methods
 /// around error handling in a config file.
-#[derive(Clone, Debug)]
+#[derive(Debug, Deserialize)]
 pub struct Config {
-    pub file: Option<String>,
-    pub yaml: Option<Yaml>,
+    pub classic: Option<bool>,
+    pub blocks: Option<Sequence>,
+    pub color: Option<Color>,
+    pub date: Option<String>, // enum?
+    pub dereference: Option<bool>,
+    pub display: Option<String>, // enum?
+    pub icons: Option<Icons>,
+    pub ignore_globs: Option<Sequence>,
+    pub indicators: Option<bool>,
+    pub layout: Option<String>, // enum?
+    pub recursion: Option<Recursion>,
+    pub size: Option<String>, // enum?
+    pub sorting: Option<Sorting>,
+    pub no_symlink: Option<bool>,
+    pub total_size: Option<bool>,
+    pub styling: Option<Styling>,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct Color {
+    pub when: String, // enum?
+}
+
+#[derive(Debug, Deserialize)]
+pub struct Icons {
+    pub when: String, // enum?
+    pub theme: String,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct Recursion {
+    pub enabled: bool,
+    pub depth: i32,
+}
+
+#[derive(Debug, Deserialize)]
+pub struct Sorting {
+    pub column: String, // enum?
+    pub reverse: bool,
+    pub dir_grouping: String, // enum?
+}
+
+#[derive(Debug, Deserialize)]
+pub struct Styling {
+    pub symlink_arrow: bool,
 }
 
 impl Config {
     /// This constructs a Config struct without a file [String] and without a [Yaml].
     pub fn with_none() -> Self {
-        Self {
-            file: None,
-            yaml: None,
-        }
+        Self::default()
     }
 
     /// This constructs a Config struct with a passed file [String] and without a [Yaml].
-    pub fn with_file(file: String) -> Self {
-        Self {
-            file: Some(file),
-            yaml: None,
+    // TODO(zhangwei) Box<Error>
+    pub fn with_file(file: String) -> Option<Self> {
+        match fs::read(&file) {
+            Ok(f) => Self::with_yaml(&String::from_utf8_lossy(&f)),
+            Err(e) => {
+                match e.kind() {
+                    std::io::ErrorKind::NotFound => {}
+                    _ => print_error!("bad config file: {}, {}\n", &file, e),
+                };
+                None
+            }
         }
     }
 
     /// This constructs a Config struct with a passed [Yaml] and without a file [String].
-    #[cfg(test)]
-    pub fn with_yaml(yaml: Yaml) -> Self {
-        Self {
-            file: None,
-            yaml: Some(yaml),
-        }
-    }
-
-    /// This tries to read a configuration file like in the XDG_BASE_DIRS specification and returns
-    /// the contents of the YAML config file.
-    pub fn read_config(name: &str) -> Self {
-        let config_file_long_path;
-        let config_file_short_path;
-        match Self::config_file_paths(name) {
-            Some((long, short)) => {
-                config_file_long_path = long;
-                config_file_short_path = short;
-            }
-            _ => return Self::with_none(),
-        }
-
-        let mut out_config;
-        let mut config_file;
-        match File::open(&config_file_long_path) {
-            Ok(result) => {
-                config_file = result;
-                out_config = Self::with_file(config_file_long_path.as_path().display().to_string());
-            }
-            Err(_) => match File::open(&config_file_short_path) {
-                Ok(result) => {
-                    config_file = result;
-                    out_config =
-                        Self::with_file(config_file_short_path.as_path().display().to_string());
-                }
-                Err(_) => return Self::with_none(),
-            },
-        }
-
-        let mut config_content = String::new();
-        if let Err(error) = config_file.read_to_string(&mut config_content) {
-            print_error!("Found a config file, but could not read it: {}", error);
-            return out_config;
-        }
-
-        match YamlLoader::load_from_str(&config_content) {
-            Ok(result) => {
-                if !result.is_empty() {
-                    out_config.yaml = Some(result[0].clone());
-                }
-                out_config
-            }
-            Err(error) => {
-                print_error!("Error parsing config: {}\n", error);
-                out_config
+    fn with_yaml(yaml: &str) -> Option<Self> {
+        match serde_yaml::from_str(yaml) {
+            Ok(c) => Some(c),
+            Err(e) => {
+                print_error!("configuration file format error, {}\n\n", e);
+                None
             }
         }
     }
 
-    /// This provides two paths for a configuration file (the first with the long yaml extension,
-    /// the second with the short yml extension), according to the XDG_BASE_DIRS specification.
+    /// This provides the path for a configuration file, according to the XDG_BASE_DIRS specification.
+    /// not checking the error because this is static
     #[cfg(not(windows))]
-    pub fn config_file_paths(name: &str) -> Option<(PathBuf, PathBuf)> {
-        let base_dirs;
-        match BaseDirectories::with_prefix(CONF_DIR) {
-            Ok(result) => base_dirs = result,
-            _ => return None,
-        }
-
-        let config_file_long_path;
-        match base_dirs.place_config_file([name, YAML_LONG_EXT].join(".")) {
-            Ok(result) => config_file_long_path = result,
-            _ => return None,
-        }
-
-        let config_file_short_path;
-        match base_dirs.place_config_file([name, YAML_SHORT_EXT].join(".")) {
-            Ok(result) => config_file_short_path = result,
-            _ => return None,
-        }
-
-        Some((config_file_long_path, config_file_short_path))
+    fn config_file_path() -> PathBuf {
+        BaseDirectories::with_prefix(CONF_DIR)
+            .unwrap()
+            .place_config_file([CONF_FILE_NAME, YAML_LONG_EXT].join("."))
+            .unwrap()
     }
 
-    /// This provides two paths for a configuration file (the first with the long yaml extension,
-    /// the second with the short yml extension) inside the %APPDATA% directory.
+    /// This provides the path for a configuration file, inside the %APPDATA% directory.
+    /// not checking the error because this is static
     #[cfg(windows)]
-    pub fn config_file_paths(name: &str) -> Option<(PathBuf, PathBuf)> {
-        let mut config_file_long_path;
-        match dirs::config_dir() {
-            Some(path) => config_file_long_path = path,
-            _ => return None,
-        }
-
-        config_file_long_path.push(CONF_DIR);
-        let mut config_file_short_path = config_file_long_path.clone();
-
-        config_file_long_path.push([name, YAML_LONG_EXT].join("."));
-        config_file_short_path.push([name, YAML_SHORT_EXT].join("."));
-
-        Some((config_file_long_path, config_file_short_path))
-    }
-
-    /// Returns whether the Config has a [Yaml].
-    pub fn has_yaml(&self) -> bool {
-        self.yaml.is_some()
-    }
-
-    /// This prints the provided warning message to stderr, prepending the executable name and the
-    /// configuration file path that likely caused the warning.
-    pub fn print_warning(&self, message: &str) {
-        print_error!(
-            "lsd: {} - {}\n",
-            self.file.as_ref().unwrap_or(&String::from("")),
-            message
-        );
-    }
-
-    /// This prints a predetermined warning message to stderr, warning about an invalid value for a
-    /// configuration element.
-    pub fn print_invalid_value_warning(&self, name: &str, value: &str) {
-        self.print_warning(&format!("Not a valid {} value: {}", name, value));
-    }
-
-    /// This prints a predetermined warning message to stderr, warning about a wrong [Yaml] data
-    /// type for a configuration value.
-    pub fn print_wrong_type_warning(&self, name: &str, type_name: &str) {
-        self.print_warning(&format!(
-            "The {} config value has to be a {}.",
-            name, type_name
-        ));
+    fn config_file_path() -> PathBuf {
+        dirs::config_dir()
+            .unwrap()
+            .join(CONF_DIR)
+            .join(CONF_FILE_NAME)
+            .set_extension(YAML_LONG_EXT)
     }
 }
 
 impl Default for Config {
     fn default() -> Self {
-        Self::read_config(CONF_FILE_NAME)
+        if let Some(c) = Self::with_file(Self::config_file_path().to_string_lossy().to_string()) {
+            c
+        } else {
+            Config {
+                classic: Some(false),
+                blocks: None,
+                color: None,
+                date: None,
+                dereference: None,
+                display: None,
+                icons: None,
+                ignore_globs: None,
+                indicators: None,
+                layout: None,
+                recursion: None,
+                size: None,
+                sorting: None,
+                no_symlink: None,
+                total_size: None,
+                styling: None,
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::Config;
+    #[test]
+    fn test_read_config_ok() {
+        let c = Config::with_yaml("classic: true").unwrap();
+        println!("{:?}", c);
+        assert!(c.classic.unwrap())
+    }
+
+    #[test]
+    fn test_read_config_bad_bool() {
+        let c = Config::with_yaml("classic: notbool");
+        println!("{:?}", c);
+        assert!(c.is_some())
+    }
+
+    #[test]
+    fn test_read_config_file_not_found() {
+        let c = Config::with_file("not-existed".to_string());
+        assert!(c.is_none())
     }
 }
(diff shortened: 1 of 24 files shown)


meain:
Yeah, I had thought of doing something similar when the initial config file support was proposed.

> From what I understood, you are looking up if keys exist, but I was thinking that it might be better to have a struct defined and serialize into that. But since this is not a rigid spec(as in items are optional) I am not sure if this is the right way to go.
https://github.com/Peltoche/lsd/pull/317#issuecomment-643787138

... but did not continue that discussion then. That said, I am not fully sure as to how it will work out with clap as we would ideally have to merge what we get from clap into the same struct before passing to `display`. We could always do what we are doing now with checking clap arguments and updating the config(but this would need us to rework flags and make config the central piece which gets updated by both config file and cli flags). Also, as @kmoschcau mentioned we will still have to write validation and serialization logic, so not a whole lot of improvement there.

I guess this does improves on the code structure, but will end up rewriting most parts of the config and flag I guess (Not sure though). @zwpaper , what are the improvements that you have in mind which this would help with?

zwpaper:
hi @meain , as you said, it may be more universal serializing into a struct, reading yaml literally and separately in different modules is not that safe and ideally, especially in a safe language Rust😃

let's continue the discussion in this PR.

1. 

> how it will work out with clap as we would ideally have to merge what we get from clap into the same struct before passing to display.

as you said, we have merged the cli and config into `Flags` currently, we could simply reuse the `Flags`, which is currently the `central piece` of cli and config, this should not be a problem.

2. 
> we will still have to write validation and serialization logic, so not a whole lot of improvement there

adding validation is about the result, not meaning that we can leave the process in a bad way.

3. 
> I guess this does improves on the code structure, but will end up rewriting most parts of the config and flag I guess (Not sure though). @zwpaper , what are the improvements that you have in mind which this would help with?

Yes, we have to rewrite much of the config logic, but this is a one time job and long term profit, after one this rewrite, we could gain more readability and maintenance. 

As the improvements, let me try to express my humble opinions and we could discuss it together.
1. we could centralize the configuration into a struct, the struct always represents the configuration file.
2. isolate the yaml logic inside config_file module, the `yaml` knowledge is only related to config_file, it should not spread to the other modules.
3. type and name check, the compiler does not know anything about `yaml["layout"]`, but it would always check `config.layout`, this is much safer.

all in all, I think this is a big rewrite but we can gain long-term profit with readability and maintenance.
also like the OOP said, maybe we should `encapsulation` the config file inside its own module, and `abstraction` the config logic outside

kmoschcau:
When this was first proposed, one of the main showstoppers for me was the requirement to give error output to the user when providing invalid values or other common mistakes in Yaml. Not all of those are currently in, but I don't think this would easily be possible when deserialising into a struct in one go.

For example what would happen when you expect a bool, but the user specified a string? Would the struct's member just be `None` in that case? You would have no clear way of knowing what went wrong then.

Yes the paths to the values in the Yaml are hard coded, but they have to be to some extent. Other than that the current deserialization handles every possible case of the value (or key) not being found, having a wrong type or an invalid value.

Other than encapsulation I don't see that much of a benefit to move all of the deserialization into one struct/namespace. But please do enlighten me if/why you think otherwise.

0jdxt:
I haven't used [`structopt`](https://docs.rs/structopt/0.3.20/structopt/) but I know its for unifying structs and cli args, might be helpful

zwpaper:
hi @0jdxt thanks for the info, but we currently working on the yaml parsing, maybe the cli args could be discussed in another PR.

Hi @kmoschcau , IMHO, configuration is one of the modules of lsd, yaml is one of the implementations of the configuration. we currently added the yaml parsing, [22 of the 37 source files](https://github.com/Peltoche/lsd/search?q=yaml) added the yaml related, updating this is really hard work, this is about the maintainace. 

but this is not the main discussion point. let's talk about the showstoppers.

this is the error output if type mismatch when serding the config into struct:

```
./target/debug/lsd                                                                                                                                                                  [12:53:09]
configuration file format error, classic: invalid type: string "a", expected a boolean at line 1 column 10

 build.rs   Cargo.lock   Cargo.toml   CHANGELOG.md   ci   CODEOWNERS   LICENSE   README.md   src   target   tests
```

1. error key, error type, position pointed out
2. return None Configuration type
3. we can use a default Configuration when configuration file error.

0jdxt:
Ah sorry I misunderstood the PR

zwpaper:
BTW, I would like to drop the short ext `yml` support, as we support both `yaml` and `yml` automatically, but always check `yml` after `yaml`, we have to document this and users have to make sure there is not a `yaml` overwriting the `yml` one.

This may be confusing, just like the README said, it should be enough to support the `yaml` one, and we could tell users that always use a `yaml`.

what is your idea? @meain

meain:
@zwpaper I don't think having `yaml` and `yml` would cause a whole lot of confusion, but I guess supporting just one might make things simpler.

kmoschcau:
Alright, that sounds good to me. Both the serialization and only keeping the long form file extension.

zwpaper:
hi @meain this should be ready to go, could you please help to review this PR

src/flags/blocks.rs
```diff
@@ -97,43 +97,47 @@ impl Blocks {
...
-                Yaml::Array(values) => Self::from_yaml_array(values, config),
-                _ => {
-                    config.print_wrong_type_warning("blocks", "array");
-                    None
+        if let Some(c) = &config.blocks {
+            let mut blocks: Vec<Block> = vec![];
+            for b in c.iter() {
+                match Block::try_from(b.as_str().unwrap()) {
+                    Ok(block) => blocks.push(block),
+                    Err(err) => print_error!("bad blocks: {}", err),
```
meain, line 109: `bad blocks: Not a valid block name: random.` seems a bit old. `Not a valid block name: random.` should be good enough.

src/main.rs
```diff
@@ -48,17 +48,19 @@ use std::path::PathBuf;
...
-                std::process::exit(0);
+            use std::io::Write;
+
+            let stderr = std::io::stderr();
+
+            {
+                let mut handle = stderr.lock();
+                // We can write on stderr, so we simply ignore the error and don't print
+                // and stop with success.
+                let res = handle.write_all(std::format!($($arg)*).as_bytes());
```
meain, line 60: We could use a newline at the end.

src/flags/size.rs
```diff
@@ -19,6 +21,23 @@ pub enum SizeFlag {
...
 
+impl SizeFlag {
+    fn from_str(value: &str) -> Option<Self> {
+        match value {
+            "default" => Some(Self::Default),
+            "short" => Some(Self::Short),
+            "bytes" => Some(Self::Bytes),
+            _ => {
+                print_error!(
+                    "size can only be one of default, short or bytes, but got {}",
```
meain, line 32: We probably will not reach here, but I think it would be better for all the error messages to be proper like capitalizing first letter, `.` at end and some basic stuff like that. You probably might have kept it to the end, just thought I would mention so as to not forget.

meain (commented):
I did a quick rundown, just a few surface level changes. Btw, just make sure all the error messages are formatted properly.

meain:
@kmoschcau would like your review on this codebase as well if you are available since you wrote the initial config logic.

src/main.rs
zwpaper: I updated this macro to use a format like `lsd: Message.\n\n`

`lsd` and two new lines are inspired by some previous error output, this shows the lsd name and adds a new line to separate the err msg and ls output.

src/flags/recursion.rs
```diff
@@ -206,38 +147,43 @@ mod test {
         assert_eq!(Some(true), Recursion::enabled_from_arg_matches(&matches));
     }
 
-    #[test]
-    fn test_enabled_from_config_none() {
-        assert_eq!(None, Recursion::enabled_from_config(&Config::with_none()));
-    }
-
     #[test]
     fn test_enabled_from_config_empty() {
```
kmoschcau, line 151: This test should now be renamed, to make clear it is testing the `enabled_from` function, not the `enabled_from_config` function. Alternatively, extract that bit of logic from the `enabled_from` function and put it back into the `enabled_from_config` function. The same goes for the other tests of this method and the tests for `depth_from[_config]`.

src/flags/recursion.rs
```diff
@@ -35,25 +34,18 @@ impl Recursion {
...
-    /// out warnings.
     fn enabled_from(matches: &ArgMatches, config: &Config) -> bool {
-        let mut result: bool = Default::default();
-
-        if config.has_yaml() {
-            if let Some(value) = Self::enabled_from_config(config) {
-                result = value;
+        if let Some(recursion) = &config.recursion {
+            if let Some(enabled) = recursion.enabled {
+                return enabled;
```
kmoschcau, line 40: This method breaks the previously agreed on precedence. Command line options should always override the config file options.

src/flags/recursion.rs
```diff
@@ -95,27 +68,24 @@ impl Recursion {
...
     /// returns it.
     fn depth_from(matches: &ArgMatches, config: &Config) -> Result<usize, Error> {
-        let mut result: Result<usize, Error> = Ok(usize::max_value());
-
-        if config.has_yaml() {
-            if let Some(value) = Self::depth_from_config(config) {
-                result = Ok(value);
+        if let Some(recursion) = &config.recursion {
+            if let Some(depth) = recursion.depth {
+                return Ok(depth);
```
kmoschcau, line 80: This method breaks the previously agreed on precedence. Command line options should always override the config file options.

src/flags/layout.rs
```diff
@@ -1,15 +1,16 @@
...
+use super::Configurable;
+
 use clap::ArgMatches;
-use yaml_rust::Yaml;
+use serde::Deserialize;
 
 /// The flag showing which output layout to print.
-#[derive(Clone, Debug, Copy, PartialEq, Eq)]
+#[derive(Clone, Debug, Copy, PartialEq, Eq, Deserialize)]
+#[serde(rename_all = "lowercase")]
```
kmoschcau, line 13: This should be changed to "kebab-case" to be future proof and consistent.

src/flags/size.rs
```diff
@@ -19,6 +21,23 @@ pub enum SizeFlag {
...
 }
 
+impl SizeFlag {
+    fn from_str(value: &str) -> Option<Self> {
+        match value {
+            "default" => Some(Self::Default),
+            "short" => Some(Self::Short),
+            "bytes" => Some(Self::Bytes),
+            _ => {
+                print_error!(
```
kmoschcau, line 31: No sure whether this should still panic or not.

src/flags/sorting.rs
```diff
@@ -156,13 +136,24 @@ impl Default for SortOrder {
...
     Last,
 }
 
+impl DirGrouping {
+    fn from_str(value: &str) -> Option<Self> {
+        match value {
+            "first" => Some(Self::First),
+            "last" => Some(Self::Last),
+            "none" => Some(Self::None),
+            _ => None,
```
kmoschcau, line 153: If the other `from_str` methods either print warnings or panic, this should as well. Especially since the previous implementation panicked.

src/flags/sorting.rs
```diff
@@ -330,67 +300,70 @@ mod test_sort_column {
     }
 
     #[test]
-    fn test_from_config_empty() {
-        let yaml_string = "---";
-        let yaml = YamlLoader::load_from_str(yaml_string).unwrap()[0].clone();
-        assert_eq!(None, SortColumn::from_config(&Config::with_yaml(yaml)));
-    }
+    fn test_from_config_none_column() {
```
kmoschcau, line 303: Change the test name back to `test_from_config_empty`. It is more consistent and essentially testing the same thing as the previous implementation.

src/flags/sorting.rs
```diff
@@ -426,31 +397,34 @@ mod test_sort_order {
...
-    }
-
     #[test]
     fn test_from_config_default() {
-        let yaml_string = "sorting:\n  reverse: false";
-        let yaml = YamlLoader::load_from_str(yaml_string).unwrap()[0].clone();
         assert_eq!(
             Some(SortOrder::Default),
-            SortOrder::from_config(&Config::with_yaml(yaml))
+            SortOrder::from_config(&Config::default())
```
kmoschcau, line 404: This test will break, if the default value in the `Config` changes, even though the implementation is still correct. The config variant should rather be specified directly in the test setup.

src/flags/sorting.rs
```diff
@@ -426,31 +397,34 @@ mod test_sort_order {
...
         );
     }
 
+    #[test]
+    fn test_from_config_none_reverse() {
```
kmoschcau, line 409: Change the test name back to `test_from_config_empty`. It is more consistent and essentially testing the same thing as the previous implementation.

src/flags/sorting.rs
```diff
@@ -459,10 +433,13 @@ mod test_dir_grouping {
     use super::DirGrouping;
 
     use crate::app;
-    use crate::config_file::Config;
+    use crate::config_file::{Config, Sorting};
     use crate::flags::Configurable;
 
-    use yaml_rust::YamlLoader;
+    #[test]
+    fn test_from_str_bad_value() {
```
kmoschcau, line 440: This needs to be handled differently, if an invalid value will cause a panic or print an error.

src/flags/sorting.rs
```diff
@@ -516,50 +493,48 @@ mod test_dir_grouping {
...
+        c.sorting = Some(Sorting {
+            column: None,
+            reverse: None,
+            dir_grouping: Some(DirGrouping::Last),
+        });
+        assert_eq!(Some(DirGrouping::Last), DirGrouping::from_config(&c));
     }
 
     #[test]
     fn test_from_config_explicit_none() {
```
kmoschcau, line 519: Change the test name back to `test_from_config_empty`. It is more consistent and essentially testing the same thing as the previous implementation.

kmoschcau (changes requested):
Sorry for the long wait, I finally had time to have a thorough look at this and I have some comments and requests for changes.

src/config_file.rs
```diff
@@ -1,175 +1,353 @@
...
-            file: None,
-            yaml: Some(yaml),
+    /// This constructs a Config struct with a passed file path [String].
+    pub fn with_file(file: String) -> Option<Self> {
+        match fs::read(&file) {
+            Ok(f) => Self::with_yaml(&String::from_utf8_lossy(&f)),
+            Err(e) => {
+                match e.kind() {
+                    std::io::ErrorKind::NotFound => {}
+                    _ => print_error!("Can not open config file {}: {}.", &file, e),
```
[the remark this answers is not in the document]
zwpaper: actually, the filename is printed by `&file`.

the logic behind this error handing is that:
1. most of the `lsd` user has not configuration file currently
2. skip the not found error to avoid printing unnecessary error msg
3. if an error other than `FileNotFound` happened, print the filename and error msg

the problem may be that is users used a wrong file, and `lsd` has not found it, there will be no error msg

src/flags/display.rs
```diff
@@ -45,37 +30,26 @@ impl Configurable<Self> for Display {
...
             None
         }
     }
 
     /// Get a potential `Display` variant from a [Config].
     ///
-    /// If the Config's [Yaml] contains a [String](Yaml::String) value pointed to by "display" and
-    /// it is either "all", "almost-all" or "directory-only", this returns the corresponding
-    /// `Display` variant in a [Some]. Otherwise this returns [None].
+    /// If the `Config::display` has value and is one of "all", "almost-all" or "directory-only",
```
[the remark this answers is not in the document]
zwpaper: `VisibleOnly` is much more like a `Default` value in my opinion, and can not be set by arg flags, this is why I did not add it to the comment.

but the missing value in doc-comment also seems like a problem, added as your comment.

As the enum comment part, IMHO, I think we could leave the doc comment as it is now:
1. the `serde.rs` simply deserializes the yaml into Config, although it checked the enum value, it is something like `lexical analysis`
2. `from_config` cares more like `syntactic analysis`, some other logic will  be added here(although it is not for `display`)

but I am open to accepting your idea if you insist.

src/flags/display.rs
```diff
@@ -128,40 +100,24 @@ mod test {
...
-        let yaml_string = "display: directory-only";
-        let yaml = YamlLoader::load_from_str(yaml_string).unwrap()[0].clone();
-        assert_eq!(
-            Some(Display::DirectoryItself),
-            Display::from_config(&Config::with_yaml(yaml))
-        );
+        let mut c = Config::with_none();
+        c.display = Some(Display::DirectoryOnly);
+        assert_eq!(Some(Display::DirectoryOnly), Display::from_config(&c));
     }
```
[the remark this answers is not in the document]
zwpaper: added the `VisibleOnly` test, it is not a bad thing adding those tests, maybe, just maybe, someday we use a brand new method to implement the configuration part, we still have to make sure flags mod work as expected, right? 😆

src/flags/recursion.rs
zwpaper: there is not more `enabled_from_config` function, so there may not be any more confusion?

src/flags/sorting.rs
zwpaper: Sorry, I did not get the point. could you please explain more about the change?

we use this test to make sure we are returning the right `default` value from config, and could be use in sorting, right?
IMO, It work as expected that failed if the default value in config changed.

src/config_file.rs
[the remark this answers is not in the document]
kmoschcau: I only wrote this comment here, because this is where the file information is lost in the flow of the configuration logic. My problem is that any other errors, caused by invalid configuration after this point, have no information any more in which file they happened.

src/flags/display.rs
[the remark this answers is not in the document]
kmoschcau: We will have to have a go at redesigning the logic in some places anyway.ike you stated, VisibleOnly can not be set from the CLI args, but this could actually be a problem if it is set to something different in the config, but the user wants to have it VisibleOnly for a single call. Right now there is no way to achieve that aside from modifying the config or ignoring it. But that is outside of the scope of this pull request.

src/flags/recursion.rs
kmoschcau: I always had the convention for test names of `test_<name of method under test>_<test case>`. Right now this breaks that convention.

src/flags/sorting.rs
kmoschcau: Your confusion here likely comes from the fact, that the SortOrder enum has a variant called "Default", however is has nothing to do with any real default, it is just a name. So if someone suddenly decides that the default value in the Config is now the "Reverse" variant instead of the "Default" variant, your test will immediately start failing, even though the method you are testing (SortOrder::from_config) is still working as expected. You should rather set up an empty Config in the test setup and and set the SortOrder to the "Default" variant explicitly.

zwpaper: I was thinking that this test case should make sure the default config should match the default sorting value,
so if someone changed the `Default`, then he should also change the config file part.
the case should fail, if config file part missed.

maybe the test case name is not that obvious, I changed it to `test_from_config_default_config`

src/flags/display.rs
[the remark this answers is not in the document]
zwpaper: https://github.com/Peltoche/lsd/issues/451
issue created

src/flags/recursion.rs
zwpaper: we were testing the `Recursion::enabled_from` function

`test_` `enabled_from` `config_empty`, looks like this is following the convention?

`config` before `empty` is because it's one of the functions args

src/config_file.rs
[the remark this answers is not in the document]
zwpaper: I have returned the error from `from_yaml` and print the error with the file name, when deserialized error.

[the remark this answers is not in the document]
kmoschcau: Yep, looks good to me.

src/flags/recursion.rs
kmoschcau: No quite correct. This test was originally testing the `Recursion::enabled_from_config` method, in the case that the passed Yaml config is empty. Now you are testing the `Recursion::enabled_from` method, with the case that no command line args were passed and that the config is empty as well. I guess we could just leave it like that, from that perspective, but to me it would not be immediately clear what this test does, just from the name. It would have been easier if there was still the `Recursion::enabled_from_config` method, to be able to test more granular.

src/flags/sorting.rs
kmoschcau: I mean sure, but in that case please also add a test case where the reverse is set to be explicitly `Some(false)`, just as there already is a test with `Some(true)`.

kmoschcau (changes requested):
I reviewed this again.
Also for future reference, please don't amend and force push, if your pull request is already under review. This just makes it unnecessarily hard for reviewers to find the changes you made since they last looked at the pull request.

zwpaper:
hi @kmoschcau , sorry for the rebase, I found that there is a WIP commit in my commit history and forgot that will break your review history.

I have updated the testing code according to the comment, please check it again, thanks

kmoschcau:
Yeps this looks good from my side now.

zwpaper:
@meain I have rebased onto master, could you check this PR one more time

meain:
Hey, this looks good. @zwpaper thanks for working on this. @kmoschcau , thanks a lot for the review.

One thing I think would be good to add in the future would be to improve the error message if we expect a struct and the user provide something else. For example, if we have config like this:

```yaml
color: never
```

the output is

```
lsd: Configuration file /Users/meain/.config/lsd/config.yaml format error, color: invalid type: string "never", expected struct Color at line 1 column 8.
```

 but struct `Color` means nothing to the user. We should be able to implement a [deserialize](https://serde.rs/impl-deserialize.html) for those fields, something generic for all struct fields I guess which could either output something a bit more generalized like "expected a map" or just ignore that part of the message.

PR #445, opened 2020-11-03
Add MacPorts as installation option on macOS

ylluminarious:
`lsd` is also available via MacPorts. Thought it was worth mentioning in the README.


meain:
Thank you. :)

Issue #365, opened 2020-05-04
Missing colors after installing nerd-fonts


tnorthcutt:
I recently moved to a new machine, and as a result, icons were missing in the output of `lsd` (instead I saw a [?] icon).

Today I followed the instructions [in this comment](https://github.com/Peltoche/lsd/issues/199#issuecomment-494218334) to install and configure nerd-fonts for iTerm2, which worked. However, now the color output of `lsd` is gone.

I tried uninstalling and reinstalling lsd (with `brew uninstall lsd` and `brew install lsd`), but that did not result in a change.

Current output:
<img width="478" alt="CleanShot 2020-05-04 at 07 03 27@2x" src="https://user-images.githubusercontent.com/796639/80968550-62998480-8dd5-11ea-9e08-aceac5d51499.png">
<br>
iTerm2 settings:
<img width="575" alt="CleanShot 2020-05-04 at 07 04 08@2x" src="https://user-images.githubusercontent.com/796639/80968599-7a710880-8dd5-11ea-86f7-27d263132a6e.png">

meain:
What are the values for `$TERM` and `$LS_COLORS`?

tnorthcutt:
@meain `$TERM` is `xterm-256color`; `echo $LS_COLORS` produces no output

meain:
Could you check if `lsd --color always` works?

tnorthcutt:
@meain `lsd --color always` results in the same output as a regular ls (which is aliased to lsd): icons and bolding, but no colors.

meain:
Hey, so far I don't have much of a clue as to what is happening. Could you give a few more details.

- `lsd --version`
- your shell (if zsh, are you using oh-my-zsh?)
- value of `which lsd`
- is there any color with `lsd -l` for items like date and user?

tnorthcutt:
@meain Absolutely! I appreciate your attempts to help 😄 

- `lsd --version`: `0.17.0`
- zsh, and yes using oh-my-zsh
- `which lsd`: `/usr/local/bin/lsd`
- `lsd -l`: yes, color for some things! Output below:

<img width="828" alt="CleanShot 2020-05-07 at 08 15 35@2x" src="https://user-images.githubusercontent.com/796639/81304960-ff109080-903a-11ea-9bc5-99b99dcbf88d.png">

meain:
I have a feeling that `LS_COLORS` might be set as blank string or something. 
[...]

PR #452, opened 2020-12-06
yaml based theme support

zwpaper:
fix https://github.com/Peltoche/lsd/issues/92 
related: https://github.com/Peltoche/lsd/issues/365

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,7 +225,7 @@ color:
 # This specifies the date format for the date column. The freeform format
 # accepts an strftime like string.
 # When "classic" is set, this is set to "date".
-# Possible values: date, relative, '+<date_format>' 
+# Possible values: date, relative, '+<date_format>'
 # `date_format` will be a `strftime` formatted value. e.g. `date: '+%d %b %y %X'` will give you a date like this: 17 Jun 21 20:14:55
 date: date
 
@@ -318,8 +318,6 @@ symlink-arrow: ⇒
 Theme can be configured in the [configuration file](#configuration)(color.theme),
 The valid theme configurations are:
 - `default`: the default color scheme shipped in `lsd`
-- `no-color`: classic mode, there will be just black and white.
-- `no-lscolors`: do not check the `LSCOLORS` env
 - theme-file-name(yaml): use the theme file to specify colors
 
 when configured with the `theme-file-name` which is a `yaml` file,
@@ -336,6 +334,8 @@ configure the colors, check [crossterm](https://docs.rs/crossterm/0.20.0/crosste
 for the supported colors.
 
 Color table: https://upload.wikimedia.org/wikipedia/commons/1/15/Xterm_256color_chart.svg
+
+Please notice that color value would ignore case, both lowercase and UPPERCASE is supported.
 
 This is the default theme scheme shipped with `lsd`.
 
--- a/README.md
+++ b/README.md
@@ -215,9 +215,9 @@ color:
   when: auto
   # How to colorize the output.
   # When "classic" is set, this is set to "no-color".
-  # Possible values: default, no-color, no-lscolors, <theme-file-name>
-  # when specifying <theme-file-name>, lsd will look up theme file in
-  # XDG Base Directory if relative
+  # Possible values: default, <theme-file-name>
+  # when specifying <theme-file-name>, lsd will look up theme file
+  # XDG Base Directory if relative, e.g. ~/.config/lsd/themes/<theme-file-name>,
   # The file path if absolute
   theme: default
 
@@ -322,7 +322,7 @@ The valid theme configurations are:
 
 when configured with the `theme-file-name` which is a `yaml` file,
 `lsd` will look up the theme file in the following way:
-- relative name: check the XDG Base Directory
+- relative name: check the themes under XDG Base Directory, e.g. ~/.config/lsd/themes/<theme-file-name>
 - absolute name: use the file path and name to find theme file
 
 Check [Theme file content](#theme-file-content) for details.
--- a/README.md
+++ b/README.md
@@ -331,8 +331,8 @@ Check [Theme file content](#theme-file-content) for details.
 
 ### Theme file content
 
-Theme file use the [ansi_term](https://docs.rs/ansi_term)
-configure the colors, check [ansi_term](https://docs.rs/ansi_term/0.12.1/ansi_term/enum.Colour.html)
+Theme file use the [crossterm](https://crates.io/crates/crossterm)
+configure the colors, check [crossterm](https://docs.rs/crossterm/0.20.0/crossterm/style/enum.Color.html)
 for the supported colors.
 
 Color table: https://upload.wikimedia.org/wikipedia/commons/1/15/Xterm_256color_chart.svg
@@ -340,68 +340,47 @@ Color table: https://upload.wikimedia.org/wikipedia/commons/1/15/Xterm_256color_
 This is the default theme scheme shipped with `lsd`.
 
 ```yaml
-user:
-  Fixed: 230
-group:
-  Fixed: 187
+user: 230
+group: 187
 permissions:
   read: Green
   write: Yellow
   exec: Red
-  exec-sticky: Purple
-  no-access:
-    Fixed: 245
+  exec-sticky: 5
+  no-access: 245
 file-type:
   file:
-    exec-uid:
-      Fixed: 40
-    uid-no-exec:
-      Fixed: 184
-    exec-no-uid:
-      Fixed: 40
-    no-exec-no-uid:
-      Fixed: 184
+    exec-uid: 40
+    uid-no-exec: 184
+    exec-no-uid: 40
+    no-exec-no-uid: 184
   dir:
-    uid:
-      Fixed: 33
-    no-uid:
-      Fixed: 33
-  pipe:
-    Fixed: 44
+    uid: 33
+    no-uid: 33
+  pipe: 44
   symlink:
-    default:
-      Fixed: 44
-    broken:
-      Fixed: 124
-  block-device:
-    Fixed: 44
-  char-device:
-    Fixed: 172
-  socket:
-    Fixed: 44
-  special:
-    Fixed: 44
+    default: 44
+    broken: 124
+  block-device: 44
+  char-device: 172
+  socket: 44
+  special: 44
 modified:
-  hour-old:
-    Fixed: 40
-  day-old:
-    Fixed: 42
-  older:
-    Fixed: 36
+  hour-old: 40
+  day-old: 42
+  older: 36
 size:
-  none:
-    Fixed: 245
-  small:
-    Fixed: 229
-  medium:
-    Fixed: 216
-  large:
-    Fixed: 172
+  none: 245
+  small: 229
+  medium: 216
+  large: 172
 inode:
-  valid:
-    Fixed: 13
-  invalid:
-    Fixed: 245
+  valid: 13
+  invalid: 245
+links:
+  valid: 13
+  invalid: 245
+tree-edge: 245
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -327,12 +327,6 @@ when configured with the `theme-file-name` which is a `yaml` file,
 
 Check [Theme file content](#theme-file-content) for details.
 
-### Filename coloring
-
-`lsd` use both [LS_COLORS](https://github.com/trapd00r/LS_COLORS) and this theme feature to
-colorize the filename, `lsd` will use the color and format option defined in `LS_COLORS`,
-and then the theme configurations if no `LS_COLOR` are found.
-
 ### Theme file content
 
 Theme file use the [crossterm](https://crates.io/crates/crossterm)
@@ -348,30 +342,13 @@ This is the default theme scheme shipped with `lsd`.
 ```yaml
 user: 230
 group: 187
-permissions:
+permission:
   read: Green
   write: Yellow
   exec: Red
   exec-sticky: 5
   no-access: 245
-file-type:
-  file:
-    exec-uid: 40
-    uid-no-exec: 184
-    exec-no-uid: 40
-    no-exec-no-uid: 184
-  dir:
-    uid: 33
-    no-uid: 33
-  pipe: 44
-  symlink:
-    default: 44
-    broken: 124
-  block-device: 44
-  char-device: 172
-  socket: 44
-  special: 44
-modified:
+date:
   hour-old: 40
   day-old: 42
   older: 36
@@ -427,7 +404,7 @@ Some further examples of useful aliases:
 In the future the possibility to customize the colors might be implemented.
 For now, the default colors are:
 
-| User/Group | Permissions | File Types | Last time Modified | File Size |
+| User/Group | Permission | File Types | Date | File Size |
 |:---|:---|:---|:---|:---|
 |![#ffffd7](https://via.placeholder.com/15/ffffd7/000000?text=+) User|![#00d700](https://via.placeholder.com/15/00d700/000000?text=+) Read |![#0087ff](https://via.placeholder.com/15/0087ff/000000?text=+) Directory|![#00d700](https://via.placeholder.com/15/00d700/000000?text=+) within the last hour|![#ffffaf](https://via.placeholder.com/15/ffffaf/000000?text=+) Small File|
 |![#d7d7af](https://via.placeholder.com/15/d7d7af/000000?text=+) Group|![#d7ff87](https://via.placeholder.com/15/d7ff87/000000?text=+) Write|![#00d700](https://via.placeholder.com/15/00d700/000000?text=+) Executable File|![#00d787](https://via.placeholder.com/15/00d787/000000?text=+) within the last day|![#ffaf87](https://via.placeholder.com/15/ffaf87/000000?text=+) Medium File|
--- a/README.md
+++ b/README.md
@@ -213,6 +213,13 @@ color:
   # When "classic" is set, this is set to "never".
   # Possible values: never, auto, always
   when: auto
+  # How to colorize the output.
+  # When "classic" is set, this is set to "no-color".
+  # Possible values: default, no-color, no-lscolors, <theme-file-name>
+  # when specifying <theme-file-name>, lsd will look up theme file in
+  # XDG Base Directory if relative
+  # The file path if absolute
+  theme: default
 
 # == Date ==
 # This specifies the date format for the date column. The freeform format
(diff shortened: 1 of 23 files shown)


zwpaper:
@meain there is a theme config `no-lscolors` maybe we need to discuss.

previously, we use `LSCOLORS` by default, so that there is a configuration `no-lscolors`, but after this `theme` implementation, `no-lscolors` seems has not effect, should we change it to `lscolors` and read `LSCOLORS` instead of `theme file`.
but doing this, may break some compatibility if people used `--no-lscolors` before.

meain (commented):
Hey @zwpaper , we had never exposed `--no-lscolors` to the user. It was an internal only thing.

Even if we implement themes via config file, we should still have `LS_COLORS` used by default as that will work better with user's existing terminal colors. All they would need to override this would be to specify a theme in the config file. I believe most people will expect `LS_COLORS` to work as that is a more general standard across multiple tools.

Also, I think we should let the user configure the values of non-name blocks via the config file but still let them use `LS_COLORS` for the name block. There are projects like [trapd00r/LS_COLORS](https://github.com/trapd00r/LS_COLORS) which has great support for coloring files based on their extension etc. I am not exactly sure as how to approach this yet though.

meain:
Two more things:

- Color names would be better as lowercase values
- Need to support something like `default` for a color value which would not apply any color (uses terminal foreground color)

README.md
```diff
@@ -271,6 +278,100 @@ total-size: false
...
+configure the colors, check [ansi_term](https://docs.rs/ansi_term/0.12.1/ansi_term/enum.Colour.html)
+for the supported colors.
+
+Color table: https://upload.wikimedia.org/wikipedia/commons/1/15/Xterm_256color_chart.svg
+
+This is the default theme scheme shipped with `lsd`.
+
+```yaml
+user:
+  Fixed: 230
```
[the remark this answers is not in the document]
zwpaper: actually, this is also what I previously thought, but the `ansi_term` deserialize it in this way, I will try to impl custom deserialization.

zwpaper:
waiting on this PR https://github.com/ogham/rust-ansi-term/pull/71

zwpaper:
As we were blocked by the `ansi-term` PR, I will later try a new crate https://github.com/crossterm-rs/crossterm , it seems more active.

I will do some tests and benchmarks for the change later

zwpaper:
new PR created https://github.com/crossterm-rs/crossterm/pull/576

zwpaper:
status update, I have reimplemented the theme support based on `crossterm`, it could be built and basically work as expected with my latest commit.

I will then work on the tests and make sure not to break the original functions.

TheTrio:
Any updates about this PR? Can't wait to have custom theme support. 

Thanks!

README.md
```diff
@@ -271,6 +278,100 @@ total-size: false
...
 ```
 
+## Theme
+
+`lsd` can be configured with a theme file to set the colors.
+
+Theme can be configured in the [configuration file](#configuration)(color.theme),
+The valid theme configurations are:
+- `default`: the default color scheme shipped in `lsd`
+- `no-color`: classic mode, there will be just black and white.
```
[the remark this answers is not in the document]
zwpaper: I deleted this flag from the readme, but leave it in source code as it was used to define the theme color-related stuff, and it was used before this yaml support, we may update it in another PR if needed

README.md
```diff
@@ -271,6 +278,100 @@ total-size: false
...
+- `no-lscolors`: do not check the `LSCOLORS` env
```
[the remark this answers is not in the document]
zwpaper: ditto, deleted in readme

zwpaper:
pin bitflags dependency to `1.2.1` due to https://github.com/clap-rs/clap/issues/2691

zwpaper:
hi @meain, I have done this implementation, could we leave the code coverage for later improvement.

and for the comment:

1. color value would ignore the cases so that both lowercase and UPPERCASE would work, I have added the description to readme.
2. new issue https://github.com/Peltoche/lsd/issues/549 is created for the default part, we may discuss it there and let the theme support release first

zwpaper:
finally, this one-third year PR is coming true 🤩

src/flags/layout.rs
```diff
@@ -10,7 +10,7 @@ use serde::Deserialize;
 
 /// The flag showing which output layout to print.
 #[derive(Clone, Debug, Copy, PartialEq, Eq, Deserialize)]
-#[serde(rename_all = "kebab-case")]
+#[serde(rename_all = "lowercase")]
```
meain, line 13: I guess we can drop the special casing for `OneLine` now.

src/meta/mod.rs
```diff
@@ -248,3 +248,16 @@ impl Meta {
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::Meta;
```
meain, line 254: We can probably move the `#[cfg(unix)]` thing above so that Github actions will not add this warning of unused import.

src/color.rs
```diff
@@ -60,90 +66,134 @@ impl Elem {
...
+            ThemeOption::Custom(ref file) => {
+                Some(Theme::from_path(file).unwrap_or_else(Theme::default))
+            }
         };
-        let lscolors = match theme {
-            Theme::NoColor => None,
-            Theme::Default => Some(LsColors::from_env().unwrap_or_default()),
-            Theme::NoLscolors => None,
+        let lscolors = match t {
+            ThemeOption::Default => Some(LsColors::from_env().unwrap_or_default()),
```
meain, line 144: `ThemeOption::Custom` should also use `LsColors`. Everything except `NoColor` or `NoLscolors` should make use of the lscolors values.

README.md
```diff
@@ -304,6 +311,79 @@ total-size: false
...
+
+Theme can be configured in the [configuration file](#configuration)(color.theme),
+The valid theme configurations are:
+- `default`: the default color scheme shipped in `lsd`
+- theme-file-name(yaml): use the theme file to specify colors
+
+when configured with the `theme-file-name` which is a `yaml` file,
+`lsd` will look up the theme file in the following way:
+- relative name: check the XDG Base Directory
+- absolute name: use the file path and name to find theme file
```
meain, line 326: Not sure if absolute will be of all that much use, but I guess it couldn't hurt.

README.md
```diff
@@ -304,6 +311,79 @@ total-size: false
...
+`lsd` can be configured with a theme file to set the colors.
+
+Theme can be configured in the [configuration file](#configuration)(color.theme),
+The valid theme configurations are:
+- `default`: the default color scheme shipped in `lsd`
+- theme-file-name(yaml): use the theme file to specify colors
+
+when configured with the `theme-file-name` which is a `yaml` file,
+`lsd` will look up the theme file in the following way:
+- relative name: check the XDG Base Directory
```
meain, line 325: Should we probably give an example filepath? Also, what is your opinion on having the themes under a `themes` subdirectory?

Also, when the user mentioned `color.theme: my-theme`, it should lookup `CONFIG_DIR/themes/my-theme.yaml`.

meain (commented):
Hey, just a have few changes.

Also, for some reason the colors for the filenames(lscolors) look different compared to how it was before when I have no lscolors set. Let me see if I can figure out what is going on.

Screenshot 2021-08-31 at 9 53 20 AM

README.md
```diff
@@ -213,12 +213,19 @@ color:
   # When "classic" is set, this is set to "never".
   # Possible values: never, auto, always
   when: auto
+  # How to colorize the output.
+  # When "classic" is set, this is set to "no-color".
+  # Possible values: default, no-color, no-lscolors, <theme-file-name>
```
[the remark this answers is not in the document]
zwpaper: oops, I have deleted those 2 options in the readme, but forgot to delete it here, will delete both `no-lscolor` and `no-lscolor`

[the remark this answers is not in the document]
zwpaper: for the color part, there is no difference with the older version in my machine, I will check the diff for color later

src/flags/icons.rs
```diff
@@ -100,6 +100,17 @@ pub enum IconTheme {
     Fancy,
 }
 
+impl IconTheme {
+    // /// Get a value from a string.
```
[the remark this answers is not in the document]
zwpaper: yes, but how about leaving the empty impl here? I have a plan for the configuration file for icons.

README.md
zwpaper: the absolute path will be useful in tests scenario, and some others like trying newly downloaded themes, just like you said, it couldn't hurt

README.md
zwpaper: actually, I was also hesitant about the `themes` subdir, maybe adding it would be a better choice considering the icon-theme plan, maybe we could also add one like `icons` for the icon configuration.

meain: Yeah, otherwise it will be a mess of yaml files in the root config dir if the user has a lot of themes.

README.md
```diff
@@ -213,12 +213,19 @@ color:
   # When "classic" is set, this is set to "never".
   # Possible values: never, auto, always
   when: auto
+  # How to colorize the output.
+  # When "classic" is set, this is set to "no-color".
+  # Possible values: default, <theme-file-name>
+  # when specifying <theme-file-name>, lsd will look up theme file
+  # XDG Base Directory if relative, e.g. ~/.config/lsd/themes/<theme-file-name>,
```
meain, line 220: I think it would be better to automatically add the `.yaml` extension
So, it will be `~/.config/lsd/themes/<theme-file-name>.yaml`

README.md
```diff
@@ -304,6 +311,79 @@ total-size: false
...
+```yaml
+user: 230
+group: 187
+permissions:
+  read: Green
+  write: Yellow
+  exec: Red
+  exec-sticky: 5
+  no-access: 245
+file-type:
```
meain, line 351: We don't really use these colors as we are always leveraging LSCOLORS for this. We can drop this. Anything that colors files had been dropped. Right now all file base coloring is handled by LSCOLORS.

We could have kept this, but then we will miss out on being able to use LSCOLORS which has way more configuration options.

src/flags/icons.rs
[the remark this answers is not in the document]
meain: Not sure what it would do more than a single line comment with/without a TODO entry.

README.md
zwpaper: I do really get this, do you talking about issue #549?

README.md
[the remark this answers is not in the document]
meain: Screenshot 2021-08-31 at 5 15 12 PM

The whole color thing is a bit weird to me. There is a subtle change in the colors. Not exactly sure what is going on.

README.md
meain: We only make use of these colors when `ThemeOption::NoLscolors` is set.
This is not available to the user and is only use it tests. So there is no point in exposing these options to the user.

src/flags/color.rs
```diff
@@ -23,7 +26,73 @@ impl Color {
...
+            }
+
+            fn visit_str<E>(self, value: &str) -> Result<ThemeOption, E>
+            where
+                E: de::Error,
+            {
+                match value {
+                    "no-color" => Ok(ThemeOption::NoColor),
+                    "default" => Ok(ThemeOption::Default),
+                    "no-lscolors" => Ok(ThemeOption::NoLscolors),
```
meain, line 89: This can be simplified. If the user has not mentioned any, just don't do anything and the `Default` will be used. If the user has mentioned something we do a file lookup.

This would mean that if the user mentions `default` we will be looking up the `default.yaml` file and erroring out if it does not exist. This might prevent us from packaging themes with `lsd` binary but other than that it might be a better option. What do you think?

README.md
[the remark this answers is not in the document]
zwpaper: Hmmm... it is really weird, it works as expected in my mac, even I use your tested LS_COLORS

image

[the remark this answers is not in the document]
meain: Hey, ignore the color issue. It seems like it was something messy going on with my shell.

README.md
zwpaper: do you mean the entire `file-type` section? it makes sense that we should delete it if it could not make changes, but should we add an option for configuration it later?

and you raised an important point, we can only configure the color but not the format, should we consider adding the support for font formats like `bold`, `italy`, or something else?

src/flags/color.rs
zwpaper: I would like to keep the current implementation, because,

1. the `default.yaml` disk io is not necessary
2. it may be necessary for us to package default dark and default white theme with `lsd`, even we support theme config now.

meain: It might end up being useful for dark and light if we end up shipping it in future. I guess we can keep the code around but just drop `no-color` and `no-lscolors` for now.
We can drop `no-color` as it is exactly the same as `color: never` and is technically not a theme. As for `no-lscolors`, that is also not really a theme, but an option to disable LSCOLORS in the default theme.

README.md
meain: If we are to let the users configure `file-type` section, then we probably will have to have a flag to disable/enable LSCOLORS, as in users ca provide a theme name as well as set lscolors to false, but that is probably an overkill for now. LSCOLORS is capable enough to handle most things.

As for `bold`, `italics`; if you are talking about for the file type ones LSCOLORS can actually handle it. `bold` and `italics` for the other sections can come in a future pr if it is really needed(which I don't really think it is).

zwpaper: yes, we have the same opinion for the `file-type` section, and I prefer we would implement it later in another PR.

for the `bold`, `italics` part, It is also not that necessary for me, but what I was considering is that there would be breaking changes if we have to add it.

as now we support themes, I prefer we could use either `LSCOLORS` or `themes` for all the coloring, not mixing them.

after a quick thought, we should be able to handle the breaking changes for the `bold` part, I will delete the `file-type` first and then create 2 issues for those 2 discussions.

meain: The reason why I did not want to switch to using `theme` for file name coloring is becasuse LSCOLORS is way more powerful as of now. It can color based on the name and supports things like bold, italics, underline. Also, there is a lot of projects like https://github.com/trapd00r/LS_COLORS which we will miss out on if we drop that functionality. If I remember correctly, even exa seems to be doing the same.

As for the bold/italics part of the non file-type icons, I don't think we should really bother about it now. If at all that comes up, we can still support it without breaking anything in config by just checking if the value of something is just a string/number or an object which contains bold and italics keys.

zwpaper: I recheck the implementation, confirm that the theme does affect the lscolors, we will always check lscolors for file-name and then theme if no lscolors are found, I will document it to readme.

https://github.com/zwpaper/lsd/blob/03f406dd7717dba2d1074d456e4a925d24c8a30b/src/color.rs#L154-L160

so that It should be necessary to leave the file-type part in the theme and then we should discuss the formatting like bold, italics, underline here https://github.com/Peltoche/lsd/issues/550, and cowork with LSCOLORS here https://github.com/Peltoche/lsd/issues/551

src/flags/color.rs
zwpaper: the `ThemeOption::default` and `ThemeOption::NoLscolor` are originally used by `lsd`, which controls most of coloring logic.

there is quite a lot of work to change this logic here, I have deleted these two options in the readme, maybe we should just leave this implementation here and improve it later

README.md
meain: As of now, we let LSCOLORS pick the default colors if we are not able to find LSCOLORS variable. This was done because we got a lot of issues saying that the colors that lsd had did not match their terminal colors. I don't think we should change that as of now.

meain: We can probably discuss more on how to override the use of LSCOLORS in #551 but I don't think we should be changing the behavior of picking default LSCOLORS if no env value is set.

src/flags/color.rs
meain: What I meant by remove those is to just not parse the string. No need to change anything in the enum. For example if the user writes `no-color`, instead of assigning to `ThemeOption::default` we should look up `themes/no-color.yaml`.

README.md
zwpaper: like I said, this theme feature does not change the origin logic picking up colors, it replaces the last default option hardcoded before.

https://github.com/Peltoche/lsd/pull/452/files?file-filters%5B%5D=.lock&file-filters%5B%5D=.md&file-filters%5B%5D=.rs#diff-12d1cf56d5d03c679a5f8f1c48cfe969a5c4276a1bee8647cc548de13f18e465L100-L111

the `LS_COLORS` env you mentioned is handled here before the theme coloring https://github.com/Peltoche/lsd/pull/452/files#diff-12d1cf56d5d03c679a5f8f1c48cfe969a5c4276a1bee8647cc548de13f18e465L87-L91

zwpaper: we are not breaking anything here, the workflow is still:
1. create `lscolors` from env, and default if not found 
   ```rust
   ThemeOption::Default => Some(LsColors::from_env().unwrap_or_default())
   ```
2. check `lscolors` is here a color option for the filename(LS_COLORS env already considered in `)
3. if no lscolors are found, use the theme configurations

zwpaper:
new issue created for the `NoLscolor` dead code https://github.com/Peltoche/lsd/issues/552

README.md
zwpaper: @meain what do you think? this PR should be ready if we resolve this question.

meain: I don't think we should ship with `file-type` when that is something that the user will never be able to use. The config should not have it. Few other nitpicks as well, just to keep the flags and config consistent, `permissions` -> `permission`, `modified` -> `date`.

We should also probably also accept both `yaml` and `yml` as extensions for the theme file but that is not that big a deal.

Sorry about the nitpicks, but once this goes out, it will be really hard to change anything without breaking the config for users.

README.md
```diff
@@ -304,6 +311,85 @@ total-size: false
...
+- theme-file-name(yaml): use the theme file to specify colors(without the `yaml` extension)
+
+when configured with the `theme-file-name` which is a `yaml` file,
+`lsd` will look up the theme file in the following way:
+- relative name: check the themes under XDG Base Directory, e.g. ~/.config/lsd/themes/<theme-file-name>.yaml
+- absolute name: use the file path and name to find theme file
+
+Check [Theme file content](#theme-file-content) for details.
+
+### Filename coloring
```
meain, line 330: We should drop this once we make the change as well.

meain:
Hey, I was looking more into the original color issue that I was talking about. The color values being outputted is different. From what I can tell, old one was using default colors and the new one is using bright colors. If the shell is set up with different normal and bright colors, this will cause issues.

Screenshot 2021-09-09 at 5 47 38 PM


new:
```
[38;5;12m[1md[0m[38;5;10mr[39m[38;5;11mw[39m[38;5;9mx[39m[38;5;10mr[39m[38;5;245m-[39m[38;5;9mx[39m[38;5;10mr[39m[38;5;245m-[39m[38;5;9mx[39m [38;5;229m160[39m[38;5;229mB[39m [38;5;36m3 months ago[39m [38;5;12m[1mci[0m
```

old:
```
[1;34md[0m[32mr[33mw[31mx[32mr[38;5;245m-[31mx[32mr[38;5;245m-[31mx[0m [38;5;229m160B[0m [38;5;36m3 months ago[0m [1;34mci[0m
```

README.md
meain: Something along the lines of https://github.com/meain/lsd/commit/0788c6f21c9f75685f1cfbfd2e936f841f3fca47

meain:
Also, lscolors supports `crossterm`. We should be using that instead of converting it internally btw. https://github.com/sharkdp/lscolors/blob/bc95124f23d2f1c14616e6a97f5ae9e6ab786697/Cargo.toml#L22

README.md
zwpaper: will do this for consistency, but `modified` seems more accurate than `date`.

meain: The reason why I suggested `date` instead of `modified` is so that they will be able to map it directly from the `blocks` values. All the other values are same as the one in `blocks`.

zwpaper:
@meain I have changed to dark color, for the consistency with the older version.

for the lscolors supported crossterm part, there is not yet a released version supporting the crossterm, I created an issue for a newer version https://github.com/sharkdp/lscolors/issues/34.

I also noticed that the lscolors map its red to the red both `ansi_term` and `crossterm`, maybe we should pick up an ansi value instead of using the color name later.

but the choice of default colors would be a new problem.

zwpaper:
btw, I think we could make theme happen and do not have to wait for the https://github.com/sharkdp/lscolors/issues/34, we can change to the lscolors implementation any time later.

meain:
Looks great, I think we are good unless you have something else to add/tweak.

zwpaper:
hi @meain, I have done the jobs in my mind currently, let's make this real now! we can add the others in other PRs.

meain:
Awesome, thanks a lot @zwpaper for working on this.

PR #456, opened 2020-12-11
supplementing PR for tree -d

coolaj86:
Re: https://github.com/Peltoche/lsd/issues/295 https://github.com/Peltoche/lsd/pull/426

- only follows regular directories (no `-l` implementation)
- adds test for flags/display.rs

I'm not clear on how the test should work with for the YAML files. I tried this, expecting that it would parse the same as the flags, but it did not work as I had hoped:

```rs
    #[test]
    fn test_from_config_tree_d() {
        let yaml_string = "display: directory-only\nlayout: tree";
        let yaml = YamlLoader::load_from_str(yaml_string).unwrap()[0].clone();
        assert_eq!(
            Some(Display::TreeD),
            Display::from_config(&Config::with_yaml(yaml))
        );
    }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for `--extensionsort` `-X` from [aldhsu](https://github.com/aldhsu)
 - Add support for `--versionsort` `-v` from [zwpaper](https://github.com/zwpaper)
 - Add support for config symlink arrow from [zwpaper](https://github.com/zwpaper) [#409](https://github.com/Peltoche/lsd/issues/409)
-- Implement `--tree -d`, analogous to `tree -d` from [0jdxt](https://github.com/0jdxt)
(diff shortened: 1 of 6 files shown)


meain:
@coolaj86 I had skipped the config part earlier, thanks for bringing it up.

Since `--tree -d` is a combination that we are adding, we have to specifically look for that in config and modify the code similar to [this](https://github.com/Peltoche/lsd/pull/456/files#diff-02ac82c5458d114dc5673edf836f46cfd72832f04df332cc25698dc364ce5c9dL48). But I think a better way to do that would be to just check if `DirectoryItself` flag is set instead of combining `--tree` and `-d` into a new display type `TreeD`.

coolaj86:
@meain I changed the code to combine the Display and Layout rather than adding a new Display.

meain (approved):
Looks good. Thanks :)

coolaj86:
Thanks.

What needs to happen to get a new release out?

As soon as you publish to Github Releases, it will be available on [webinstall.dev/lsd](https://webinstall.dev/lsd) for all supported platforms, including Apple Silicon, if you include that in set of build.

meain:
Planning to get a new release out pretty soon. Just waiting for https://github.com/Peltoche/lsd/pull/443 to land which should happen pretty soon.

Issue #407, opened 2020-08-30
show hard link number with -l (second column of ls -l) 


M4tT3d:
excuse me. I know this is not the right section, but I would like to know how to show hard links too. In help I can't find anything and there is no manual. Thanks and sorry again

meain:
I am not sure if there is a standard way to do this, but you can check the inode number of files by using `lsd -i` and see if they are the same.

sklages:
Well that works for files in the same directory:
```
$ lsd -li s*
1803502 .rwxr-xr-x user_b user_b 2.7 KB Wed Sep 30 21:22:58 2020  run.hg38.sh
1787951 .rwxr-xr-x user_b user_b 1.5 KB Wed Sep 30 21:25:42 2020  run.mm10.sh
1787951 .rwxr-xr-x user_b user_b 1.5 KB Wed Sep 30 21:25:42 2020  run.mm10.sh.link
```
When files are not in the same directory, it gets complicated :-)

What @M4tT3d is referring to is to get column number two of `ls -l`:
```
-rwxr-xr-x 1 user_b user_b 2804 30. Sep 21:22 run.hg38.sh
-rwxr-xr-x 2 user_b user_b 1514 30. Sep 21:25 run.mm10.sh
-rwxr-xr-x 2 user_b user_b 1514 30. Sep 21:25 run.mm10.sh.link
```

M4tT3d:
> Well that works for files in the same directory:
> 
> ```
> $ lsd -li s*
> 1803502 .rwxr-xr-x user_b user_b 2.7 KB Wed Sep 30 21:22:58 2020  run.hg38.sh
> 1787951 .rwxr-xr-x user_b user_b 1.5 KB Wed Sep 30 21:25:42 2020  run.mm10.sh
> 1787951 .rwxr-xr-x user_b user_b 1.5 KB Wed Sep 30 21:25:42 2020  run.mm10.sh.link
> ```
> 
> When files are not in the same directory, it gets complicated :-)
> 
> What @M4tT3d is referring to is to get column number two of `ls -l`:
> 
> ```
> -rwxr-xr-x 1 user_b user_b 2804 30. Sep 21:22 run.hg38.sh
> -rwxr-xr-x 2 user_b user_b 1514 30. Sep 21:25 run.mm10.sh
> -rwxr-xr-x 2 user_b user_b 1514 30. Sep 21:25 run.mm10.sh.link
> ```

Yes. It is exactly what I mean

[...]

PR #459, opened 2020-12-13
Fix for #407: Show number of hardlinks for lsd -l

TheAlakazam:
+ Added implementation for displaying number of hardlinks to a file.
+ Displays '-' in windows
---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -49,7 +49,7 @@ pub enum Elem {
         valid: bool,
     },
 
-    INodeCount {
+    Links {
         valid: bool,
     },
 }
@@ -169,7 +169,7 @@ impl Colors {
                 true => Some("so"),
                 false => Some("no"),
             },
-            Elem::INodeCount { valid } => match valid {
+            Elem::Links { valid } => match valid {
                 true => Some("so"),
                 false => Some("no"),
             },
@@ -251,8 +251,8 @@ impl Colors {
         // INode
         m.insert(Elem::INode { valid: true }, Colour::Fixed(13)); // Pink
         m.insert(Elem::INode { valid: false }, Colour::Fixed(245)); // Grey
-        m.insert(Elem::INodeCount { valid: true }, Colour::Fixed(134));
-        m.insert(Elem::INodeCount { valid: false }, Colour::Fixed(222));
+        m.insert(Elem::Links { valid: true }, Colour::Fixed(134));
+        m.insert(Elem::Links { valid: false }, Colour::Fixed(222));
 
         m
     }
--- a/src/color.rs
+++ b/src/color.rs
@@ -251,8 +251,8 @@ impl Colors {
         // INode
         m.insert(Elem::INode { valid: true }, Colour::Fixed(13)); // Pink
         m.insert(Elem::INode { valid: false }, Colour::Fixed(245)); // Grey
-        m.insert(Elem::Links { valid: true }, Colour::Fixed(134));
-        m.insert(Elem::Links { valid: false }, Colour::Fixed(222));
+        m.insert(Elem::Links { valid: true }, Colour::Fixed(13));
+        m.insert(Elem::Links { valid: false }, Colour::Fixed(245));
 
         m
     }
--- a/src/color.rs
+++ b/src/color.rs
@@ -48,6 +48,10 @@ pub enum Elem {
     INode {
         valid: bool,
     },
+
+    INodeCount {
+        valid: bool,
+    },
 }
 
 impl Elem {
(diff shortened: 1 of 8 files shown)


src/color.rs
```diff
@@ -243,6 +251,8 @@ impl Colors {
         // INode
         m.insert(Elem::INode { valid: true }, Colour::Fixed(13)); // Pink
         m.insert(Elem::INode { valid: false }, Colour::Fixed(245)); // Grey
+        m.insert(Elem::INodeCount { valid: true }, Colour::Fixed(134));
```
meain, line 254: It is better to use colors <16. I think we can reuse `13` and `245` here.

src/flags/blocks.rs
```diff
@@ -136,7 +136,8 @@ impl Blocks {
 
     /// Prepends a [Block] of variant [INode](Block::INode) to `self`.
     fn prepend_inode(&mut self) {
-        self.0.insert(0, Block::INode)
+        self.0.insert(0, Block::INode);
+        self.0.insert(1, Block::INodeCount)
```
meain, line 140: As of now, I think we can have `inode-count` enabled only when they specify it explicitly in blocks. Let `--inode` only include the `inode` block.

src/flags/blocks.rs
```diff
@@ -181,6 +183,7 @@ impl TryFrom<&str> for Block {
             "date" => Ok(Self::Date),
             "name" => Ok(Self::Name),
             "inode" => Ok(Self::INode),
+            "inode_count" => Ok(Self::INodeCount),
```
meain, line 186: `inode-count`.

meain (commented):
Looks good, a few minor changes.

meain:
I was thinking we could change the name to `links`. Single word name would be better in this case.

src/flags/blocks.rs
TheAlakazam: But the number of hard links is only show when the inode option is selected they, are tied in together. I could seperate and create a `prepend_inode_count()` method to do so and call it after the `prepend_inode()` method, would that be alright?

meain: What I meant was we can leave the `--inode` flag as is and have `links`(`inode-count`) show up only when it is passed explicitly in `--blocks`. Something like `lsd --blocks name,size,links`.

TheAlakazam: I had a doubt, `links` should only be shown with the long option or also as is when passed to blocks.

meain: Actually both, but that is part of a separate logic and you don't really have to handle that. All you have to do is to add a new block and the logic to parse the `links` block in from config and cli. Something like [this](https://github.com/meain/lsd/blob/e3649fede8f7fc64041b709c56a3438ddbf813f0/src/flags/blocks.rs#L174)

src/flags/blocks.rs
meain: When just long is passed, don't show it. I think we'll keep the current default long blocks as they are for now.

TheAlakazam: Right now it doesn't show if only long is passed, but only shows the links if blocks is passed.

meain:
Similar to changing `inode_count` to `links`, can we also change the name `INodeCount` to `Links`.

src/color.rs
meain: Also, can we update the colors here to be similar to inode count. Adding colors over 16 might cause issues with the user's terminal colorscheme.

src/flags/blocks.rs
```diff
@@ -124,6 +124,7 @@ impl Blocks {
             Block::Size,
             Block::Date,
             Block::Name,
+            Block::Links,
```
meain, line 127: No links by default
```suggestion
            Block::Name,
```

src/color.rs
```diff
@@ -243,6 +251,8 @@ impl Colors {
         // INode
         m.insert(Elem::INode { valid: true }, Colour::Fixed(13)); // Pink
         m.insert(Elem::INode { valid: false }, Colour::Fixed(245)); // Grey
+        m.insert(Elem::Links { valid: true }, Colour::Fixed(134));
+        m.insert(Elem::Links { valid: false }, Colour::Fixed(222));
```
meain, line 255: We could reuse the `Inode` colors
```suggestion
        m.insert(Elem::Links { valid: true }, Colour::Fixed(13));
        m.insert(Elem::Links { valid: false }, Colour::Fixed(245));
```

meain (changes requested):
Just a few things that you missed.

meain:
Awesome, thanks a lot!

FischLu:
please add `links` option in the config.yaml example in the README.md file, today I tired to config my lsd with this function, but didn't see it anywhere till I see this pr

meain:
Added in https://github.com/Peltoche/lsd/pull/808

PR #467, opened 2020-12-27
adding a manpage

SirCipherz:
i wrote a manpage for the program, we still need to install it but i haven't figured out how the program install itself


meain:
Once we have a man page we can bundle that along with the releases. I believe the package maintainers should be able to use that to install the man pages.

The PR seems to be empty though. I think you might have forgotten to commit the file after edits.

edneville:
@SirCipherz, there's a ticket open to package up lsd. I am in the process of doing that now, and just need the manpage at the moment. Can you let me know if the manpage is around, or does it need to be created from new?

meain:
@edneville There is no manpage available for lsd as of now.

edneville:
Ok, I've raised #474 if you wish to consider that. If the man page turns up, maybe they can be combined or mine deprecated later if it is merged? Either way, it will help with packaging to have a man page.

meain:
Closing in favor of https://github.com/Peltoche/lsd/pull/474

Issue #363, opened 2020-05-02
Missing space between icon and folder name


andys8:
## Expected behavior

Space between icon and folder name

![image](https://user-images.githubusercontent.com/13085980/80864342-4c2ee400-8c82-11ea-8cd8-353479f31799.png)

## Actual behavior

There is no space between icon and folder name. Both overlap a bit.

![image](https://user-images.githubusercontent.com/13085980/80864347-5650e280-8c82-11ea-91d9-e2eb18a35872.png)

Is this an expected change?

## Info

Terminal: st (https://github.com/andys8/st)
Font: Iosevka Nerd Font-14

nik312123:
I can confirm that this has come up between versions 16.0 and 17.0 for me as well.

andys8:
I suspect the change from 2 spaces to 1 for one line is causing this behavior.

https://github.com/Peltoche/lsd/commit/200c5d2c68c7f0519d1eddfef2e30fe7dd8d11e5

andys8:
Over the past history this has changed some times. But with a single space there will be a problem with double width fonts. Will a change to 2 spaces be accepted and merged?

carlocaione:
+1 for this issue

meain:
Related: https://github.com/Peltoche/lsd/issues/316#issuecomment-641407582

kakulukia:
+1 for more spacing

This looks odd compared to colorls (coloring aside)

![Bildschirmfoto 2020-10-20 um 08 55 57](https://user-images.githubusercontent.com/96352/96551095-1920f100-12b2-11eb-8131-de7d00195410.png)

andys8:
Over at `exa` they introduced `EXA_ICON_SPACING`

https://github.com/ogham/exa/commit/51be9f4c43e61bbea61e70e1459e494207d1d979#diff-48b676e2b107da679512b793d5fd4cc4329f0c7c17a97cf6e0e3d1005b600b03

meain:
Now that we have support for config file, I guess we could add this as a config option `icon_spacing`.

meain:
This should now be available via https://github.com/Peltoche/lsd/pull/469 once `0.20.0` lands.

PR #469, opened 2021-01-04
Add `--icon-spacing` flag and `icon.spacing` config option

OrangeFran:
The spacing option takes a string and puts it between the icon and the name. The default would be one space: " ". Requested feature from the following issue: #363.

Example:

``` sh
lsd --icon-spacing "  " # I removed this now
```
``` yaml
icons:
    separator: "  "
```

Edit: I don't really know what `codecov/project` does and why it fails. Please explain.

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests (I hope the updated tests are sufficient)
- [x] Add changelog entry


zwpaper:
hi @OrangeFran , thanks for the PR, for the `codecov/project` part, you can click the `Detail` on the right side, it shows that you changed the `src/flags/icons.rs` and make the coverage go down 0.07, you should add more test for it.

image

meain:
I am thinking we could give the icon spacing option just in the cofig file and not in cli options as this is will be changed that often. Also, we could read the icon spacing value as an int rather than a string of spaces.

OrangeFran:
The cli option could be removed, I agree with you here.

I don't think that we should replace the string with an int. The string is just more customizable. I might use ` |` or something like that in the future. So I'd like to keep the string if possible.

meain:
Ahh, I see your point. In that case I think we should probably call it icon.separator instead of spacing.

OrangeFran:
Oh, I spelled `separator` wrong. Let me change that. Sorry.

src/icon.rs
```diff
@@ -370,10 +370,10 @@ mod test {
         File::create(&file_path).expect("failed to create file");
         let meta = Meta::from_path(&file_path, false).unwrap();
 
-        let icon = Icons::new(Theme::Fancy);
-        let icon = icon.get(&meta.name);
+        let icon = Icons::new(Theme::Fancy, " ".to_string());
+        let icon_str = icon.get(&meta.name);
 
-        assert_eq!(icon, format!("{}{}", "\u{f016}", ICON_SPACE)); // 
+        assert_eq!(icon_str, format!("{}{}", "\u{f016}", icon.icon_separator)); // 
```
meain, line 376: Can we also add a test for a non default icon separator is being used.

meain (commented):
Just a few more things

src/app.rs
```diff
@@ -53,6 +53,13 @@ pub fn build() -> App<'static, 'static> {
                 .number_of_values(1)
                 .help("Whether to use fancy or unicode icons"),
         )
+        // .arg(
```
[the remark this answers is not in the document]
OrangeFran: Sorry, I forgot to delete them.

src/config_file.rs
```diff
@@ -212,6 +213,9 @@ icons:
   # Which icon theme to use.
   # Possible values: fancy, unicode
   theme: fancy
+  # The string between the icons and the name.
+  # Possible values: any string
```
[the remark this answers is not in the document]
OrangeFran: Makes sense.

src/icon.rs
OrangeFran: Should I put the test there (`src/icon.rs`) or in the `src/flags/icon.rs` file, where I already created a test for the default case?

meain: I think we will keep it in `src/flags/icon.rs`.

OrangeFran: I made the changes.  You can review again now.

meain:
Looks good! 🎉

Issue #450, opened 2020-12-03
lsd should support changing time format by setting `TIME_STYLE` env like gnu ls


zxlxz:
set TIME_STYLE, eg:
```sh
export TIME_STYLE='+%F %T'
```
run ls:
```sh
ls /usr -l

drwxr-xr-x   6 root root 131072 2020-12-02 20:21:01 bin
```

run lsd:
```sh
lsd /usr -l

drwxr-xr-x root root 128 KB Wed Dec  2 20:21:01 2020 bin
```

meain:
This should now be available via https://github.com/Peltoche/lsd/pull/471 once `0.20.0` release lands.

PR #471, opened 2021-01-07
Add support for TIME_STYLE environment variable

999eagle:
This PR‌ adds support for parsing the date/time format from the TIME_STYLE environment variable as requested in #450.

For full GNU `ls` compatibility, an upgrade of `time` to version 0.2 is required as the currently used version 0.1.44 doesn't support nanosecond output as format string (placeholder `%N`).

The `date` option in the default configuration has been commented out to allow the environment to specify a date/time format. As `DateFlag` itself defaults to `DateFlag::Date`, this does not change default behaviour.

Changes to `Configurable<T>` allow other or future flags to be specified from the environment as well. The trait method `from_environment` has a default implementation for minimal change throughout the code.

EDIT: For some reason, the tests fail sometimes, even without recompiling. Probably because cargo's test harness runs multithreaded, but environment variables are modified process-wide. Using `std::env::vars()`, passing the iterator to every call to `Configurable::from_environment` and iterating through all environment variables would fix this issue, but this would be a performance hit as every call would require a new iterator to be created.

---
#### TODO

- [X] Use `cargo fmt`
- [X] Add necessary tests
- [X] Add changelog entry


src/flags.rs
```diff
@@ -111,6 +112,10 @@ where
     fn configure_from(matches: &ArgMatches, config: &Config) -> T {
         let mut result: T = Default::default();
 
+        if let Some(value) = Self::from_environment() {
```
meain, line 115: We should be able to invert the order and do an early return for all these cases.

src/flags.rs
```diff
@@ -98,10 +98,11 @@ pub trait Configurable<T>
 where
     T: std::default::Default,
 {
-    /// Returns a value from either [ArgMatches], a [Config] or a [Default] value. The first value
-    /// that is not [None] is used. The order of precedence for the value used is:
+    /// Returns a value from either [ArgMatches], a [Config], a [Default] or the environment value.
+    /// The first value that is not [None] is used. The order of precedence for the value used is:
     /// - [from_arg_matches](Configurable::from_arg_matches)
     /// - [from_config](Configurable::from_config)
+    /// - [from_environment](Configurable::from_environment)
```
meain, line 105: I am thinking the order could be arg -> env -> config. What do you think?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 ### Added
+- Add support for `TIME_STYLE` environment variable.
```
meain, line 9: Could you add your Github handle along with the change. Just to keep all the messages consistent.

src/flags/date.rs
```diff
@@ -193,4 +212,37 @@ mod test {
         c.classic = Some(true);
         assert_eq!(Some(DateFlag::Date), DateFlag::from_config(&c));
     }
+
```
meain, line 215: Could you also add a test which ensures the order of precedence.

src/flags/date.rs
```diff
@@ -83,6 +83,25 @@ impl Configurable<Self> for DateFlag {
             None
         }
     }
+
+    /// Get a potential `DateFlag` variant from the environment.
+    fn from_environment() -> Option<Self> {
+        if let Ok(value) = std::env::var("TIME_STYLE") {
+            match value.as_str() {
+                // TODO: time 0.2 supports the %N specifier for nanoseconds, proper format for
```
meain, line 91: We could bump the version of `time` to latest.

meain (changes requested):
Hey, thanks for the PR.

For the tests, I think we could use [serial_test](https://crates.io/crates/serial_test) crate to run just those test in series.

src/flags.rs
999eagle: I wasn't sure about the correct order to use. If you'd like to change the order around I can do that.

src/flags/date.rs
999eagle: Sadly `time 0.2` is a near complete rewrite of `time 0.1`‌ and is not compatible. I've switched to using `chrono` because `chrono-humanize` is being using anyway.

src/flags.rs
meain: I think we  will go with arg -> env -> config .

999eagle:
I've changed the things you mentioned. Tests now run using `serial-test`, flags are now parsed in the order arg -> env -> config and return early and the time is now stored using `chrono` instead of `time`. I've also added my GitHub name to the changelog.

999eagle:
The two failing checks in the last CICD build are two Windows builds that somehow failed to download the Rust toolchain. Can you please re-run the checks?

src/flags/date.rs
```diff
@@ -83,6 +83,24 @@ impl Configurable<Self> for DateFlag {
             None
         }
     }
+
+    /// Get a potential `DateFlag` variant from the environment.
+    fn from_environment() -> Option<Self> {
+        if let Ok(value) = std::env::var("TIME_STYLE") {
+            match value.as_str() {
+                "full-iso" => Some(Self::Formatted("%F %T.%f %z".into())),
+                "long-iso" | "iso" => Some(Self::Formatted("%F %R".into())),
```
meain, line 92: `iso` and `local`(`locale`?) seem to deviate from what gnu ls offers. I am not sure if these are things we can achieve with format string. If not, lets skip it.

---

Other than this (except for some minor difference in format string available in [gnu ls](https://www.cyberciti.biz/faq/linux-unix-formatting-dates-for-display/) and [chrono](https://docs.rs/chrono/0.4.19/chrono/format/strftime/index.html#specifiers) which I think should be fine), I think we are good to go.

999eagle: Hm you're right, `iso` and `locale` don't match GNU `ls` output. Both should be easy enough though, I'll fix those

999eagle: Okay, fixed both in a new commit

meain: Actually `iso` and `locale` is a bit more involved if the year is not current year (example below).

```
$ TIME_STYLE='locale' gls -l LICENSE README.md
-rw-r--r-- 1 meain staff 11359 Dec  8  2018 LICENSE
-rw-r--r-- 1 meain staff 10245 Jan 13 15:26 README.md

$ TIME_STYLE='iso' gls -l LICENSE README.md
-rw-r--r-- 1 meain staff 11359 2018-12-08  LICENSE
-rw-r--r-- 1 meain staff 10245 01-13 15:26 README.md
```

999eagle: You're right. I've only tested with files from 2020 and none of them showed the year. Seems like `iso` and `locale` are impossible with a simple format string. Still, with new enum values in `DateFlag` it should be possible.

999eagle: Okay, `iso` and `locale` now look like GNU `ls` as far as I can tell.

meain: `iso` changes the output format if it is [six months ago](https://github.com/coreutils/coreutils/blob/06f70d9ee69163fb2b18647963d6d6e81b6bd35d/src/ls.c#L4385) instead of an year. Plus, I think we could use `%e` instead of `%d` in the format string to avoid the leading zero in date for ones less than 10.

`locale` looks like it is a [bit more involved](https://github.com/coreutils/coreutils/blob/06f70d9ee69163fb2b18647963d6d6e81b6bd35d/src/ls.c#L2433). It needs to show the output based on the user's locale. The [libs](https://github.com/Koka/gettext-rs) that that help with this seems to be providing `ffi` bindings to the c lib and so I think we could probably separate it out to a different PR and skip in this.

```
$ LC_ALL='am_ET' TIME_STYLE='locale' gls -l AUTHORS
-rw-r--r-- 1 meain staff 3812 ጃንዩ 13 21:09 AUTHORS

$ LC_ALL='en_US.UTF-8' TIME_STYLE='locale' gls -l AUTHORS
-rw-r--r-- 1 meain staff 3812 Jan 13 21:09 AUTHORS
```

src/meta/date.rs
```diff
@@ -38,7 +38,9 @@ impl Date {
             DateFlag::Date => self.0.format("%c").to_string(),
             DateFlag::Relative => format!("{}", HumanTime::from(self.0 - Local::now())),
             DateFlag::ISO => {
-                if self.0 > Local::now() - Duration::days(365) {
+                // 356.2425 * 24 * 60 * 60 = 31556952 seconds per year
```
meain, line 41: A tiny typo, other than that we are good to go.
```suggestion
                // 365.2425 * 24 * 60 * 60 = 31556952 seconds per year
```

meain: Also, rebase with the current master. There seems to be a conflict in the `CHANGELOG.md` file.

999eagle: Done

meain:
Awesome, thanks!

Issue #356, opened 2020-04-18
OS 1332 Error under Windows


graphixillusion:
Hi there. I'm getting this error when i run **lsd** command. It occurrs when i switch to another drive: on the main **C:** drive, it works; if i run the command on **D:** i get the OS 1332 error with this complete error message: **No mapping between account names and security IDs (SID)**.

72siddharth:
I get the same error on windows when calling lsd in any drive other than 'C:'. 

`lsd: .: No mapping between account names and security IDs was done. (os error 1332)`

Tested it in both powershell and cmd, same error.
What exactly is the problem here? And is there any workaround?

isavegas:
I'm getting it in C: as well. `lsd` just straight up doesn't work on my Windows install. It appears to be related to having SMB shares mounted as drive letters. I'll do some research into the error and see if there's anything I can figure out on that front. 👍

Skips75:
I get the same error with lsd version 0.19.0 executing it on a removeable D: drive on Windows 10 Pro 20H2.
I installed lsd via scoop utility and I'm using it on a Powershell 5.1.19041.610.

PR #473, opened 2021-01-17
Report unidentified on Microsoft Windows for foriegn items

Starz0r:
Decided to smoke test what goes wrong when enumerating items in a directory, and it turns out that the function `lookup_account_sid` in `meta\windows_utils.rs` was the culprit! According to the Win32 API documentation on [LookupAccountSidW](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-lookupaccountsidw):
```
If the function cannot find an account name for the SID, GetLastError returns ERROR_NONE_MAPPED. This can occur if a network time-out prevents the function from finding the name. It also occurs for SIDs that have no corresponding account name, such as a logon SID that identifies a logon session.
```
This can generally happen if you create a file or folder on a system using NTFS and copy it over to another system using NTFS as information about the name and domain are preserved and not pummeled. This may or may not happen and is probably dependent on the filesystem and or OS configurations. Instead of failing this function outright, and since we already know the item exists if we got up to this point, we should just report that we don't actually know what permissions we can use or, account name, and domain owns this item. There is probably some way to calculate this properly since Explorer does it, but for now this is good enough.

Fixes https://github.com/Peltoche/lsd/issues/356

### DIFF
--- a/src/meta/windows_utils.rs
+++ b/src/meta/windows_utils.rs
@@ -255,10 +255,12 @@ unsafe fn lookup_account_sid(sid: *mut c_void) -> Result<(Vec<u16>, Vec<u16>), s
             // name_size and domain_size are already set, just loop
             continue;
         } else {
-            // Some other failure
-            // Assumptions: None (GetLastError shouldn't ever fail)
-            return Err(io::Error::from_raw_os_error(
-                winapi::um::errhandlingapi::GetLastError() as i32,
+            // Unknown account and or system domain identification
+            // Possibly foreign item coming from another machine
+            // TODO: Calculate permissions since it has to be possible if Explorer knows.
+            return Ok((
+                buf_from_os(OsStr::new("UNIDENTIFIED")),
+                buf_from_os(OsStr::new("UNIDENTIFIED")),
             ));
         }
     }


meain:
Hey, thanks for looking into this. I think we could report something like this for now until we have a better way of figuring out the proper user.

That said, I am keeping the PR open for a few days just in case someone who had the original issue in https://github.com/Peltoche/lsd/issues/356 / or knows a bit more about how Windows permission system works has any suggestion on how to go about it.

meain:
I have had some time to think about a better way that we could do this. Just hardcoding "UNIDENTIFIED" might not be the best idea. We could make the `Owner` struct optional for windows and have it display `-` in case of error with the greyed out color. This will make sure we keep the output consistent. (Also just `-` instead of `-\-` which would be the case with this).

Starz0r:
I hope this satisfies the requirements, at least temporarily!

Starz0r:
Actually re-read the requirements and saw that my previous commit did not actually satisfy those requirements. The error is now handled by the caller to ensure that the displayed metadata appears as `-` and not `-\-`.

Aankhen:
Would be great to see this merged! I keep running into the 1332 error all over the place.

src/meta/windows_utils.rs
```diff
@@ -61,26 +61,42 @@ pub fn get_file_data(path: &Path) -> Result<(Owner, Permissions), io::Error> {
         return Err(std::io::Error::from_raw_os_error(error_code as i32));
     }
 
+    // Format into domain\name format
```
zwpaper, line 64: maybe we should drop those comments?

src/meta/windows_utils.rs
```diff
@@ -61,26 +61,42 @@ pub fn get_file_data(path: &Path) -> Result<(Owner, Permissions), io::Error> {
...
+    match unsafe { lookup_account_sid(owner_sid_ptr) } {
+        Ok((n, d)) => {
+            let owner_name = os_from_buf(&n);
+            let owner_domain = os_from_buf(&d);
 
-    // Format into domain\name format
-    let mut owner = owner_domain.to_string_lossy().into_owned();
-    owner.push('\\');
-    owner.push_str(&owner_name.to_string_lossy());
+            owner = owner_domain.to_string_lossy().into_owned();
```
zwpaper, line 77: how about return the string direct by

```rust
format!("{}\\{}", owner_domain.to_string_lossy(), &owner_name.to_string_lossy())
```

src/meta/windows_utils.rs
```diff
@@ -61,26 +61,42 @@ pub fn get_file_data(path: &Path) -> Result<(Owner, Permissions), io::Error> {
...
+            owner.push('\\');
+            owner.push_str(&owner_name.to_string_lossy());
+        }
+        Err(_) => {
+            owner.push('-');
+        }
+    };
+
+    let mut group = String::new();
```
zwpaper, line 86: ditto

zwpaper (changes requested):

Starz0r:
> 
> 
> also, please check the edit locally, it seems there are syntax errors

Builds just fine on my end, looks like the issue is with the version of Rust that's used to build it.

PR #474, opened 2021-01-18
Adding lsd man page and md

edneville:
* lsd.md produces lsd.1 when used with pandoc.

For packaging convenience adding both.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Add support for changing the string between icon and name from [Finn Hediger](https://github.com/orangefran) [#363](https://github.com/Peltoche/lsd/issues/363)
 - Add support for `TIME_STYLE` environment variable from [999eagle](https://github.com/999eagle)
+- Add man page from [edneville](https://github.com/edneville)
 ### Changed
 ### Fixed
 - Fix handling blocks passed without -l in cli from [meain](https://github.com/meain)
--- a/doc/lsd.md
+++ b/doc/lsd.md
@@ -1,0 +1,142 @@
+---
+title: lsd
+section: 1
+header: User Manual
+footer: lsd <version>
+date: <date>
+---
+
+# NAME
+
+lsd - LSDeluxe
+
+# SYNOPSIS
+
+`lsd [FLAGS] [OPTIONS] [--] [FILE]...`
+
+# DESCRIPTION
+
+lsd is a ls command with a lot of pretty colours and some other stuff to enrich and enhance the directory listing experience.
+
+# OPTIONS
+
+`-a`, `--all`
+: Do not ignore entries starting with **.**
+
+`-A`, `--almost-all`
+: Do not list implied **.** and **..**
+
+`--classic`
+: Enable classic mode (no colours or icons)
+
+`-L`, `--dereference`
+: When showing file information for a symbolic link, show information for the file the link references rather than for the link itself
+
+`-d`, `--directory-only`
+: Display directories themselves, and not their contents (recursively when used with --tree)
+
+`-X`, `--extensionsort`
+: Sort by file extension
+
+`--help`
+: Prints help information
+
+`-h`, `--human-readable`
+: For ls compatibility purposes ONLY, currently set by default
+
+`--ignore-config`
+: Ignore the configuration file
+
+`-F`, `--classify`
+: Append indicator (one of \*/=>@|) at the end of the file names
+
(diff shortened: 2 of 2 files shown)


lsd.md
```diff
@@ -0,0 +1,140 @@
...
+header: User Manual
+footer: lsd 0.19.0
+date: 2021
+---
+
+# NAME
+
+lsd - LSDeluxe
+
+An ls command with a lot of pretty colours and some other stuff to enrich and enhance the directory listing experience.
```
meain, line 13: I think we could move this under description section.

lsd.md
```diff
@@ -0,0 +1,140 @@
...
+
+# SYNOPSIS
+
+`lsd [FLAGS] [OPTIONS] [--] [FILE]...`
+
+# DESCRIPTION
+
+`lsd`
+
```
meain, line 22: We should probably have `OPTIONS` section start here.

lsd.md
```diff
@@ -0,0 +1,140 @@
...
+`--size <size>...`
+: How to display size [default: default]  [possible values: default, short, bytes]
+
+`--sort <WORD>...`
+: sort by WORD instead of name [possible values: size, time, version, extension]
+
+# ARGS
+
+`<FILE>...`
```
meain, line 123: Maybe add in something like?
```
A file or directory to list.
```

lsd.md
```diff
@@ -0,0 +1,140 @@
...
+`lsd`
+: display listing for current directory
+
+`lsd /etc`
+: display listing of /etc
+
+`lsd -la`
+: display listing of current directory, including files starting with `.` and the current directory's entry.
+
+# CONTRIBUTIONS
```
meain, line 137: I think we could skip `CONTRIBUTIONS` section in the manpage.
But we should probably add `ENVIRONMENT` section for [`LS_COLORS`](https://github.com/sharkdp/fd/blob/9bc20e38b169e880eeeb6fba59c31cd201cf4081/doc/fd.1#L306) and [`XDG_CONFIG_HOME`](https://github.com/sharkdp/fd/blob/9bc20e38b169e880eeeb6fba59c31cd201cf4081/doc/fd.1#L312).

meain (changes requested):
Hey, thanks a lot for starting the effort on the man pages.

Can we move the man pages to `doc` folder. Plus I think it would be better to not have the generated file in the repo as that might get hard to maintain and keep in sync with the changes in the repo as people who contributes to the repo will have to regenerate the man pages.

Plus it would be better to have the version/year string to be a template(`<version>`) so that we could setup the proper values when trying to release it via Github Actions. This would mean that we introduce a build step anyways for the release.

lsd.md
```diff
@@ -0,0 +1,142 @@
...
+`lsd [FLAGS] [OPTIONS] [--] [FILE]...`
+
+# DESCRIPTION
+
+lsd is a ls command with a lot of pretty colours and some other stuff to enrich and enhance the directory listing experience.
+
+# OPTIONS
+
+`-a`, `--all`
+: Do not ignore entries starting with **.**.
```
meain, line 24: Think there is an extra dot here.

meain: Also move the file to `doc` folder just to keep the root clean.

lsd.md
```diff
@@ -0,0 +1,142 @@
...
+`lsd -la`
+: display listing of current directory, including files starting with `.` and the current directory's entry.
+
+# ENVIRONMENT
+
+`LS_COLORS`
+: see **dir_colors**
+
+`XDG_CONFIG_HOME`
+: location for optional config.yaml
```
meain, line 141: How about something like
```suggestion
: location for optional config file `config.yaml`
```

lsd.md
```diff
@@ -0,0 +1,142 @@
...
+`lsd /etc`
+: display listing of /etc
+
+`lsd -la`
+: display listing of current directory, including files starting with `.` and the current directory's entry.
+
+# ENVIRONMENT
+
+`LS_COLORS`
+: see **dir_colors**
```
meain, line 138: We could probably make it a bit more descriptive:

```suggestion
: Used to determine color for displaying filenames. See **dir_colors**.
```

lsd.md
```diff
@@ -0,0 +1,142 @@
...
+
+# ARGS
+
+`<FILE>...`
+: a file or directory to list [default: .]
+
+# EXAMPLES
+
+`lsd`
+: display listing for current directory
```
meain, line 127: Could you also make sure all these text start with capital letters. That seems to be what I see generally in other man pages.

meain:
Looks good, thanks!

PR #484, opened 2021-02-09
Add no-windows-permissions feature, to disable access/owner checks

vvuk:
Doing this on Windows is very slow; this adds a feature flag to disable doing these checks, making lsd as fast as any other directory listing tool.  Without this, running `lsd` on a directory with ~100 files takes around 600ms.  With this feature enabled, it takes 40ms.

Adds a bunch of compilation warnings unfortunately; the structure could be better to clean this up.

---
#### TODO

- [ ] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry

### DIFF
--- a/src/meta/mod.rs
+++ b/src/meta/mod.rs
@@ -212,14 +212,15 @@ impl Meta {
         };
 
         #[cfg(unix)]
-        let owner = Owner::from(&metadata);
-        #[cfg(unix)]
-        let permissions = Permissions::from(&metadata);
+        let (owner, permissions) = (Some(Owner::from(&metadata)), Some(Permissions::from(&metadata)));
 
-        #[cfg(windows)]
+        #[cfg(all(windows, feature = "no-windows-permissions"))]
+        let (owner, permissions) = (None, None);
+
+        #[cfg(all(windows, not(feature = "no-windows-permissions")))]
         let (owner, permissions) = windows_utils::get_file_data(&path)?;
 
-        let file_type = FileType::new(&metadata, symlink_meta.as_ref(), &permissions);
+        let file_type = FileType::new(&metadata, symlink_meta.as_ref(), &permissions.unwrap_or_default());
         let name = Name::new(&path, file_type);
         let inode = INode::from(&metadata);
         let links = Links::from(&metadata);
@@ -232,8 +233,8 @@ impl Meta {
             size: Size::from(&metadata),
             date: Date::from(&metadata),
             indicator: Indicator::from(file_type),
-            owner,
-            permissions,
+            owner: owner.unwrap_or_default(),
+            permissions: permissions.unwrap_or_default(),
             name,
             file_type,
             content: None,
--- a/src/meta/owner.rs
+++ b/src/meta/owner.rs
@@ -15,6 +15,13 @@ impl Owner {
     }
 }
 
+impl Default for Owner {
+    fn default() -> Owner
+    {
+        Owner { user: String::from(""), group: String::from("") }
+    }
+}
+
 #[cfg(unix)]
 impl<'a> From<&'a Metadata> for Owner {
     fn from(meta: &Metadata) -> Self {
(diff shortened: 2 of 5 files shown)


meain:
Having this behind a feature flag seems like a good compromise until we have a better solution, which I guess should be possible from https://github.com/Peltoche/lsd/pull/475#issuecomment-777101864
> One thing that's kind of interesting is that dir /q can pull this info quite quickly. cmd /c dir /q takes about 50ms for my test dir from another PR (unpatched lsd takes about 600ms; patched to hack out permission/ownership, 40ms). That pays the same process creation cost (cmd.exe) as lsd would. Might be useful to systrace & profile cmd.exe and see what it's doing, as dir is an internal call there.

What do you think about having all the values as `false` instead of `true`?

vvuk:
false works for me, done!

src/meta/windows_utils.rs
```diff
@@ -13,6 +13,32 @@ use super::{Owner, Permissions};
...
+    let permissions = Permissions {
+        user_read: false,
+        user_write: false,
+        user_execute: false,
+
+        group_read: false,
+        group_write: false,
+        group_execute: false,
+
+        other_read: true,
```
meain, line 29: We could make this as false as well.

src/meta/windows_utils.rs
```diff
@@ -13,6 +13,32 @@ use super::{Owner, Permissions};
 
 const BUF_SIZE: u32 = 256;
 
+#[cfg(feature = "no-windows-permissions")]
+pub fn get_file_data(path: &Path) -> Result<(Owner, Permissions), io::Error> {
+    let owner = Owner::new("?".to_string(), "?".to_string());
```
meain, line 18: I think it would be better to make owner optional in the struct in case of windows(we could probably have https://github.com/Peltoche/lsd/pull/473 added in after that). 
Also have the user/group as `-` with a greyish colors simliar to when there are no permissions. Something like this.
Screenshot 2021-02-13 at 8 45 14 PM

meain (changes requested):
Hey, just a few changes. Plus could also handle/suppress the unnecessary warnings.

src/meta/windows_utils.rs
vvuk: That's a great idea -- I'll get on it today.  I may split the no-windows-perms stuff into a separate .rs file, so that it's easier to cfg out stuff to avoid warnings too.

vvuk: (sorry, life got in the way this past week. will get to it when I can!)

Aankhen:
I tried this locally combined with #473 (after fixing a tiny merge conflict) and it makes a *huge* difference. Running lsd in a particular directory consistently took about 8 seconds before. Now it takes 479 *milliseconds*.

src/meta/mod.rs
```diff
@@ -212,14 +212,15 @@ impl Meta {
...
-        let owner = Owner::from(&metadata);
-        #[cfg(unix)]
-        let permissions = Permissions::from(&metadata);
+        let (owner, permissions) = (Some(Owner::from(&metadata)), Some(Permissions::from(&metadata)));
+
+        #[cfg(all(windows, feature = "no-windows-permissions"))]
+        let (owner, permissions) = (None, None);
 
-        #[cfg(windows)]
+        #[cfg(all(windows, not(feature = "no-windows-permissions")))]
```
zwpaper, line 220: how about checking this feature inside `windows_utils.rs`, so that we can keep checking `unix/windows` and drop the necessary to set `owner` and `permissions` as Optional, and keep the logic of Windows inside windows_utils.rs?

@meain

src/meta/owner.rs
```diff
@@ -37,10 +44,18 @@ impl<'a> From<&'a Metadata> for Owner {
 
 impl Owner {
     pub fn render_user(&self, colors: &Colors) -> ColoredString {
-        colors.colorize(self.user.clone(), &Elem::User)
+        if self.user.len() == 0 {
+            colors.colorize("?".to_owned(), &Elem::User)
```
zwpaper, line 48: as https://github.com/Peltoche/lsd/pull/484#discussion_r575677675 mentioned, this should be `-`, right?

@meain

zwpaper (changes requested):
I have one more question, as this is added as a build-time feature, should we need to release a new binary for this feature?

@meain 

and, @vvuk did you have some cycle to update this PR again, and we can make it happen after the discuss resolved?

or I can help to send a patch if you did not have time.

vvuk:
> and, @vvuk did you have some cycle to update this PR again, and we can make it happen after the discuss resolved?

I don't at the moment -- if you're able to take it over, please do!

PR #485, opened 2021-02-14
not showing . and .. when tree with all

zwpaper:
Not showing `.` and `..` when `--tree` with `--all`

fix https://github.com/Peltoche/lsd/issues/477

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry


meain:
Thanks for the PR. LGTM.

PR #488, opened 2021-02-15
Fix test involving config file for sorting

meain:
---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry

### DIFF
--- a/src/config_file.rs
+++ b/src/config_file.rs
@@ -276,6 +276,13 @@ symlink-arrow: ⇒
 "#;
 
 #[cfg(test)]
+impl Config {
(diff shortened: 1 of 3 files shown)


src/config_file.rs
```diff
@@ -148,6 +148,14 @@ impl Config {
     }
 }
 
+impl Config {
```
zwpaper, line 151: how about marking this `#[cfg(test)]` instead of `dead_code`

meain: Good call. Thanks.

Issue #468, opened 2020-12-28
Change the display of --tree when used alongside --long


jlagneau:
Currently, the tree is displayed on the leftmost side and it shift all the columns around depending on the depth of the tree, making the output sort of hard to read.

![image](https://user-images.githubusercontent.com/7811277/103192593-8451f980-48d9-11eb-841a-940986fb33a5.png)

Wouldn't it make sense to display the tree depth character alongside the file/directory names or at least on its own column to allow configuration to change that.

BinaryBen:
I've changed the config to:

```yaml
blocks:
  - name
  - size
  - date
```

This gives the effect you are after when using `lsd -l --tree` and you can add the permission to the end. It will change it for `lsd -l`  as well though

jlagneau:
That's pretty clever and simple.

Your config does fix the issue for file/folders names readability in the tree, and I didn't think of that, thanks a lot. I'll be using that in the mean time until there's something a bit better. But it's not really fixing the issue of columns shifting around depending on the depth of the tree though and that's part of my issue with the current implementation of tree with lsd and how hard to read some info end up. It does a fine job at aligning all the info in the long view usually, it makes searching for specific info a breeze and I wish it was the case with the tree view.

If you want a comparison with exa for example, here what it looks like (sorry for the icons missing, didn't install nerd fonts on that windows machine for them to work with wsl / ssh, pay no mind to them):

![image](https://user-images.githubusercontent.com/7811277/104879418-1f307780-595e-11eb-86a4-871b7f65bf9d.png)

I do very much prefer lsd for many things on my different machines (configurations, icons and all that) but I believe exa got the tree view mixed with long view right at the very least.

PR #489, opened 2021-02-15
show tree edge before name block

zwpaper:
show tree edge before name block

fix https://github.com/Peltoche/lsd/issues/468
---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -256,7 +256,8 @@ impl Colors {
         m.insert(Elem::Links { valid: true }, Colour::Fixed(13));
         m.insert(Elem::Links { valid: false }, Colour::Fixed(245));
 
-        m.insert(Elem::TreeEdge, Colour::Fixed(44)); // DarkTurquoise
+        // TODO add this after we can use file to configure theme
+        // m.insert(Elem::TreeEdge, Colour::Fixed(44)); // DarkTurquoise
         m
     }
 }
--- a/src/color.rs
+++ b/src/color.rs
@@ -52,6 +52,8 @@ pub enum Elem {
     Links {
         valid: bool,
     },
+
+    TreeEdge,
 }
 
 impl Elem {
@@ -254,6 +256,7 @@ impl Colors {
         m.insert(Elem::Links { valid: true }, Colour::Fixed(13));
         m.insert(Elem::Links { valid: false }, Colour::Fixed(245));
 
+        m.insert(Elem::TreeEdge, Colour::Fixed(44)); // DarkTurquoise
         m
     }
 }
--- a/src/display.rs
+++ b/src/display.rs
@@ -10,7 +10,7 @@ use terminal_size::terminal_size;
 use unicode_width::UnicodeWidthStr;
 
 const EDGE: &str = "\u{251c}\u{2500}\u{2500}"; // "├──"
-const LINE: &str = "\u{2502}  "; // "├  "
+const LINE: &str = "\u{2502}  "; // "│  "
 const CORNER: &str = "\u{2514}\u{2500}\u{2500}"; // "└──"
 const BLANK: &str = "   ";
 
@@ -81,6 +81,7 @@ fn inner_display_grid(
             &flags,
             &display_option,
             &padding_rules,
+            "",
         );
 
         for block in blocks {
@@ -156,7 +157,18 @@ fn inner_display_tree(
         direction: Direction::LeftToRight,
     });
 
-    for meta in metas.iter() {
+    for (idx, meta) in metas.iter().enumerate() {
+        let current_prefix = if depth > 0 {
+            if idx + 1 != last_idx {
+                // is last folder elem
+                format!("{}{} ", prefix, EDGE)
+            } else {
+                format!("{}{} ", prefix, CORNER)
+            }
+        } else {
+            prefix.to_string()
+        };
+
         for block in get_output(
             &meta,
             &colors,
@@ -164,6 +176,7 @@ fn inner_display_tree(
             &flags,
             &DisplayOption::FileName,
             &padding_rules,
+            &current_prefix,
         ) {
             let block_str = block.to_string();
 
@@ -178,32 +191,20 @@ fn inner_display_tree(
     let mut lines = content.lines();
 
     for (idx, meta) in metas.iter().enumerate() {
-        let is_last_folder_elem = idx + 1 != last_idx;
-
-        if depth > 0 {
-            output += prefix;
-
-            if is_last_folder_elem {
-                output += EDGE;
-            } else {
-                output += CORNER;
-            }
-            output += " ";
-        }
-
         output += &String::from(lines.next().unwrap());
         output += "\n";
 
         if meta.content.is_some() {
-            let mut new_prefix = String::from(prefix);
-
-            if depth > 0 {
-                if is_last_folder_elem {
-                    new_prefix += LINE;
+            let new_prefix = if depth > 0 {
+                if idx + 1 != last_idx {
+                    // is last folder elem
+                    format!("{}{}", prefix, LINE)
                 } else {
-                    new_prefix += BLANK;
+                    format!("{}{}", prefix, BLANK)
                 }
-            }
+            } else {
+                prefix.to_string()
+            };
 
             output += &inner_display_tree(
                 &meta.content.as_ref().unwrap(),
@@ -252,6 +253,7 @@ fn get_output<'a>(
     flags: &'a Flags,
     display_option: &DisplayOption,
     padding_rules: &HashMap<Block, usize>,
+    tree_prefix: &'a str,
 ) -> Vec<ANSIString<'a>> {
     let mut strings: Vec<ANSIString> = Vec::new();
     for block in flags.blocks.0.iter() {
@@ -279,19 +281,20 @@ fn get_output<'a>(
                 let s: String =
                     if flags.no_symlink.0 || flags.dereference.0 || flags.layout == Layout::Grid {
                         ANSIStrings(&[
+                            ANSIString::from(tree_prefix),
                             meta.name.render(colors, icons, &display_option),
                             meta.indicator.render(&flags),
                         ])
                         .to_string()
                     } else {
                         ANSIStrings(&[
+                            ANSIString::from(tree_prefix),
                             meta.name.render(colors, icons, &display_option),
                             meta.indicator.render(&flags),
                             meta.symlink.render(colors, &flags),
                         ])
                         .to_string()
                     };
-
                 strings.push(ColoredString::from(s));
             }
         };
--- a/src/display.rs
+++ b/src/display.rs
@@ -85,7 +85,7 @@ fn inner_display_grid(
     for meta in metas {
         // Maybe skip showing the directory meta now; show its contents later.
         if skip_dirs
-            && (matches!(meta.file_type, FileType::Directory{..})
+            && (matches!(meta.file_type, FileType::Directory { .. })
                 || (matches!(meta.file_type, FileType::SymLink { is_dir: true })
                     && flags.layout != Layout::OneLine))
         {
@@ -362,9 +362,11 @@ mod tests {
     use super::*;
     use crate::color;
     use crate::color::Colors;
-    use crate::icon;
     use crate::icon::Icons;
     use crate::meta::{FileType, Name};
+    use crate::Config;
+    use crate::{app, flags, icon, sort};
+    use assert_fs::prelude::*;
     use std::path::Path;
 
     #[test]
@@ -501,6 +503,16 @@ mod tests {
         }
     }
 
+    fn sort(metas: &mut Vec<Meta>, sorters: &Vec<(flags::SortOrder, sort::SortFn)>) {
+        metas.sort_unstable_by(|a, b| sort::by_meta(sorters, a, b));
+
+        for meta in metas {
+            if let Some(ref mut content) = meta.content {
+                sort(content, sorters);
+            }
+        }
+    }
+
     #[test]
     fn test_display_tree_with_all() {
         let argv = vec!["lsd", "--tree", "--all"];
@@ -511,11 +523,12 @@ mod tests {
         dir.child("one.d").create_dir_all().unwrap();
         dir.child("one.d/two").touch().unwrap();
         dir.child("one.d/.hidden").touch().unwrap();
-        let metas = Meta::from_path(Path::new(dir.path()), false)
+        let mut metas = Meta::from_path(Path::new(dir.path()), false)
             .unwrap()
             .recurse_into(42, &flags)
             .unwrap()
             .unwrap();
+        sort(&mut metas, &sort::assemble_sorters(&flags));
         let output = tree(
             &metas,
             &flags,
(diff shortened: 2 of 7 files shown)


meain:
@zwpaper , thanks for taking this up.

One thing that I am concerned with is handling `--tree` along with  `--blocks`. I am not sure how we should handle if someone does something like `lsd --tree --blocks permissions,name,date`.

Btw with this PR, if I am using date relative and the entire folder has a "smaller" date. The tree is misaligned.
```
.rw-r--r-- 2.4K a month ago   │  ├── indicators.rs
.rw-r--r-- 4.0K a month ago   │  ├── layout.rs
.rw-r--r-- 8.8K 2 minutes ago │  ├── recursion.rs
.rw-r--r-- 4.0K 2 minutes ago │  ├── size.rs
.rw-r--r--  16K 2 minutes ago │  ├── sorting.rs
.rw-r--r-- 2.1K a month ago   │  ├── symlink_arrow.rs
.rw-r--r-- 2.4K a month ago   │  ├── symlinks.rs
.rw-r--r-- 2.4K a month ago   │  └── total_size.rs
.rw-r--r-- 4.5K 2 days ago    ├── flags.rs
.rw-r--r--  18K 2 days ago    ├── icon.rs
.rw-r--r-- 2.7K 2 days ago    ├── main.rs
drwxr-xr-x 448B 2 days ago    ├── meta
.rw-r--r-- 8.1K 2 days ago   │  ├── date.rs
.rw-r--r-- 8.6K a month ago  │  ├── filetype.rs
.rw-r--r-- 2.6K a month ago  │  ├── indicator.rs
.rw-r--r-- 1.5K 2 months ago │  ├── inode.rs
.rw-r--r-- 1.5K 2 days ago   │  ├── links.rs
.rw-r--r-- 7.1K 2 days ago   │  ├── mod.rs
.rw-r--r--  16K 2 days ago   │  ├── name.rs
.rw-r--r-- 1.2K 2 days ago   │  ├── owner.rs
.rw-r--r-- 4.6K a month ago  │  ├── permissions.rs
.rw-r--r--  10K a month ago  │  ├── size.rs
.rw-r--r-- 3.3K a month ago  │  ├── symlink.rs
.rw-r--r--  11K 2 days ago   │  └── windows_utils.rs
.rw-r--r--  10K 2 weeks ago   └── sort.rs
```

zwpaper:
hi @meain , previously I have thought about making the `tree edge` as a block, but it would be weirder if users have to specify the `tree edge` option when using `--blocks`.

so I decided to make it stick with the name block, IMHO the `tree edge` seems to be useful when it stays with the name block.
if users specified `lsd --tree --blocks permissions,name,date`, they should be expecting the date block is not align in its column, or we can document it.

as the "smaller" date misaligned part, let me will dig deeper into it

meain:
I initially did think of having tree as a block as well. We can default it to the start of the block list if `--tree` is set but no `tree` block is given by the user. But to be frank that kinda felt like a bit too much.

Another issue with assigning `--tree` to just `name` is that if someone does `lsd --tree --blocks permission,date` it does not show a tree at all.

Here is one other idea, I am not fully convinced of this myself but thought I would throw it out there. We can have another flag `--tree-position` which can accept `start`, `name`. Also if the user gives `--tree-position` as `name` but does not have `name` in the blocks, we can change it to `start` automatically.

zwpaper:
I just feel like that both the `block` and `--tree-position` seems a little bit too heavy.

standing on the user case,  is there really a case that users did not show the `name` block?

meain:
I don't think there is a strong usecase for blocks without`name`, but `--tree` should probably draw a tree even if name is not specified in the blocks. How about we revert to drawing the tree symbols in the start if no name  is specified else with tree next to name? This wold work when user does an simple `-l` listing. One issue that I can think for this is that it might change the workflow of someone who might be using `name` neither at the start or at the end(not that bad though actually).

Screenshot 2021-02-18 at 12 24 11 PM

zwpaper:
there is a bug in the original tree column calculation, I am working on it and then the last case you mentioned, nerther start nor end , columns will stay aligned.

following your idea, if there is no name block, we add the tree prefix to the first column, making it align like a tree, while others align in column, ok?

zwpaper:
I have fixed the align problems including
1. check alignment for both dir and all sub dirs
2. check size padding for bot dir and all sub dirs

now there should not be align problems.

I will work onto show the tree edge when there is no `name` block

zwpaper:
hi @meain , this PR should be ready now.

with name:

```
λ lsd --tree --date relative --blocks permission,size,date,name
drwxr-xr-x 416 B  3 minutes ago  .
.rw-r--r--  11 KB 5 days ago    ├──  app.rs
.rw-r--r-- 7.3 KB a day ago     ├──  color.rs
.rw-r--r--  10 KB 3 days ago    ├──  config_file.rs
.rw-r--r-- 4.7 KB 3 weeks ago   ├──  core.rs
.rw-r--r--  20 KB 3 minutes ago ├──  display.rs
drwxr-xr-x 544 B  3 days ago    ├──  flags
.rw-r--r--  16 KB 3 days ago    │  ├──  blocks.rs
.rw-r--r-- 5.9 KB 3 weeks ago   │  ├──  color.rs
.rw-r--r-- 9.1 KB 3 weeks ago   │  ├──  date.rs
.rw-r--r-- 2.4 KB 3 weeks ago   │  ├──  dereference.rs
.rw-r--r-- 3.9 KB 3 weeks ago   │  ├──  display.rs
.rw-r--r--  11 KB 3 weeks ago   │  ├──  icons.rs
.rw-r--r-- 6.6 KB 3 weeks ago   │  ├──  ignore_globs.rs
.rw-r--r-- 2.4 KB 3 weeks ago   │  ├──  indicators.rs
.rw-r--r-- 4.0 KB 3 weeks ago   │  ├──  layout.rs
.rw-r--r-- 8.8 KB 3 weeks ago   │  ├──  recursion.rs
.rw-r--r-- 4.0 KB 3 weeks ago   │  ├──  size.rs
.rw-r--r--  16 KB 3 days ago    │  ├──  sorting.rs
.rw-r--r-- 2.1 KB 3 weeks ago   │  ├──  symlink_arrow.rs
.rw-r--r-- 2.4 KB 3 weeks ago   │  ├──  symlinks.rs
.rw-r--r-- 2.4 KB 3 weeks ago   │  └──  total_size.rs
.rw-r--r-- 4.5 KB 3 weeks ago   ├──  flags.rs
.rw-r--r--  18 KB 3 weeks ago   ├──  icon.rs
.rw-r--r-- 2.7 KB 3 weeks ago   ├──  main.rs
```

without name:

```
λ lsd --tree --date relative --blocks size,permission,date
416 B         drwxr-xr-x 4 minutes ago
├── 11 KB     .rw-r--r-- 5 days ago
├── 7.3 KB    .rw-r--r-- a day ago
├── 10 KB     .rw-r--r-- 3 days ago
├── 4.7 KB    .rw-r--r-- 3 weeks ago
├── 20 KB     .rw-r--r-- 4 minutes ago
├── 544 B     drwxr-xr-x 3 days ago
│  ├── 16 KB  .rw-r--r-- 3 days ago
│  ├── 5.9 KB .rw-r--r-- 3 weeks ago
│  ├── 9.1 KB .rw-r--r-- 3 weeks ago
│  ├── 2.4 KB .rw-r--r-- 3 weeks ago
│  ├── 3.9 KB .rw-r--r-- 3 weeks ago
│  ├── 11 KB  .rw-r--r-- 3 weeks ago
│  ├── 6.6 KB .rw-r--r-- 3 weeks ago
│  ├── 2.4 KB .rw-r--r-- 3 weeks ago
│  ├── 4.0 KB .rw-r--r-- 3 weeks ago
│  ├── 8.8 KB .rw-r--r-- 3 weeks ago
│  ├── 4.0 KB .rw-r--r-- 3 weeks ago
│  ├── 16 KB  .rw-r--r-- 3 days ago
│  ├── 2.1 KB .rw-r--r-- 3 weeks ago
│  ├── 2.4 KB .rw-r--r-- 3 weeks ago
│  └── 2.4 KB .rw-r--r-- 3 weeks ago
├── 4.5 KB    .rw-r--r-- 3 weeks ago
├── 18 KB     .rw-r--r-- 3 weeks ago
├── 2.7 KB    .rw-r--r-- 3 weeks ago
```

src/display.rs
```diff
@@ -32,7 +32,24 @@ pub fn grid(metas: &[Meta], flags: &Flags, colors: &Colors, icons: &Icons) -> St
...
+    let mut grid = Grid::new(GridOptions {
+        filling: Filling::Spaces(1),
+        direction: Direction::LeftToRight,
+    });
+
+    let padding_rules = get_padding_rules(&metas, flags);
+    let mut index = 0;
+    for (i, block) in flags.blocks.0.iter().enumerate() {
+        if let Block::Name = block {
+            index = i;
```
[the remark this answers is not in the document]
meain: We should probably also prevent someone from using something like `--blocks name,size,name` by making sure all of them are unique.

src/display.rs
```diff
@@ -252,51 +257,54 @@ fn get_output<'a>(
...
+                        meta.name.render(colors, icons, &display_option),
+                        meta.indicator.render(&flags),
+                    ])
+                } else {
+                    block_vec.extend(vec![
+                        meta.name.render(colors, icons, &display_option),
+                        meta.indicator.render(&flags),
+                        meta.symlink.render(colors, &flags),
+                    ])
+                };
```
meain, line 303: We could probably simplify this.

```suggestion
                block_vec.extend(vec![
                    meta.name.render(colors, icons, &display_option),
                    meta.indicator.render(&flags),
                ]);
                if !(flags.no_symlink.0 || flags.dereference.0 || flags.layout == Layout::Grid) {
                    block_vec.push(meta.symlink.render(colors, &flags))
                }
```

src/display.rs
```diff
@@ -143,80 +161,67 @@ fn inner_display_tree(
     flags: &Flags,
     colors: &Colors,
     icons: &Icons,
-    depth: usize,
-    prefix: &str,
-) -> String {
-    let mut output = String::new();
+    tree_depth_prefix: (usize, &str),
+    padding_rules: &HashMap<Block, usize>,
+    target_index: usize,
```
meain, line 166: `tree_index` might be a better name for this.

src/display.rs
```diff
@@ -143,80 +161,67 @@ fn inner_display_tree(
     flags: &Flags,
     colors: &Colors,
     icons: &Icons,
-    depth: usize,
-    prefix: &str,
-) -> String {
-    let mut output = String::new();
+    tree_depth_prefix: (usize, &str),
```
meain, line 164: IMO, I think it would be cleaner to keep `depth` and `prefix` as two separate values itself.

meain (changes requested):

src/display.rs
zwpaper: yes, I change this only because of the `too_many_arguments` warning, sigh...

and I can not figure out a better way to reduce the number of args, is there a better solution in your sight?

```
warning: this function has too many arguments (8/7), #[warn(too_many_arguments)] on by default
```

src/display.rs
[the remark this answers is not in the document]
zwpaper: IMHO, I think we can let users specify the duplicated block names, such as sometimes we can specify 2 name blocks near the blocks we care about for an easy reference

and I keep the tree edge in front of the first name block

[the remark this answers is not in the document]
meain: Ahh, makes sense.

meain:
@zwpaper The `--tree` test that is failing in all of these is actually something that caused me issues when releasing 0.20.0. Not exactly sure what is going on, but that test behaves different in macos vs linux. Also for some reason only randomly on this repo's CI alone. It does not fail in the CI run in my fork. I have removed this temporarily in master for now. Need to figure out what is going on with this one. I even try adding `--timesort` and it was still causing this issue.

zwpaper:
@meain I have found the problem for the test, sorting is done before display, so that timesort is ignored.

I have updated the code, and sorry for the force push, Have to rebase on to master

meain:
Thanks for looking into the tree test. I am a bit busy now, will try to get the reviewed before next week.

About force push, all good. In fact that is what you should be doing to fix issues in PRs. Just don't force push squashing every commit while working on the changes as that will make it harder to review in case someone has already taken a look at the previous commit.

SuperSandro2000:
If the branch is not rebased on master you can also view the diff between the force pushes https://github.com/Peltoche/lsd/compare/5a558275fd37fe8efe0fa0e78bfc97da1b89c3dd..6a8ae5bdaabf1b941438e9962cc45fa60b56d58a

src/color.rs
```diff
@@ -254,6 +256,8 @@ impl Colors {
         m.insert(Elem::Links { valid: true }, Colour::Fixed(13));
         m.insert(Elem::Links { valid: false }, Colour::Fixed(245));
 
+        // TODO add this after we can use file to configure theme
+        // m.insert(Elem::TreeEdge, Colour::Fixed(44)); // DarkTurquoise
```
meain, line 260: Wanted to do this a little while ago, I think we can make the color that is returned an `Option` and make this return `None`? I think we can take that up in another PR as well. This would let us remove the `#[allow(dead_code)]` above the struct definition.

src/display.rs
```diff
@@ -143,80 +162,67 @@ fn inner_display_tree(
...
         if meta.content.is_some() {
-            let mut new_prefix = String::from(prefix);
-
-            if depth > 0 {
-                if is_last_folder_elem {
-                    new_prefix += LINE;
+            let new_prefix = if tree_depth_prefix.0 > 0 {
+                if idx + 1 != last_idx {
+                    // is last folder elem
+                    format!("{}{}", tree_depth_prefix.1, LINE)
```
meain, line 205: Just noticed that we could make one tiny change.

```diff
diff --git a/src/display.rs b/src/display.rs
index a8b474c..132341b 100644
--- a/src/display.rs
+++ b/src/display.rs
@@ -202,9 +202,9 @@ fn inner_display_tree(
             let new_prefix = if tree_depth_prefix.0 > 0 {
                 if idx + 1 != last_idx {
                     // is last folder elem
-                    format!("{}{}", tree_depth_prefix.1, LINE)
+                    format!("{}{} ", tree_depth_prefix.1, LINE)
                 } else {
-                    format!("{}{}", tree_depth_prefix.1, BLANK)
+                    format!("{}{} ", tree_depth_prefix.1, BLANK)
                 }
             } else {
                 tree_depth_prefix.1.to_string()
```


|Before|After|
|---|---|
|Screenshot 2021-03-23 at 10 59 08 PM|Screenshot 2021-03-23 at 10 57 01 PM|

*Check the tree thing under `meta`*

meain (changes requested):

src/display.rs
jlagneau: I agree that it would be better, especially with folder icons. :+1:

src/color.rs
zwpaper: https://github.com/Peltoche/lsd/issues/506 issue created

meain:
LGTM

PR #493, opened 2021-03-07
Fix file ordering in tree test

meain:
This fixes a test which acts differently in macOS&Windows vs Linux.
Fixing this so that we can make a new release.

For some reason this only showed up the final release build somehow.


meain:
Completely removed the flaky test for now. Releasing 0.20.1 for now. Will look into the actual issue once the release is out and a patch release will be made fixing this if this is an issue in the code.

PR #501, opened 2021-03-14
Add separator to sample config file.

LHY-iS-Learning:
Add separator to sample config file

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,6 +228,10 @@ icons:
   # Which icon theme to use.
   # Possible values: fancy, unicode
   theme: fancy
+  # Separator between icon and the name
+  # Default to 2 spaces
+  separator: ' '
+
 
 # == Ignore Globs ==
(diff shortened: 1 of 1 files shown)


zwpaper:
hi @LHY-iS-Learning , thanks for updating the doc, I have one more concern, we specify the default values in the config file example, but looks like there is only one space in the separator value?
maybe we should make it the default 2 spaces

meain:
@zwpaper the default is single space ([ref](https://github.com/Peltoche/lsd/blob/7d04dcbeafaa53020d2800f30fec5f5d9922c386/src/flags/icons.rs#L172)). I think we will stick to that for now.

LHY-iS-Learning:
Em,,, my bad.
Update to one space comment.

meain:
Thanks @LHY-iS-Learning !

PR #511, opened 2021-04-17
Added Devicons theme

Lenan:
Added devicons theme for file and folder icons to match most vim/ranger themes.
References:
https://github.com/ryanoasis/vim-devicons
https://github.com/cdump/ranger-devicons2


meain:
Hey @Lenan , thanks for the PR but I don't think adding in multiple icon themes in the binary would be a good idea. We have plans to eventually add option to let the user configure all the icons. The idea with `unicode` was to let users who don't have any icon fonts to be able to use the basic set of icons at least.

PR #520, opened 2021-05-11
add --config-file flag

fabricio7p:
So here it is my progress so far. I figured that if we are assembling a custom Config object it should be parsed before parsing the arguments, correct?

I might be terribly wrong tho.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -74,9 +74,10 @@ pub fn build() -> App<'static, 'static> {
         )
         .arg(
             Arg::with_name("config-file")
-                .short("f")
                 .long("config-file")
                 .help("Provide a custom lsd configuration file")
+                .value_name("config-file")
+                .takes_value(true)
         )
         .arg(
             Arg::with_name("oneline")
--- a/src/app.rs
+++ b/src/app.rs
@@ -73,6 +73,12 @@ pub fn build() -> App<'static, 'static> {
                 .help("Ignore the configuration file"),
         )
         .arg(
+            Arg::with_name("config-file")
+                .short("f")
+                .long("config-file")
+                .help("Provide a custom lsd configuration file")
+        )
+        .arg(
             Arg::with_name("oneline")
                 .short("1")
                 .long("oneline")
--- a/src/config_file.rs
+++ b/src/config_file.rs
@@ -117,6 +117,12 @@ impl Config {
         serde_yaml::from_str::<Self>(yaml)
     }
 
+    #[cfg(not(windows))]
+    fn _custom_config_path(_file: String) -> Option<PathBuf> {
+        // somehow get file path from args here...
+        todo!()
+    }
+
     /// This provides the path for a configuration file, according to the XDG_BASE_DIRS specification.
     /// return None if error like PermissionDenied
     #[cfg(not(windows))]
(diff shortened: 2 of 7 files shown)


src/app.rs
```diff
@@ -72,6 +72,12 @@ pub fn build() -> App<'static, 'static> {
                 .long("ignore-config")
                 .help("Ignore the configuration file"),
         )
+        .arg(
+            Arg::with_name("config-file")
+                .short("f")
```
zwpaper, line 77: `-f` is taken by `GNU ls`:

> -f     do not sort, enable -aU, disable -ls --color

meain: I think just `--config-file` would be enough.

fabricio7p: Ok, will do.

src/main.rs
```diff
@@ -104,6 +104,9 @@ fn main() {
 
     let config = if matches.is_present("ignore-config") {
         Config::with_none()
+    } else if matches.is_present("config-file") {
+        // Here we should call new method to load from custom path
+        Config::default()
```
meain, line 109: Yup, pretty much. If we have a config file location known by default, we can just derive from `Config::from_path` instead of default.

src/config_file.rs
```diff
@@ -117,6 +117,12 @@ impl Config {
         serde_yaml::from_str::<Self>(yaml)
     }
 
+    #[cfg(not(windows))]
```
meain, line 120: Don't think you will have to implement this. The reason why we have the other functions as well the `#[cfg()]` blocks was so that we can figure out the default directory to use on each system.

meain:
Just so that you understand the flow.

In `main.rs` we check if we should ignore the config file or use default location. What happens currently when you choose [`Config::default`](https://github.com/meain/lsd/blob/master/src/config_file.rs#L152), it in turn calls the [`config_file_path`](https://github.com/meain/lsd/blob/master/src/config_file.rs#L123) func to get the path where it should load the file and uses that. What we will be doing here is instead of calling `default` we will just call [`from_file`](https://github.com/meain/lsd/blob/master/src/config_file.rs#L95) directly and pass the arg that the user provided.

fabricio7p:
Ok so now it's fully working, successfully reading custom config file.
 
But there's one thing bothering me that is we cannot provide a custom path with quotes, e.g.: `lsd --config-file "~/Documents/lsd.yaml"`

`from_file` returns None with no error output :open_mouth:. Is that expected?

meain:
@fabricio7p When you provide `~` as input, the shell will usually expand it to the values of `$HOME` before it passes it to the actual command that you are calling, but not when you put in quotes around it. This is to be expected.

For example even in gnu ls, `ls ~` will work but `ls "~"` will not.

meain:
Could you add in a man page entry, a test and update the changelog as well for this.

fabricio7p:
Let me know if that's ok

zwpaper:
@fabricio7p there is conflict on `CHANGELOG.md`, could you please rebase your changes onto master, and push it again?

fabricio7p:
> @fabricio7p there is conflict on `CHANGELOG.md`, could you please rebase your changes onto master, and push it again?

Sorry i haven't noticed. There it is

meain:
LGTM. Thanks a lot. :)

PR #522, opened 2021-05-20
Add Perl module and test script icons

mjgardner:
Adds Perl module (.pm) and test script (.t) icons

---
#### TODO

- [X] Use `cargo fmt`
- [ ] Add necessary tests
- [x] Add changelog entry


meain:
Thanks!

Issue #527, opened 2021-06-10
Custom date - expected string format


robole:
Hi,

Thank you for your hard work. It is a great app.

I wanted to make a customised date format. It's not clear what I should write:

```
classic: false

# == Date ==.
# Possible values: date, relative, +<date_format>
date: '%d %b %y %X'
```

I was expecting this config to give me a date such as "10 Jun 21 13:01:00".

What am I missing?

meain:
You are missing a `+` in front. We require that for a custom date format.

```yaml
date: '+%d %b %y %X'
```

robole:
hi @meain 

thanks for the feedback. it also required me to update to the latest for this to work. i had installed lsd as a snap which is still at v0.16.0.

PR #528, opened 2021-06-11
Clarify custom date format for config in README

robole:
As per #527 , it is a bit confusing what is written in the README for the `date` field of the configuration file. In my case, I was also using an older version - v0.16.0 as a snap - which also meant that even if I entered a correctly formatted custom date, it didnt work! 😅 This might save someone else in the future!

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ color:
 # This specifies the date format for the date column. The freeform format
 # accepts an strftime like string.
 # When "classic" is set, this is set to "date".
-# Possible values: date, relative, +<date_format>
+# Possible values: date, relative, '+<strftime_date_format>'
 date: date
 
 # == Dereference ==
--- a/README.md
+++ b/README.md
@@ -13,8 +13,8 @@
 - [Installation](#installation)
 - [Configuration](#configuration)
 - [External Configurations](#external-configurations)
-  * [Required](#required)
-  * [Optional](#optional)
+  - [Required](#required)
+  - [Optional](#optional)
 - [F.A.Q.](#faq)
 - [Contributors](#contributors)
 - [Credits](#credits)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -211,7 +211,7 @@ color:
 # This specifies the date format for the date column. The freeform format
 # accepts an strftime like string.
 # When "classic" is set, this is set to "date".
-# Possible values: date, relative, +<date_format>
+# Possible values: date, relative, '+<strftime_date_format>'
```
alexxbb, line 214: A small example of such value would be useful. `date: ?`

meain: @alexxbb That would help. 
@robole I think it would be better to keep it as `date_format` itself (we use the same name everywhere else). We can probably just mention what it is separately with an example.

robole: I updated the text @alexbb @meain 

```
# Possible values: date, relative, '+<date_format>' e.g. date: '+%d %b %y %X' will give you 
# a date like this: 17 Jun 21 20:14:55
```

README.md
```diff
@@ -211,7 +218,8 @@ color:
 # This specifies the date format for the date column. The freeform format
 # accepts an strftime like string.
 # When "classic" is set, this is set to "date".
-# Possible values: date, relative, '+<strftime_date_format>'
+# Possible values: date, relative, '+<date_format>' e.g. date: '+%d %b %y %X' will give you 
+# a date like this: 17 Jun 21 20:14:55
```
meain, line 222: How about something like this? Think it would be better to split it to two lines.

```suggestion
# Possible values: date, relative, '+<date_format>' 
# `date_format` will be a `strftime` formatted value. e.g. `date: '+%d %b %y %X'` will give you a date like this: 17 Jun 21 20:14:55
```

meain:
thanks! :)

PR #530, opened 2021-06-14
Remove extra backslash in issue template

The-Compiler:
Doesn't seem to belong there.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -13,7 +13,7 @@ assignees: ''
 - `echo $LS_COLORS`: 
 
 ## Expected behavior
-If applicable, add the output of the classic ls command (`\ls -la`) in order to show the buggy file/directory.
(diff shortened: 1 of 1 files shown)


TheTechRobo (approved):

meain:
The extra `\` helps with unaliasing `ls` in case the user had aliased `ls` as `lsd`.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -13,7 +13,7 @@ assignees: ''
 - `echo $LS_COLORS`: 
 
 ## Expected behavior
-If applicable, add the output of the classic ls command (`\ls -la`) in order to show the buggy file/directory.
```
Peltoche, line 16: Hello!

The '\' is used to ensure that we use the real ls and not some aliases

The-Compiler:
Oh, I wasn't aware! I've been doing `/bin/ls ...` for years :sweat_smile: - for some more context, see [this StackOverflow answer](https://stackoverflow.com/a/16506263/2085149). Thanks @meain and @Peltoche!

PR #536, opened 2021-07-24
Non-zero exit status in case of failure

### LINKED ISSUE #502: Return correct error code when error occurred like GNU ls
from https://github.com/Peltoche/lsd/issues/498 I found that `lsd` will not return an error code besides 0 even error occurred

we should implement this like the `GNU ls`

## Expected behavior
GNU ls output:

```
λ \ls a
ls: a: No such file or directory

✗ echo $?
1
```

## Actual behavior
lsd:

```
λ ls a
lsd: a: No such file or directory (os error 2).

λ echo $?
0
```

zarkone:
Mimics the value of exit status from GNU ls in case of failure:

> Exit status:
> 0      if OK,
> 1      if minor problems (e.g., cannot access subdirectory),
> 2      if serious trouble (e.g., cannot access command-line argument).

Closes #502

---
#### TODO

- [ ] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -72,14 +72,16 @@ impl Core {
         }
     }
 
-    pub fn run(self, paths: Vec<PathBuf>) {
-        let mut meta_list = self.fetch(paths);
+    pub fn run(self, paths: Vec<PathBuf>) -> i32 {
+        let (mut meta_list, exit_status) = self.fetch(paths);
 
         self.sort(&mut meta_list);
-        self.display(&meta_list)
+        self.display(&meta_list);
+
+        exit_status
     }
 
-    fn fetch(&self, paths: Vec<PathBuf>) -> Vec<Meta> {
+    fn fetch(&self, paths: Vec<PathBuf>) -> (Vec<Meta>, i32) {
         let mut meta_list = Vec::with_capacity(paths.len());
         let depth = match self.flags.layout {
             Layout::Tree { .. } => self.flags.recursion.depth,
@@ -87,11 +89,15 @@ impl Core {
             _ => 1,
         };
 
+        let mut path_not_found: bool = false;
+        let mut recourse_path_not_found: bool = false;
+
         for path in paths {
             let mut meta = match Meta::from_path(&path, self.flags.dereference.0) {
                 Ok(meta) => meta,
                 Err(err) => {
                     print_error!("{}: {}.", path.display(), err);
+                    path_not_found = true;
                     continue;
                 }
             };
@@ -106,6 +112,7 @@ impl Core {
                     }
                     Err(err) => {
                         print_error!("lsd: {}: {}\n", path.display(), err);
+                        recourse_path_not_found = true;
                         continue;
                     }
                 };
@@ -117,9 +124,15 @@ impl Core {
             for meta in &mut meta_list.iter_mut() {
                 meta.calculate_total_size();
             }
-        }
+        };
 
-        meta_list
+        let exit_status = match (path_not_found, recourse_path_not_found) {
+            (true, _) => 2,
+            (_, true) => 1,
+            _ => 0,
+        };
+
+        (meta_list, exit_status)
     }
 
     fn sort(&self, metas: &mut Vec<Meta>) {
--- a/src/core.rs
+++ b/src/core.rs
@@ -108,9 +108,10 @@ impl Core {
                 self.flags.layout == Layout::Tree || self.flags.display != Display::DirectoryOnly;
             if recurse {
                 match meta.recurse_into(depth, &self.flags) {
-                    Ok(content) => {
+                    Ok((content, exit_code)) => {
                         meta.content = content;
                         meta_list.push(meta);
+                        self.exit_code.set_if_greater(exit_code);
                     }
                     Err(err) => {
                         print_error!("{}: {}\n", path.display(), err);
--- a/src/core.rs
+++ b/src/core.rs
@@ -1,10 +1,10 @@
 use crate::color::{self, Colors};
-use crate::display;
 use crate::flags::{ColorOption, Display, Flags, IconOption, IconTheme, Layout, SortOrder};
 use crate::icon::{self, Icons};
 use crate::meta::Meta;
+use crate::{display, print_error};
 
-use crate::{print_output, sort, ExitCode, ExitStatus, PathError};
+use crate::{print_output, sort, ExitCode};
 use std::path::PathBuf;
 
 #[cfg(not(target_os = "windows"))]
@@ -21,7 +21,7 @@ pub struct Core {
     //display: Display,
     colors: Colors,
     sorters: Vec<(SortOrder, sort::SortFn)>,
-    exit_status: ExitStatus,
+    exit_code: ExitCode,
 }
 
 impl Core {
@@ -70,7 +70,7 @@ impl Core {
             //display: Display::new(inner_flags),
             colors: Colors::new(color_theme),
             icons: Icons::new(icon_theme, icon_separator),
-            exit_status: ExitStatus::new(),
+            exit_code: ExitCode::OK,
             sorters,
         }
     }
@@ -83,15 +83,7 @@ impl Core {
     }
 
     pub fn exit(&self) {
-        let errors = &self.exit_status.errors;
-        if !errors.is_empty() {
-            for error in errors {
-                println!("{}", error);
-            }
-        }
-
-        let exit_code = self.exit_status.code as i32;
-        std::process::exit(exit_code);
+        std::process::exit(self.exit_code as i32);
     }
 
     fn fetch(&mut self, paths: Vec<PathBuf>) -> Vec<Meta> {
@@ -106,9 +98,8 @@ impl Core {
             let mut meta = match Meta::from_path(&path, self.flags.dereference.0) {
                 Ok(meta) => meta,
                 Err(err) => {
-                    let path_error = PathError::new(path, err);
-                    self.exit_status
-                        .push_error(ExitCode::MajorIssue, path_error);
+                    print_error!("{}: {}.", path.display(), err);
+                    self.exit_code.set_if_greater(ExitCode::MajorIssue);
                     continue;
                 }
             };
@@ -122,9 +113,8 @@ impl Core {
                         meta_list.push(meta);
                     }
                     Err(err) => {
-                        let path_error = PathError::new(path, err);
-                        self.exit_status
-                            .push_error(ExitCode::MinorIssue, path_error);
+                        print_error!("{}: {}\n", path.display(), err);
+                        self.exit_code.set_if_greater(ExitCode::MinorIssue);
                         continue;
                     }
                 };
--- a/src/core.rs
+++ b/src/core.rs
@@ -3,7 +3,8 @@ use crate::display;
 use crate::flags::{ColorOption, Display, Flags, IconOption, IconTheme, Layout, SortOrder};
 use crate::icon::{self, Icons};
 use crate::meta::Meta;
-use crate::{print_error, print_output, sort};
+
+use crate::{print_output, sort, ExitCode, ExitStatus, PathError};
 use std::path::PathBuf;
 
 #[cfg(not(target_os = "windows"))]
@@ -20,6 +21,7 @@ pub struct Core {
     //display: Display,
     colors: Colors,
     sorters: Vec<(SortOrder, sort::SortFn)>,
+    exit_status: ExitStatus,
 }
 
 impl Core {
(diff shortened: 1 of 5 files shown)


zarkone:
Hello! 

I've noticed this "good first issue" label and decided give it a try. I haven't covered all the cases (including tests) -- metadata fails are not handled -- but decided to push it as is to ask if it is a good approach to return status like this. 

Thanks!

zwpaper:
Hi @zarkone, thanks for the implementation! you have almost done the work.

maybe some updates like below could make further exit code cases easier to implement:

1. create a struct, `ExitStatus` for example, containing the code and a vector of errors
2. add a macro or fn like the `print_error!` to add error code and error to the `ExitStatus`
3. print all the errors in the main and exit with the code

what's your idea? @meain

zarkone:
@zwpaper thanks for your feedback! Please check this recent commit https://github.com/Peltoche/lsd/pull/536/commits/3bca4047097b487b8b6675a5e7405ce56b1a71be if I understood your approach correct. If it is the case, I'll proceed with mod.rs and tests.

zarkone:
BTW, I've noticed that `Meta::recurse_into` from `mod.rs` never returns an `Error`.. I think because of this [Err branch from recurse in fetch](https://github.com/zarkone/lsd/blob/502-file-not-found-exit-status/src/core.rs#L124)  never gets executed if you test it with `400` subdir and `--tree`.  

Am I missing something?

meain:
@zwpaper  I don't think we should collect errors into a `vec` and print them in main. From what I can see, gnu ls just seems to print the errors as it goes through the directories. I think we can stick to that itself(especially useful in case of `tree` or other recursive items). Other than that, lets have the error returned as you mentioned in an `ExitStatus` enum, but that can be a simple enum just for clarity and convert that to the proper exit codes in `main`.

zarkone:
@meain this is a good observation! And even more -- GNU `ls` doesn't collect/buffer anything anything at all -- it prints all the info right away. `lsd`, on the contrary, uses buffered approach.

 It actually brings quite a significant delay if you compare `ls` and `lsd` on a folder with big-enough amount of file / recursion depth: `ls` will start printing right away while `lsd` will have a noticeable silent delay before printing. 

With the current buffered approach of `lsd`, in case of any access errors, you'll see this errors first -- before printing any kind of fileinfo. And then, after delay of buffering, actual fileinfo will be printed. Erorrs and fileinfo output will be separated from eachother, which is not the case for GNU `ls`.

To summarize, I think `lsd` should either buffer everything or not buffer anything at all, to be consistent and not to produce confusing output (which I described in a paragraph above).

meain:
I think in `lsd` we are printing the errors as we go while printing the actual output only at the end. But if we were to collect the errors and print it out at the end, the only change is that instead of printing the errors at the top, we will be print errors at the bottom which does not help much.

Also if we were to do something like:
```
ls existing-file1 non-existing-file existing-file2
```

gnu ls throws the error at top anyways.

```
ls: cannot access 'non-existing-file': No such file or directory
existing-file1  existing-file2
```

I don't personally see any use for collecting the errors into a vec as of now.

zarkone:
@meain sounds good, thanks for the feedback! 

How about the other points? About error propagation from mod.rs and the implementation in general?

zwpaper:
as I tested in `ubuntu 18.04` we can see `GNU ls` cached the error msg and show together like this:

test case, return 0:

```shell-session
lstest # ls -l
total 24
drwxr-xr-x 2 root root 4096 Jul 27 10:58 2
lrwxrwxrwx 1 root root    1 Jul 27 11:03 3 -> 3
drwxr-xr-x 2 root root 4096 Jul 27 10:58 4
drwxr-xr-x 2 root root 4096 Jul 27 10:58 6
drwxr-xr-x 2 root root 4096 Jul 27 10:58 a
lrwxrwxrwx 1 root root    9 Jul 27 11:05 b -> non-exist
drwxr-xr-x 2 root root 4096 Jul 27 10:58 c
drwxr-xr-x 2 root root 4096 Jul 27 10:58 e
```

symbol link dereference error, return 1:
```shell-session
lstest # ls -l -L
ls: cannot access '3': Too many levels of symbolic links
ls: cannot access 'b': No such file or directory
total 24
drwxr-xr-x 2 root root 4096 Jul 27 10:58 2
l????????? ? ?    ?       ?            ? 3
drwxr-xr-x 2 root root 4096 Jul 27 10:58 4
drwxr-xr-x 2 root root 4096 Jul 27 10:58 6
drwxr-xr-x 2 root root 4096 Jul 27 10:58 a
l????????? ? ?    ?       ?            ? b
drwxr-xr-x 2 root root 4096 Jul 27 10:58 c
drwxr-xr-x 2 root root 4096 Jul 27 10:58 e
```

symbol link dereference error, and non existed files, return 1:

```shell-session
lstest # ls -l -L . 5 d
ls: cannot access '5': No such file or directory
ls: cannot access 'd': No such file or directory
.:
ls: cannot access '3': Too many levels of symbolic links
ls: cannot access 'b': No such file or directory
total 24
drwxr-xr-x 2 root root 4096 Jul 27 10:58 2
l????????? ? ?    ?       ?            ? 3
drwxr-xr-x 2 root root 4096 Jul 27 10:58 4
drwxr-xr-x 2 root root 4096 Jul 27 10:58 6
drwxr-xr-x 2 root root 4096 Jul 27 10:58 a
l????????? ? ?    ?       ?            ? b
drwxr-xr-x 2 root root 4096 Jul 27 10:58 c
drwxr-xr-x 2 root root 4096 Jul 27 10:58 e
```

what's more, the `GNU ls` should also cache the result, it has to because it would rearrange the result for sorting and aligning, right?

zwpaper:
I also noticed that `GNU ls` would print errors before each folder result:

```shell-session
lstest # ls -l -L -R .
.:
ls: cannot access '3': Too many levels of symbolic links
ls: cannot access 'b': No such file or directory
total 24
drwxr-xr-x 2 root root 4096 Jul 27 11:24 2
l????????? ? ?    ?       ?            ? 3
drwxr-xr-x 2 root root 4096 Jul 27 10:58 4
drwxr-xr-x 2 root root 4096 Jul 27 10:58 6
drwxr-xr-x 2 root root 4096 Jul 27 10:58 a
l????????? ? ?    ?       ?            ? b
drwxr-xr-x 2 root root 4096 Jul 27 10:58 c
drwxr-xr-x 2 root root 4096 Jul 27 10:58 e

./2:
ls: cannot access './2/4': Not a directory
total 0
-rw------- 1 root root 0 Jul 27 10:58 2
l????????? ? ?    ?    ?            ? 4
```

meain:
@zwpaper Yeah, that does make sense. Since we will not be sorting across multiple directories, I don't think we have buffer all the output, but rather just per folder. But let us not get that done in this PR.  We could keep that as a separate PR.

@zarkone Yeah, you are right. As of now we are not returning any errors. Mostly it is just print the error and continue, but we can change that to return the `ExitCode` enum.

---

One other thing that I noticed when trying out what @zarkone mentioned is that `tree` does not recurse into the directory if the perm is `400`, but `lsd` as of now does. I believe we are currently emulating the behaviour of `ls -R` for tree which actually does recurse into the dirs.

Setup:
```
$ mkdir -p a/{b,c}
$ touch a/b/one a/c/two
$ chmod 400 a/b
```

`tree` (no errors here)
```
$ tree
.
└── a
    ├── b
    └── c
        └── two

3 directories, 1 file
```

`lsd --tree`
```
$ ll --tree
lsd: ./a/b/one: Permission denied (os error 13).

drwxr-xr-x  96B 42 seconds ago .
drwxr-xr-x 128B 42 seconds ago └── a
dr--------  96B now                ├── b
drwxr-xr-x  96B 35 seconds ago     └── c
.rw-r--r--   0B now                    └── two
```

`lsd -R`
```
$ ll -R a
lsd: a/b/one: Permission denied (os error 13).

dr-------- 96B 10 minutes ago b
drwxr-xr-x 96B 11 minutes ago c

a/b:

a/c:
.rw-r--r-- 0B 10 minutes ago two
```

zwpaper:
ok, it make sense to me now printing the error message without caching it.

as the `tree` and `400` part, I would stand for the current for the option printing error.

1. `ls -R` would print the same error
2. the error message would help to verify there are files but not accessible

what's more, it seems that `ls -R` would print the filename with question mark in the unaccessible part:

```shell-session
$ ls -l -R
.:
total 4
drwxrwxr-x 4 ubuntu ubuntu 4096 Jul 28 10:41 a

./a:
total 8
dr-------- 2 ubuntu ubuntu 4096 Jul 28 10:41 b
drwxrwxr-x 2 ubuntu ubuntu 4096 Jul 28 10:41 c

./a/b:
ls: cannot access './a/b/one': Permission denied
total 0
-????????? ? ? ? ?            ? one

./a/c:
total 0
-rw-rw-r-- 1 ubuntu ubuntu 0 Jul 28 10:41 two
```

---

we should create 2 other issues for those 2 discuss

zarkone:
> @zarkone Yeah, you are right. As of now we are not returning any errors. Mostly it is just print the error and continue, but we can change that to return the ExitCode enum.

@meain thanks, sounds good! 

@zwpaper, @meain:
Thanks for your investigations -- I was under impression that `ls` prints everything without any caching but It seems like I didn't red `ls` sources with enough attention. 

I'll remove error messages buffering, leaving only status code as an internal state of `core` then.

I also agree with @zwpaper that those issues are related, but out of scope of this PR -- just wanted to share my findings. I'm happy to jump on it as well after finishing this one -- if you feel it makes sense.

zarkone:
@zwpaper @meain updated the PR -- kindly ask you to check my yet unfinished code if it is a good direction overall.

I feel that some tests and some code at least in `mod.rs` is still missing -- but if what I have now looks good than I'll proceed further with it.

src/core.rs
```diff
@@ -20,6 +21,7 @@ pub struct Core {
     //display: Display,
     colors: Colors,
     sorters: Vec<(SortOrder, sort::SortFn)>,
+    exit_code: ExitCode,
```
meain, line 24: Don't think we should initialize it here. We can just have a return from the recursive parse.

src/core.rs
```diff
@@ -68,18 +70,23 @@ impl Core {
...
-    pub fn run(self, paths: Vec<PathBuf>) {
+    pub fn run(&mut self, paths: Vec<PathBuf>) {
         let mut meta_list = self.fetch(paths);
 
         self.sort(&mut meta_list);
-        self.display(&meta_list)
+        self.display(&meta_list);
+    }
+
+    pub fn exit(&self) {
```
meain, line 85: Don't think we need this. We can just call `std::process::exit` directly.

src/core.rs
```diff
@@ -92,6 +99,7 @@ impl Core {
                 Ok(meta) => meta,
                 Err(err) => {
                     print_error!("{}: {}.", path.display(), err);
+                    self.exit_code.set_if_greater(ExitCode::MajorIssue);
```
meain, line 102: Instead of having exit code be a field in `Core` and pass into fetch. It can just be a local value that is initialized here and then returned back to the run function similar to what you have done for `recurse_into`.

src/core.rs
```diff
@@ -68,18 +70,23 @@ impl Core {
...
-        self.display(&meta_list)
+        self.display(&meta_list);
+    }
+
+    pub fn exit(&self) {
+        std::process::exit(self.exit_code as i32);
     }
 
-    fn fetch(&self, paths: Vec<PathBuf>) -> Vec<Meta> {
+    fn fetch(&mut self, paths: Vec<PathBuf>) -> Vec<Meta> {
```
meain, line 89: These `mut` can be removed.

src/main.rs
```diff
@@ -115,7 +128,9 @@ fn main() {
         Config::default()
     };
     let flags = Flags::configure_from(&matches, &config).unwrap_or_else(|err| err.exit());
-    let core = Core::new(flags);
+    let mut core = Core::new(flags);
 
     core.run(inputs);
+
+    core.exit();
```
meain, line 135: `core.run` can return the `ExitStatus` and we can just directly call `std::process::exit` here at the end.

meain (changes requested):
Hey, just a quick change on how you approach this. I don't think `ExitStatus` is something that we could add into `Core`. I think it would be better if `ExitStatus` stays separate and we just return it  from `fetch`/`run` at the end of mutating a value inside `self`.

meain:
Closing in favor of https://github.com/Peltoche/lsd/pull/739 . Thanks for starting the work on this PR :D

PR #540, opened 2021-08-16
Add .mjs extension (ESM)

neezer:
Reuse the JS icon for [ESM Modules with `.mjs` extension](https://nodejs.org/api/esm.html#esm_modules_ecmascript_modules).


zwpaper:
thanks for contributing!

PR #544, opened 2021-08-21
Add NO_COLOR env support

AnInternetTroll:
This allows the program to not output colours when using the `NO_COLOR` environmental variable. For example when running `NO_COLOR=1 lsd`. To read more about this environmental variable take a look at https://no-color.org/

#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased] - ReleaseDate
 ### Added
 - update minimal rust version to 1.42.0 from [zwpaper](https://github.com/zwpaper) [#534](https://github.com/Peltoche/lsd/issues/534)
-- [`NO_COLOR`](https://no-color.org/) environment variable support. 
+- [`NO_COLOR`](https://no-color.org/) environment variable support from [AnInternetTroll](https://github.com/aninternettroll)
 ### Changed
 - Change size to use btyes in classic mode from [meain](https://github.com/meain)
 - Show tree edge before name block or first column if no name block from [zwpaper](https://github.com/zwpaper) [#468](https://github.com/Peltoche/lsd/issues/468)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased] - ReleaseDate
 ### Added
 - update minimal rust version to 1.42.0 from [zwpaper](https://github.com/zwpaper) [#534](https://github.com/Peltoche/lsd/issues/534)
+- [`NO_COLOR`](https://no-color.org/) environment variable support. 
 ### Changed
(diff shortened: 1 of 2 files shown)


meain:
@AnInternetTroll Could you fix the clippy lints in the CI?

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased] - ReleaseDate
 ### Added
 - update minimal rust version to 1.42.0 from [zwpaper](https://github.com/zwpaper) [#534](https://github.com/Peltoche/lsd/issues/534)
+- [`NO_COLOR`](https://no-color.org/) environment variable support.
```
meain, line 10: Could you change it to the format:
`<change> from <person>`

src/flags/color.rs
```diff
@@ -89,6 +90,14 @@ impl Configurable<Self> for ColorOption {
             None
         }
     }
+
+    fn from_environment() -> Option<Self> {
+        if env::var("NO_COLOR").is_ok() {
```
meain, line 95: Should we be checking if the value is truthy? Should we sill be disabling colors if `NO_COLORS` is set to 0?

meain (commented):
Thanks a lot for working on this.

In the `no-colors.org` page, they mention that the config file should have higher precedence over env variable. I think that pattern should stay true for all cases. Could you flip the order of env and config in https://github.com/meain/lsd/blob/70ec1d6c51d209ea3e29ba2ae8208bbbc45690b2/src/flags.rs#L118-L124

src/flags/color.rs
AnInternetTroll: > All command-line software which outputs text with ANSI color added should check for the presence of a NO_COLOR environment variable that, when present (**regardless of its value**), prevents the addition of ANSI color.
https://no-color.org/
If someone doesn't want `NO_COLOR` then they should just unset it.

AnInternetTroll:
> Could you flip the order of env and config in https://github.com/meain/lsd/blob/70ec1d6c51d209ea3e29ba2ae8208bbbc45690b2/src/flags.rs#L118-L124

I'm super duper new to rust. I tried flipping those over and it just breaks the `NO_COLOR` code. I don't know why or how to fix this so if anyone else could that would be awesome. Thanks

meain:
@AnInternetTroll Not sure what you mean by the code breaks? Is it failing to compile?

When you flip them, if you have a config value set for color, then we will color it irrespective of the env flag.

AnInternetTroll:
image
image

meain:
Could you check your `lsd` config and make sure there is no color related settings in there. I gave it a shot locally and it was working fine for me.

AnInternetTroll:
I don't have an lsd config. It should be all default behaviour. 
image

meain:
Ahh, I just realized that we apply a base config if the user has not added in anything in their personal config. That might be a bit more work to flip them then. I'll create another issue to track that. Thanks for working on this. :)

PR #556, opened 2021-10-01
Add support for MISSING  / mi= missing symlink target LS_COLORS value.

dsully:
Added support for the MISSING / mi= dircolors variable for broken symlink targets.

See: https://github.com/arcticicestudio/nord-dircolors for an example setting of "MISSING".

Before:
CleanShot 2021-10-01 at 10 41 37@2x

After:
CleanShot 2021-10-01 at 10 41 53@2x

---
- [X] Use `cargo fmt`
- [] Add necessary tests
- [X] Add changelog entry


meain:
Thank you. I was not aware that this was a thing somehow.

meain (approved):

PR #558, opened 2021-10-07
Cope with chrono panics due to weird dates

### LINKED ISSUE #529: Panic: No such local time
- os: Archlinux
- `lsd --version`: lsd 0.20.1
- `echo $TERM`: xterm-kitty
- `echo $LS_COLORS`: quite long and hopefully not relevant :slightly_smiling_face: 

## Expected behavior
If applicable, add the output of the classic ls command (`\ls -la`) in order to show the buggy file/directory.

I have an Onyx Boox ereader which I'm mounting via [jmtpfs](https://github.com/JasonFerrara/jmtpfs) - looking at how old that project is I probably should switch to something else, but alas...

For some reason, this results in files apparently created in the year 4437052:

```
$ /bin/ls --full-time
total 0
drwxr-xr-x 26 florian florian 0 4437052-10-03 00:35:12.000000000 +0200 'Internal storage'
```

I'd expect `lsd` to show the file in the same way, or perhaps just show an "invalid time" or something.

## Actual behavior

lsd crashes with `thread 'main' panicked at 'No such local time', /build/.cargo/registry/src/github.com-1ecc6299db9ec823/chrono-0.4.15/src/offset/mod.rs:173:34`.

Stack:

```
   0: std::panicking::begin_panic
   1: <chrono::datetime::DateTime<chrono::offset::local::Local> as core::convert::From<std::time::SystemTime>>::from
   2: lsd::meta::Meta::from_path
   3: lsd::meta::Meta::recurse_into
   4: lsd::core::Core::run
   5: lsd::main
```

palfrey:
Should fix #529 

---
#### TODO

- [x] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry

### DIFF
--- a/src/meta/date.rs
+++ b/src/meta/date.rs
@@ -309,12 +309,13 @@ mod test {
     }
 
     #[test]
+    #[cfg(all(not(windows), target_arch = "x86_64"))]
     fn test_bad_date() {
         // 4437052 is the bad year taken from https://github.com/Peltoche/lsd/issues/529 that we know is both
         // a) high enough to break chrono
         // b) not high enough to break SystemTime (as Duration::MAX would)
         let end_time =
-            time::SystemTime::UNIX_EPOCH + time::Duration::new(343702 * 365 * 24 * 60 * 60, 0);
+            time::SystemTime::UNIX_EPOCH + time::Duration::new(4437052 * 365 * 24 * 60 * 60, 0);
         let colors = Colors::new(ThemeOption::Default);
         let date = Date::from(end_time);
 


palfrey:
> Thanks for working on this :)
> 
> Could you also try to add a test for this as well.

This is remarkably difficult actually. So, `touch` won't let you set anything after around May 10 2446 (i64::MAX seconds after the unix epoch), and that doesn't appear to be high enough to break chrono. I'm messing around with https://github.com/alexcrichton/filetime to see if I can get around that by avoiding using the `touch` tool, but that is looking like it's got similar limits although SystemTime might be able to go higher, so fiddling around in that...

palfrey:
> > Thanks for working on this :)
> > Could you also try to add a test for this as well.
> 
> This is remarkably difficult actually. So, `touch` won't let you set anything after around May 10 2446 (i64::MAX seconds after the unix epoch), and that doesn't appear to be high enough to break chrono. I'm messing around with https://github.com/alexcrichton/filetime to see if I can get around that by avoiding using the `touch` tool, but that is looking like it's got similar limits although SystemTime might be able to go higher, so fiddling around in that...

Ah, not a touch issue, an ext4 issue! See https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/fs/ext4/ext4.h?id=a4dad1ae24f850410c4e60f22823cba1289b8d52#n732

palfrey:
@meain @zwpaper I've figured out how to add a test for this and I think I've fixed all the issues that got raised above. Thoughts?

zwpaper:
thanks so much for working on this and finding the ext4 issue!

PR #561, opened 2021-10-07
Up minimum version in build.rs to 1.43.0

### LINKED ISSUE #560: build.rs says 1.42.0 is enough, but it's not
## Expected behavior
lsd can be built with 1.42.0 because that's what the `build.rs` will bug out without

## Actual behavior
`cargo +1.42.0 build` with current master gets you
```
error: attributes are not yet allowed on `if` expressions
   --> /home/palfrey/.cargo/registry/src/github.com-1ecc6299db9ec823/crossterm-0.21.0/src/command.rs:123:9
    |
123 |         #[cfg(windows)]
    |         ^^^^^^^^^^^^^^^

error: attributes are not yet allowed on `if` expressions
   --> /home/palfrey/.cargo/registry/src/github.com-1ecc6299db9ec823/crossterm-0.21.0/src/command.rs:224:5
    |
224 |     #[cfg(windows)]
    |     ^^^^^^^^^^^^^^^

error: unexpected `self` parameter in function
  --> /home/palfrey/.cargo/registry/src/github.com-1ecc6299db9ec823/crossterm-0.21.0/src/style/stylize.rs:15:29
   |
15 |             fn $method_name(self) -> Self::Styled {
   |                             ^^^^ not valid as function parameter
...
87 |     stylize_method!(reset Attribute::Reset);
   |     ---------------------------------------- in this macro invocation
   |
   = note: `self` is only valid as the first parameter of an associated function

error: expected one of `async`, `const`, `crate`, `default`, `extern`, `fn`, `pub`, `type`, or `unsafe`, found `fn reset(self: (/*ERROR*/)) -> Self::Styled {
    self.attribute(Attribute::Reset)
}`
   --> /home/palfrey/.cargo/registry/src/github.com-1ecc6299db9ec823/crossterm-0.21.0/src/style/stylize.rs:146:62
    |
87  |     stylize_method!(reset Attribute::Reset);
    |     ---------------------------------------- in this macro invocation
...
146 |     ($(#[doc = $doc:expr] $item:item)*) => { $(#[doc = $doc] $item)* };
    |                                                              ^^^^^ expected one of 9 possible tokens

error: aborting due to 4 previous errors

error: could not compile `crossterm`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

palfrey:
Fixes #560

---
#### TODO

- [x] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -18,11 +18,11 @@ use std::process::exit;
 include!("src/app.rs");
 
 fn main() {
-    match version_check::is_min_version("1.43.0") {
+    match version_check::is_min_version("1.43.1") {
         Some(true) => {}
         // rustc version too small or can't figure it out
         _ => {
-            writeln!(&mut io::stderr(), "'lsd' requires rustc >= 1.43.0").unwrap();
+            writeln!(&mut io::stderr(), "'lsd' requires rustc >= 1.43.1").unwrap();
             exit(1);
         }
     }


meain:
We had actually changed the min version to `1.43.1` for our CI a while back. I think we just missed this, thanks for bringing it up. We can keep it the same here as well instead of `1.43.0`.

palfrey:
> We had actually changed the min version to `1.43.1` for our CI a while back. I think we just missed this, thanks for bringing it up. We can keep it the same here as well instead of `1.43.0`.
> 
> https://github.com/Peltoche/lsd/blob/29d9c3b10e3d036eb1baa056e003006d7137d302/.github/workflows/CICD.yml#L9

Done

Issue #533, opened 2021-06-28
thousands separator for the --size=bytes option would be very useful


peter-joo:
- OS: Linux 5.12.9-1-MANJARO x86_64 GNU/Linux
- `lsd --version`:  lsd 0.20.1
- `echo $TERM`:  xterm-256color
[...]

PR #569, opened 2021-10-15
Support for bytes separated by thousand with comma 

areq212:
Solution for task #533.
Used `num-format` library as it's really simple to use and seems to be performant. It also allows further extension to support OS locale and grouping not only by thousands but also in Indian style. Used `Locale.en` as it's most obvious. 

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry

### DIFF
--- a/src/meta/size.rs
+++ b/src/meta/size.rs
@@ -1,8 +1,8 @@
 use crate::color::{ColoredString, Colors, Elem};
 use crate::flags::{Flags, SizeFlag};
+use num_format::ToFormattedString;
 use std::fs::Metadata;
 use std::iter::repeat;
-use num_format::{ToFormattedString};
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum Unit {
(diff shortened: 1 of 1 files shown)


meain:
Hey, thanks for working on this.
I remember you mentioning about the cland dependency for picking system locale, but without that I'm not sue how useful this will be. Any thought @zwpaper ?

I found a related issue in exa for this https://github.com/ogham/exa/issues/554.

arkadiuszbielewicz:
Exactly as @meain mentioned, my idea was to use only part which doesn't require `Clang` (it's only used for `SystemLocale` feature) and with some other code, detect system locale. Should I do it within this PR?

meain:
What do you have in mind to get system locale? I think it is better gone in as a single PR just to validate the idea. @arkadiuszbielewicz

arkadiuszbielewicz:
Hi, I've checked available options and finally I've decided to use `num_format` with `with-system-locale` feature, but only for Unix family. In case of Windows that feature is not present, and fallback to `Locale.en` is used.

zwpaper:
Hi @arkadiuszbielewicz, could you please fix the CI error, and rebase onto the master, also squash your commits for a git clean history.

PR #583, opened 2021-11-17
Allow human-readable to be set multiple times

rtuin:
This improves compatibility with `ls`.

For example: in my `~/.zshrc` I have the following:
```bash
alias ls='lsd -lFah'
```

When I'm off-guard and type `ls -lah` I now get the following:

```
error: The argument '--human-readable' was provided more than once, but cannot be used multiple times

USAGE:
    lsd --all --color <color>... --date <date>... --group-dirs <group-dirs>... --human-readable --icon <icon>... --icon-theme <icon-theme>... --ignore-glob <pattern>... --classify --long --size <size>...

For more information try --help
```

This should solve that.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -98,6 +98,7 @@ pub fn build() -> App<'static, 'static> {
             Arg::with_name("human_readable")
                 .short("h")
                 .long("human-readable")
+                .multiple(true)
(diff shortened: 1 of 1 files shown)


zwpaper:
@rtuin thanks for contributing!

Issue #549, opened 2021-08-24
Handling default/none values in theme config


zwpaper:
After https://github.com/Peltoche/lsd/pull/452 merged, lsd now supports yaml-based theme configuration.

but a completed configuration file is needed, if any items are missed, the whole theme would fall back to the default one.

we should update the theme struct to `Option` and allow some items missed.

we should discuss that the missed fields should indicate which one color:
1. the default foreground color, like no color
2. the default theme color, falls back the missed fields to the default item.
3. the default error color, could be used by invalid items

PR #591, opened 2021-11-28
theme: :sparkles: add default and update the others optional

zwpaper:
fix https://github.com/Peltoche/lsd/issues/549

difference:

1. use a `default` value as fallback color instead of falling back to the default theme, one may copy the default theme value from README
2. no options for the terminal default foreground color, still suffering how to represent this

---
#### TODO

- [ ] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Added support for the MISSING / mi= dircolors variable for broken symlink targets.
 - Add support for theme from [zwpaper](https://github.com/zwpaper) [#452](https://github.com/Peltoche/lsd/pull/452)
+- Update theme to support partial themes [zwpaper](https://github.com/zwpaper) [#591](https://github.com/Peltoche/lsd/pull/591)
 - Update minimal rust version to 1.42.0 from [zwpaper](https://github.com/zwpaper) [#534](https://github.com/Peltoche/lsd/issues/534)
 - [`NO_COLOR`](https://no-color.org/) environment variable support from [AnInternetTroll](https://github.com/aninternettroll)
 ### Changed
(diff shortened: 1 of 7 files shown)


meain:
Hi @zwpaper, I have not yet fully gone through the codebase but I was thinking that a better way to implement the idea of default colors is to have it use the colors from the default theme when no color is mentioned. For example, if the user has not defined the color for the size block, it will lookup the color in the default theme for size.

The idea is that they can mostly use the default colors, but just update only the ones that they don't like. The way you have done it will mean the user will have to define every single color if they need to change the colors.

---

On a related note, I was thinking that we could improve how we pick up the theme file. We could make it so that if a file called `theme.yaml` is present in the `.config/lsd` directory, we use that for the theme. This might reduce confusion on the theme file naming as seen in some issues. But yeah, this can pushed for later.

zwpaper:
Hi @meain, thanks for the input, as I mentioned in the description, I also wondering how should we handle the default color.

This PR could be a small Proof of Concept and discussion.

a fallback to the default theme is something like `patch the default theme`, I was thinking how about letting people create some brand new themes, and sharing or switching themes. 

so that a fixed `default` value seems better than falling back to the default theme, also, people could copy and modify the default theme from README if this is what they want.

this also answers your second question, why I did not choose to use `theme.yaml` as the theme file name and create a theme dir for the themes.

meain:
I think it is better to still fallback to the default theme instead of using a `default` value.

1) It makes it so that if someone want to change just the date column, they don't have to copy paste the full theme
2) There are no values that **has** to be there (`default`), everything becomes optional
3) `default` will usually be the base color which will anyways not suit if we were to introduce something new like a new optional block.

Besides, people who plan to create full new things can always just make sure they cover everything by starting off with the default theme and updating all the values.

zwpaper:
@meain now this feature should works as expected

meain (approved):
This looks great. The only thing that I would add here an option for the user to choose the default terminal foreground (`reset` in crossterm, something like a `none` option in theme file) so that they will not have to specify back/while etc and have it work in both light and dark setups, but that came come in later.

I think we could probably go for the next release as well now.

Issue #580, opened 2021-11-16
lsd --group-dirs not working correctly


WilliamBehrens:
- os: Void linux
- `lsd --version`: 0.20.1 
- `echo $TERM`: foot
- `echo $LS_COLORS`: di=1;35:fi=0;97:ln=4;36:or=5;31:ex=1;0;44:su=30;41:ow=30;42:st=30;44:

## Expected behavior
Running `lsd --group-dirs` should, atleast by the name, group all the directories together in the output.

## Actual behavior
But when I run it I get cases where it doesn't work as intended. one is shown below:
![scrn-6](https://user-images.githubusercontent.com/35979547/142068360-3027bdba-265a-4d83-9b7e-d1d4e392790d.png)

meain:
The option is actually supposed to be used as:

```
lsd --group-dirs first
lsd --group-dirs last
```

If you don't provide anything it just takes it as `--group-dirs none` which is the default. I think it would be better made so that it will error out if you just say `--group-dirs` without specifying how.

WilliamBehrens:
ah, thank you, I'll leave the issue open for you to do your patch

zwpaper:
fixed by https://github.com/Peltoche/lsd/pull/592

PR #592, opened 2021-11-29
made the group-dirs flag require a value

juansc:
group-dirs flag requires a value
---
Hi! I saw this issue https://github.com/Peltoche/lsd/issues/580. Should be straightforward. Let me know if there's any other changes I need to make.
#### TODO

- [ ] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry


zwpaper:
@juansc thanks for contributing!

Issue #554, opened 2021-09-26
Use a different time format for recent files


manno:
- os: Mac
- `lsd --version`: 0.20.1
- `echo $TERM`: xterm-256color
- `echo $LS_COLORS`: 

## Expected behavior

GNU `ls` and BSD `ls` display a different time format for recently modified files.
Old files show the year, new files the time.

For GNU `ls` this is configurable by the `TIME_STYLE` variable:
> The TIME_STYLE argument can be full-iso, long-iso, iso, locale, or +FORMAT.  FORMAT is interpreted like in date(1).  If FORMAT is FORMAT1<newline>FORMAT2, then FORMAT1 applies to non-recent files and FORMAT2 to recent files.

```
% ls -la                                                                                                    .
total 0
drwxr-xr-x  5 mm staff  160 Sep 26 15:10 .
drwx------ 32 mm staff 1024 Aug  8 13:04 ..
-rw-r--r--  1 mm staff    0 Nov 11  2018 oldfile
-rw-r--r--  1 mm staff    0 Sep 26 15:08 recentfile
% stat oldfile recentfile                                                                       .
  File: oldfile
Access: 2020-11-27 22:05:59.000000000 +0100
Modify: 2018-11-11 16:04:57.000000000 +0100
Change: 2021-09-26 15:10:32.982365307 +0200
 Birth: 2018-11-11 16:04:57.000000000 +0100
  File: recentfile
Access: 2021-09-26 15:08:22.236659000 +0200
Modify: 2021-09-26 15:08:22.236659000 +0200
Change: 2021-09-26 15:08:22.236667905 +0200
 Birth: 2021-09-26 15:08:22.236580024 +0200
```

## Actual behavior

LSD always show the same time format.

```
% lsd -l                                                                                                    .
.rw-r--r-- mm staff 0 B Sun Nov 11 16:04:57 2018  oldfile
.rw-r--r-- mm staff 0 B Sun Sep 26 15:08:22 2021  recentfile
```

Is it possible to show the time for recent files and the year for old files?

meain:
We currently do not have support for specifying different time format for recent files, but that would be a great addition.

PR #600, opened 2021-12-16
Date format supports recent/non-recent format

manno:
Support newline separator to specify a different date format for recent files.

This is similar to the `TIME_STYLE=iso` format, however the formats can be freely chosen.
(Also `iso` can only be set via the env var, not the command line.)

refers to #554

---
#### TODO

- [ ] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry


manno:
```
newline='
'
lsd -l --date "+%h %d %H:%M$newline%h %d %Y"
```

manno:
I'm new to Rust, let me know if something is off about this PR.
Should I also update the documentation?

src/app.rs
```diff
@@ -290,51 +290,63 @@ fn validate_date_argument(arg: String) -> Result<(), String> {
...
 
 pub fn validate_time_format(formatter: &str) -> Result<(), String> {
-    let mut chars = formatter.chars();
-    loop {
-        match chars.next() {
-            Some('%') => match chars.next() {
-                Some('.') => match chars.next() {
-                    Some('f') => (),
+    let str = formatter.to_string();
+    let vec: Vec<&str> = str.split('\n').collect();
```
meain, line 294:
```suggestion
    let vec: Vec<&str> = formatter.split('\n').collect();
```
No need to create an additional owned string.

src/app.rs
```diff
@@ -290,51 +290,63 @@ fn validate_date_argument(arg: String) -> Result<(), String> {
...
+
+    if vec.len() > 2 {
+        return Err("invalid format, can only contain one newline separator".to_owned());
```
meain, line 297:
```suggestion
        return Err("invalid date format, cannot contain more than two entries".to_owned());
```
We can probably make that the error is from date format as it can also come from the env variable `TIME_STYLE`.

src/meta/date.rs
```diff
@@ -65,7 +65,17 @@ impl Date {
                         val.format("%F").to_string()
                     }
                 }
-                DateFlag::Formatted(format) => val.format(format).to_string(),
+                DateFlag::Formatted(format) => {
+                    let vec: Vec<&str> = format.split('\n').collect();
+
+                    if vec.len() == 1 || *val > Local::now() - Duration::seconds(15_778_476) {
```
meain, line 71:
```suggestion
                    if vec.len() == 1 || *val < Local::now() - Duration::seconds(15_778_476) {
```

meain: Make sure to change the tests as well.

meain (changes requested):
Sorry for the late response, I was a bit occupied. Just a few suggestions on the PR. Also we could add this info to the [man pages](https://github.com/Peltoche/lsd/blob/master/doc/lsd.md) (probably also mention the `TIME_STYLE` env variable).

manno:
I rebased and added information to the README and Changelog.

manno:
I think it would be better if formatted date would support newlines. Maybe something like this: https://github.com/manno/lsd/commit/487d959ffa72ee86d7052924324c6e6248ddecbd

manno:
> Apologies for commenting on a closed PR – I’ve been using your fork from your last comment ([manno@487d959](https://github.com/manno/lsd/commit/487d959ffa72ee86d7052924324c6e6248ddecbd)) as I've found this feature very nice to have.
> 
> I was wondering if you’re happy with the current approach, and whether you’d be planning to open a new PR to upstream it!

I am super happy with the approach and using it myself. But I created that patch with AI assistance. So I would need to check if that's okay to PR. I do think, the code is fine and the tests too, but I am no Rust expert.

PR #601, opened 2021-12-18
Add an icon for .iso files

NyleCohen:
This PR adds an icon for .iso files with the "﫭" ('faed') disk icon.

### DIFF
--- a/src/icon.rs
+++ b/src/icon.rs
@@ -344,6 +344,7 @@ impl Icons {
         m.insert("zsh", "\u{f489}"); // ""
         m.insert("zsh-theme", "\u{f489}"); // ""
         m.insert("zshrc", "\u{f489}"); // ""
+        m.insert("iso", "\u{faed}"); // "﫭"
 
         m
     }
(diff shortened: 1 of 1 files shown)


zwpaper:
Hi @NyleCohen, thanks for contributing! We arranged the icons by their names, how about moving the `iso` to stay with `i*`

NyleCohen:
Sorry, moved it into alphebetical order

zwpaper:
Seems there is something wrong when tested with my local Hack Nerd Font, I will dig into it and then check this PR.

Seems there is an unexpected space:

image

zwpaper:
@meain I check in my Emacs, the `iso` icon works as expected, but not the terminal and Iterm, could you help to test this icon?

meain:
@zwpaper Looks like it is the same thing for me. Works in Emacs, but not in my terminal emulator. This looks like a terminal rendering bug though. Maybe we can switch to `f7c9`?

zwpaper:
then this is really weird, switching to `f7c9` is a solution but seems the `faed`﫭  suits the case better. 🤨

PR #611, opened 2022-01-25
Add --group-directories-first as an alias for --group-dirs=first

networkException:
`ls` from coreutils has the `--group-directories-first` flag as the only
directory grouping option. This pull request improves compatibility so that
users switching to `lsd` can continue using `--group-directories-first`.

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry


meain (changes requested):
Yeah, I guess it makes sense to have it in. Thanks for the PR. Just a few tiny nitpicks.

meain:
There were few tiny tweaks which I have fixed and pushed to your branch.
The PR looks good otherwise. Will merge it as soon as the CI is done. Thanks! :)

meain (approved):

PR #627, opened 2022-02-22
add proper debian support

ShalokShalom:
Its a shame, this isnt packaged by Debian

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,19 +18,19 @@ The project is heavily inspired by the super [colorls](https://github.com/athity
 
 Install the patched fonts of powerline nerd-font and/or font-awesome. Have a look at the [Nerd Font README](https://github.com/ryanoasis/nerd-fonts/blob/master/readme.md) for more installation instructions. Don't forget to setup your terminal in order to use the correct font.
 
-| OS/Distro                       | Command                                                                                                           |
-| ------------------------------- | ----------------------------------------------------------------------------------------------------------------- |
-| Archlinux                       | `pacman -S lsd`                                                                                                   |
-| Fedora                          | `dnf install lsd`                                                                                                 |
-| Gentoo                          | `sudo emerge sys-apps/lsd`                                                                                        |
-| macOS                           | `brew install lsd` or `sudo port install lsd`                                                                     |
-| NixOS                           | `nix-env -iA nixos.lsd`                                                                                           |
-| FreeBSD                         | `pkg install lsd`                                                                                                 |
-| NetBSD or any `pkgsrc` platform | `pkgin install lsd` or `cd /usr/pkgsrc/sysutils/lsd && make install`                                              |
-| Windows                         | `scoop install lsd`                                                                                               |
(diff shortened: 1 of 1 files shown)


meain:
I am not fully certain that this change is bringing a lot more value. For users who knows what is going on, it is better to point them to the release page. For users that are not aware of what is going on, downloading this to their current directory and not deleting at the end might not be a good idea.

But yeah, Debian package would have been great.

ShalokShalom:
> I am not fully certain that this change is bringing a lot more value. For users who knows what is going on, it is better to point them to the release page. For users that are not aware of what is going on, downloading this to their current directory and not deleting at the end might not be a good idea.
> 
> But yeah, Debian package would have been great.

I can delete it. The point is, the before it said to dpkg first, and then to download the package.

Dpkg works on a local package, so this is quite non-sensical. 

Can you agree, to a simply script, that we write and host, so the actual command become short enough?

I would do that.

zwpaper:
I am thinking about the version part of the link and package name, it may add some additional concern when releasing.

can we figure out some way to drop the version part?

zwpaper:
I am thinking about the version part of the link and package name, it may add some additional concern when releasing.

can we figure out some way to drop the version part?

meain:
For getting the latest version, we can rely on Github API `https://api.github.com/repos/Peltoche/lsd/releases/latest` ([ref](https://github.com/meain/homebrew-meain/blob/5f7d4d3df7d4fa658138ecf4f558d4178c9641f8/update#L8)).

I guess scripts would work. What I am wondering is weather to go with the deb package or use [lsd-0.21.0-x86_64-unknown-linux-gnu.tar.gz](https://github.com/Peltoche/lsd/releases/download/0.21.0/lsd-0.21.0-x86_64-unknown-linux-gnu.tar.gz) in the script. *We should also handle the architecture as well.*

Using the non deb version would let us use the script in all systems and not just in debian/ubuntu.

zwpaper:
@meain script would be a good option for general installation. as for here the Debian part, the `deb` may still be the choice.

ShalokShalom:
> Using the non deb version would let us use the script in all systems and not just in debian/ubuntu.

Well, then we arrive at AppImage, slightly

ShalokShalom:
What about https://github.com/zyedidia/eget

meain:
> What about https://github.com/zyedidia/eget

@ShalokShalom I don't think requiring users to install another binary just to install lsd is a good idea.

meain:
@zwpaper What I had in mind for the Ubuntu/Debian part was something like:
```
`curl script.sh|bash`
Or install using the `.deb` from [release](...) page.
```

This way, those who want to use `.deb` still gets to know that there is a deb package, but the install script stays universal.

zwpaper:
How about letting the script be one of the installation methods, it does not have to be specified to Debian users.

meain:
@zwpaper That works too.

meain:
Closing this, we can start another thread for possibly adding the  script that we talked about.

PR #629, opened 2022-02-23
Icons: added magnet and torrent

Kara-Zor-El:
I chose to add this since there wasn't really an icon for torrent and magnet files and figured that  would be a good fit.
Feel free to merge it or not, I thought it might be nice but if not just close it

---
#### TODO

- [ ] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry


zwpaper:
Hi @Kara-Zor-El thanks for the PR, it would be great for us to add more icons, feel free to open PRs like this.

as for the icon, I would like to discuss some more, you used is 
image

but it seems not that fit the `torrent` and `magnet` file usage, how about one of below

image

especially the `nf-mdi-download-network`,  it looks like downloading in a link which could mean part of the P2P network.

this also seems to be a good choice

image

meain:
How about this for magnet? (`f076`)

Kara-Zor-El:
Hi,
thank you @zwpaper . I have made some changes based on your and @meain 's request

Issue #630, opened 2022-02-27
Option to show permissions in Octal


pbowyer:
For all the reasons given on https://github.com/ogham/exa/issues/996 and https://news.ycombinator.com/item?id=29617182 it would be really nice if `lsd` had an option to display the `permission` column in octal rather than the traditional text format.

meain:
That sounds like a pretty useful addition. The flag could be something like `--premission <type>` as we have things like `--date <type>`.

meain:
I was looking into the exa implementation. As of now, they are adding it as a separate block. My initial thought was to replace how we display the permissions right now and change them to octal. I have feeling that doing it this way might not look right though as there is both prefix(filetype) and suffix(things like exec sticy) tacked onto permissions.

meain:
I as thinking of implementing it as an alternate for the current rwx format in that it will essentially take the place of current permissions section. One issue with this might be that it might look kinda odd when we have the file type marker to the left of it. 

Not that bad, but mentioning it here in case someone has any opinions on it.
![2022-02-27-23-59-05](https://user-images.githubusercontent.com/14259816/155894949-c1bb87b6-95da-4657-9586-3c14cc5ad454.png)


*FYI, `exa` implements it as another block*

pbowyer:
Thanks for looking at this! I like the flag your propose, and I'd thought of it taking the place of the current permissions section, and not an extra block.

However the file type marker... I hadn't considered that (I always look at the icon when using `lsd` rather than the marker) and I do find it hard to read, particularly the ones with `l` in front (brain says: "Is that a 1?"). 

Would adding a space between them be enough? We are adding something non-standard with octal formatting and scripts aren't going to parse this output, so it could be okay. It does change the number of columns for `awk` though, which could be annoying - except I never awk this listing as the date/time breaks into so many bits...

meain:
> However the file type marker... I hadn't considered that (I always look at the icon when using `lsd` rather than the marker) and I do find it hard to read, particularly the ones with `l` in front (brain says: "Is that a 1?").

There will be a lot of people who don't use icons or colors and this is useful for them. Actually it is not that bad when I use my personal theme whcih is a lot more subtle.

![2022-02-28-10-15-23](https://user-images.githubusercontent.com/14259816/155925600-7fb2ded7-4dd8-4843-a481-5fe8291270c8.png)

[...]

PR #634, opened 2022-03-13
Octal permissions

meain:
Allows you to switch showing the permissions in octal format. Added a theme key  for octal as well.

Fixes https://github.com/Peltoche/lsd/issues/630

Possible future additions:
- Add option to disable the showing file-type (`d` `l`) before the octal perms.
- Option to theme the file-permissions

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry

### DIFF
--- a/src/config_file.rs
+++ b/src/config_file.rs
@@ -1,6 +1,7 @@
 use crate::flags::display::Display;
 use crate::flags::icons::{IconOption, IconTheme};
 use crate::flags::layout::Layout;
+use crate::flags::permission::PermissionFlag;
 use crate::flags::size::SizeFlag;
 use crate::flags::sorting::{DirGrouping, SortColumn};
 use crate::flags::{ColorOption, ThemeOption};
@@ -36,6 +37,7 @@ pub struct Config {
     pub layout: Option<Layout>,
     pub recursion: Option<Recursion>,
     pub size: Option<SizeFlag>,
+    pub permission: Option<PermissionFlag>,
     pub sorting: Option<Sorting>,
     pub no_symlink: Option<bool>,
     pub total_size: Option<bool>,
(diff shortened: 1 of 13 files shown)


zwpaper:
Hi @meain, I found a consistent question, we use plurals somewhere and singular some others, should we unify them?

like singular `icon`, plural `permissions`, even more, plural `icons` in the config file.

meain:
Good catch. I will to switch to `permission` in this PR. I think it would be a good idea to unify them but we probably should not break existing configs. We can switch to using singular everywhere as that would be less things to change. We can start off with supporting both, for example `icon` and `icons` in the config file and add a deprecation warning when using `icons` instead of `icon`. Other plural ones that I see are `links` and `indicators` but for them I think plural is necessary (maybe?). Any other offenders you can think of?

zwpaper:
you got me exactly right! the only difference in my mind is that, IMO, 
1. the `indicators` in the config file indicate whether to show indicators or not, not the real indicators like `blocks`, so that we can also use the singular to keep consistency, and keep plural `blocks`.
2. like `1.`, `links` in theme could use singular like the `inode` in theme.

meain:
Ahh, makes sense. Let's start off with supporting both singular and plural for these initially (update all the documentation around it to singular) and then deprecate them after one or two versions.

src/config_file.rs
```diff
@@ -1,6 +1,7 @@
 use crate::flags::display::Display;
 use crate::flags::icons::{IconOption, IconTheme};
 use crate::flags::layout::Layout;
+use crate::flags::permissions::PermissionFlag;
```
zwpaper, line 4: how about also change this to singular?

meain: I was kinda on the fence on that one. On one had, using `permissions` will keep it consistent with `src/meta/permissions.rs`, but on the other hand everything inside it is `permission`.
Anyways, I don't really have a strong opinion on it. I'll change it to permission.

PR #639, opened 2022-03-15
Display acl and selinux data

mmatous:
Hi, I added  some requested features regarding access control. There are some differences with coreutils, but I think they're justified. `ls` doesn't display labels on systems where given LSM isn't enabled. E.g. it reports file `a.txt` with label `?` on one system and with `system_u:object_r:postgresql_db_t:s0` on another. `lsd` will display info consistently.

`ls` therefore doesn't have the problem of what to do with multiple labels, `lsd` will concat selinux and smack labels with +. This case should basically never occur since only 1 major LSM can be activated at a time, but may be possible with external drives or after swapping disk from different machine.

The colors are whatever I picked on a whim, I'd prefer someone with actual aesthetic sense to fix those.

---
#### TODO

- [ x] Use `cargo fmt`
- [ x] Add necessary tests
- [ x] Add changelog entry
- [ x] Update default config/theme in README (if applicable)
- [ x] Update man page at lsd/doc/lsd.md (if applicable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for `--no-sort` `-U` from [MichaelAug](https://github.com/MichaelAug)
 - Add `--group-directories-first` as an alias for `--group-dirs=first` to improve compatibility with `coreutils/ls`
 - Add `--permission` flag to choose permission formatting (rwx, octal) from [meain](https://github.com/meain)
+- Display MAC and ACL indicators
 ### Fixed
 - Support non-bold bright colors [#248](https://github.com/Peltoche/lsd/issues/248) from [meain](https://github.com/meain)
 - Don't automatically dereference symlinks in tree/recursive [#637](https://github.com/Peltoche/lsd/issues/637) from [meain](https://github.com/meain)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add support for `--no-sort` `-U` from [MichaelAug](https://github.com/MichaelAug)
 - Add `--group-directories-first` as an alias for `--group-dirs=first` to improve compatibility with `coreutils/ls`
 - Add `--permission` flag to choose permission formatting (rwx, octal) from [meain](https://github.com/meain)
-- Display MAC and ACL indicators
-- Display MAC labels
+- Display MAC contexts and MAC and ACL indicators from [mmatous](https://github.com/mmatous)
 ### Fixed
 - Support non-bold bright colors [#248](https://github.com/Peltoche/lsd/issues/248) from [meain](https://github.com/meain)
 - Don't automatically dereference symlinks in tree/recursive [#637](https://github.com/Peltoche/lsd/issues/637) from [meain](https://github.com/meain)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add `--group-directories-first` as an alias for `--group-dirs=first` to improve compatibility with `coreutils/ls`
 - Add `--permission` flag to choose permission formatting (rwx, octal) from [meain](https://github.com/meain)
 - Display MAC and ACL indicators
+- Display MAC labels
 ### Fixed
 - Support non-bold bright colors [#248](https://github.com/Peltoche/lsd/issues/248) from [meain](https://github.com/meain)
 - Don't automatically dereference symlinks in tree/recursive [#637](https://github.com/Peltoche/lsd/issues/637) from [meain](https://github.com/meain)
--- a/doc/lsd.md
+++ b/doc/lsd.md
@@ -128,8 +128,8 @@ lsd is a ls command with a lot of pretty colours and some other stuff to enrich 
 `-U`, `--no-sort`
 : Do not sort. List entries in directory order
 
-`-Z` `--label`
-: Display SELinux or SMACK security label
+`-Z` `--context`
+: Display SELinux or SMACK security context
 
 # ARGS
 
--- a/doc/lsd.md
+++ b/doc/lsd.md
@@ -128,6 +128,9 @@ lsd is a ls command with a lot of pretty colours and some other stuff to enrich 
 `-U`, `--no-sort`
 : Do not sort. List entries in directory order
 
+`-Z` `--label`
+: Display SELinux or SMACK security label
+
 # ARGS
 
 `<FILE>...`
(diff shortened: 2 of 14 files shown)


doc/lsd.md
```diff
@@ -125,6 +125,9 @@ lsd is a ls command with a lot of pretty colours and some other stuff to enrich
 `-U`, `--no-sort`
 : Do not sort. List entries in directory order
 
+`-Z` `--label`
```
meain, line 128: Let's use `--context` as gnu ls does that.

src/color.rs
```diff
@@ -340,6 +344,8 @@ mod elem {
                 exec: Color::Red,
                 exec_sticky: Color::Magenta,
                 no_access: Color::AnsiValue(245), // Grey
+                acl: Color::DarkCyan,
+                security_label: Color::White,
```
meain, line 348: white might break for people using light colorschemes. We could do cyan I guess. Suggestions welcome.

src/flags/blocks.rs
```diff
@@ -144,6 +150,23 @@ impl Blocks {
             self.prepend_inode()
         }
     }
+
+    /// Tnserts a [Block] of variant [INode](Block::Label), if `self` does not already contain a
+    /// [Block] of that variant. The positioning will be best-effort approximation of coreutils
+    /// ls position for a security label
+    fn optional_insert_label(&mut self) {
```
meain, line 157: We might have to join them together.

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - Add support for `--no-sort` `-U` from [MichaelAug](https://github.com/MichaelAug)
 - Add `--group-directories-first` as an alias for `--group-dirs=first` to improve compatibility with `coreutils/ls`
+- Display MAC and ACL indicators
+- Display MAC labels
```
meain, line 12:
```suggestion
- Display MAC and ACL indicators from [mmatous](https://github.com/mmatous)
```

meain (commented):
Just did a quick first pass. I will look more into it over the weekend. To be frank I am kinda new to selinux so I don't have a whole lot of clue as to what is going on here but let me take a look at it over the weekend.

src/flags/blocks.rs
mmatous: This happens if dir content doesn't fit on one line and `--long` isn't used. As far as I'm concerned this is expected behavior with grid layout having `TopToBottom` direction. It has same output as coreutils if changed to `LeftToRight`. Is there a reason for that? I tried changing it and everything worked as expected but maybe I'm not creative enough when listing directories.

What do you mean by joining?  Putting context in `block_vec` along with name when rendering blocks? That wouldn't work e.g. for `lsd -Z --blocks size` or any other context+field combination.

doc/lsd.md
mmatous: Can't believe I missed coreutils having a long option. Changed every occurence of label to context so code is consistent with CLI.

mmatous:
Fixed up everything except the grid thing. I'll push it as soon as we figure out what to do with it.

CHANGELOG.md
mmatous: To me + and . are indicators, whatever -Z prints is a context. Changing to 

> Display MAC contexts and MAC and ACL indicators from [mmatous](https://github.com/mmatous)

mmatous:
Pushed feedback fixes.
Don't know if you had time to look at this last weekend, but after some usage I'm pretty sure that changing the grid direction doesn't really affect anything other than the labels. The original code was from https://github.com/Peltoche/lsd/commit/6ca2bd948d5af61a2f7584eda1bec4c4d1923586 and I couldn't find any explanation for the `TopToBottom` and `LeftToRight` split.

src/color.rs
```diff
@@ -102,6 +104,8 @@ impl Elem {
             Elem::Exec => theme.permission.exec,
             Elem::ExecSticky => theme.permission.exec_sticky,
             Elem::NoAccess => theme.permission.no_access,
+            Elem::Acl => theme.permission.acl,
+            Elem::SecurityContext => theme.permission.security_context,
```
meain, line 108: What do you think about renaming it to just `context`?

meain:
Sorry about the delay, got a little preoccupied with a few other things.



This is the different between `LeftToRight` vs `TopToBottom`. In the first case the `Cargo.lock` is to the right of the first item an in the second case below. 

What happens in the original case is that when `?` and name are two different entries then name cuts down to the next row when we don't have space. I have not looked into it, but I'm assuming the grid lib has some way to force both the context and the name to be on the same line even when we have to display on multiple lines.

src/color.rs
mmatous: It thought that was a bit too abstract. I'm mostly used to context being whatever opaque structure user has to drag about to interact with C libraries like openssl or glib.
I'm not too hung up on it. I'll rename it if you want.

mmatous:
I think the simplest way would be adding `|| matches.is_present("context")` just like https://github.com/Peltoche/lsd/blob/f1254a6b0cb2583aa9c4d0cc3f849fdd3b62a488/src/flags/layout.rs#L32
I assume you encountered the exact same problem with inodes. The only other solution I could think of involved reorganizing the blocks vector and glueing context where it would have been.
The former is way less invasive and has the nice property of `lsd --blocks context,name` being the exact same as `lsd -Z`.
Is this OK with you?

meain:
I am thinking of maybe just combining them together [here](https://github.com/Peltoche/lsd/blob/f1254a6b0cb2583aa9c4d0cc3f849fdd3b62a488/src/display.rs#L101) and passing it in as a single cell, but I don't mind leaving that for the future. Feel free to just make it switch to `Layout::Oneline` like inode.

src/color.rs
meain: Since this is under `theme.permission` I was thinking that `context` alone would be clear enough. Plus, the flag is just `--context`. The main idea behind this was to make the name in theme file consistent with `--context` flag.

mmatous:
Ended up going with `OneLine`. Solutions involving blocks at that point are too late since they're strings and there's no truly foolproof way to tell context and other blocks that may just look like context apart.
Saving context in https://github.com/Peltoche/lsd/blob/f1254a6b0cb2583aa9c4d0cc3f849fdd3b62a488/src/display.rs#L250 and then just glueing it onto the closest block in the next cycle works but... meh. There're edge cases and it ends up being ugly hacky branchy thing while `OneLine`'s one line is at least simple to reason about.

meain:
What I was thinking about was to merge all the blocks at https://github.com/Peltoche/lsd/blob/f1254a6b0cb2583aa9c4d0cc3f849fdd3b62a488/src/display.rs#L101-L108 if we don't have `--long` or `OneLine` enabled and treat all the blocks combined for a single meta to be a single entity. I haven't considered all the edge cases, but I feel like that would work. In any case, we can merge this in with `OneLine` for now.

mmatous:
I rebased the PR and forgot to comment. Sorry if you missed it.

meain:
Thanks for working on this :D

PR #641, opened 2022-03-23
impr: Add short `-T` arg for `--tree`

5HT2:
I feel that this is a reasonable request, given `-T` is not used by anything else right now.

I don't feel that integration tests for the short option are necessary when the long option works, and I'm not sure how to make a changelog entry. 

If it is necessary, I would be more than happy to fix it.

---
#### TODO

- [x] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry

### DIFF
--- a/doc/lsd.md
+++ b/doc/lsd.md
@@ -80,7 +80,7 @@ lsd is a ls command with a lot of pretty colours and some other stuff to enrich 
 `--total-size`
 : Display the total size of directories
 
-`--tree`
+`-T` `--tree`
 : Recurse into directories and present the result as a tree
 
 `-V`, `--version`
--- a/src/app.rs
+++ b/src/app.rs
@@ -103,6 +103,7 @@ pub fn build() -> App<'static, 'static> {
         )
         .arg(
             Arg::with_name("tree")
+                .short("T")
                 .long("tree")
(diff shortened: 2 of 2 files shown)


zwpaper:
Hi @5HT2, thanks for contributing, but the reason we are not using `-T` is that it is used by the `GNU ls` which is aligning to

```
-T, --tabsize=COLS
              assume tab stops at each COLS instead of 8
```

as for the questions,
1. Integrating tests is for the CI to make sure everything works as expected in future developments, a short flag checking is also necessary not only for the functionality.
2. changelog entry is updating here https://github.com/Peltoche/lsd/blob/master/CHANGELOG.md

5HT2:
Oh, oops. I did not realize it was used by GNU `ls`. Is `-E` taken up by something?

Thank you for answering the other two questions, that's more straightforward than I expected.

meain:
@5HT2 I don't think `-E` is a good flag. `-T` would have been great but since that is not available and this is not something that might be used a lot, I think it would be better to skip the short flag altogether. That said, if you do use this often, I would suggest creating an alias like `alias tree='lsd --tree'`.

5HT2:
That makes sense, thank you.

PR #646, opened 2022-04-09
644 Do not calculate total directory size unless it will be displayed

juansc:
Addresses https://github.com/Peltoche/lsd/issues/644. 

Any suggestions for how to implement a test for this?

---
#### TODO

- [x] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry 


zwpaper:
how about adding a test for `displays_size` like https://github.com/Peltoche/lsd/pull/646/files#diff-8992e6e5f254bdfb40335a8538fbcf0f0d77baba1b2f09b4358ab8c69aa344beL264

and add an integration test to check the functionality like https://github.com/Peltoche/lsd/blob/master/tests/integration.rs#L387

juansc:
Hi @zwpaper! Looking at the integration tests it looks like they all run the command and verify the output. The change here is that we won't run a calculation that is currently not being displayed. This change can't be tested or detected by looking at the output of the command. 

Maybe I can do the following? 
``` rust
impl Flags {
  fn should_calculate_total_size(&self) -> bool {
    self.total_size.0 && self.blocks.displays_size()
  }
}
``` 
Then maybe I can test the value of `should_calculate_total_size` given different configurations. My only concern is that the decision is not really a concern for `Flags`  but rather `Core`

PR #651, opened 2022-04-30
Add icons for HEIC, PEM and TOML

nix6839:
Add icons for HEIC, PEM and TOML
---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry


meain:
Thanks for the PR.

PR #654, opened 2022-05-03
622 block headers

### LINKED ISSUE #622: Block Headers
Is there a way to have block headers?
image

MichaelAug:
closes #622 
---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry


MichaelAug:
This is how the header looks like
image

meain:
On second thought should we just use underlines instead of using `-`? It looks like we have a lot of wasted space.

MichaelAug:
> On second thought should we just use underlines instead of using `-`? It looks like we have a lot of wasted space.

Yes, it does look like there's a lot of wasted space. Will update

MichaelAug:
Underlined text looks much better imo :)
image

zwpaper:
Thanks for contributing!

PR #661, opened 2022-05-17
windows: restore colors when lsd is run under msys2, fixes #657

### LINKED ISSUE #657: Color output is broken in some combinations on Windows + MSYS2
- os: windows
- `lsd --version`:  0.21.0
- `echo $TERM`:  xterm-256color
- `echo $LS_COLORS`:  <unset>

## Expected behavior
Seeing colors.

## Actual behavior

With windows terminal, launcing bash using msys2:
image

wezterm, launching zsh using msys2:
image

See colored prompt and colored cargo output is working. Only "lsd" command output has codes showing `←[38;5;4m←[1m  bin←[0m  ←[38;5;4m←[1m  pkg←[0m`.

The same "lsd.exe" correctly shows colors in powershell and cmd.exe, and also when run thru stand-alone msys2.

martinlindhe:


martinlindhe:
I am not sure this is the best solution as it adds a dependency for ansi_term. This should ideally work with crossterm, but crossterm is doing something strange today.

martinlindhe:
This will be correctly resolved when crossterm merges https://github.com/crossterm-rs/crossterm/pull/657 and when lsd updates to a future version.

PR #669, opened 2022-06-12
Refactor icon module

TeamTamoad:
This PR do 2 things

1. Construct icon hash maps from an array to prevent the reallocation (we know the exact list of key and value)
2. Simplify the code about icon matching logic
    - The matching order is the same
    - Use match statement to make code more readable (It also requires more indentation)
    - Use cfg attribute instead of cfg macro

#### TODO

- [x] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry


meain (approved):
thanks!

PR #676, opened 2022-06-17
refactor(name): reduce unnecessary String allocation

TeamTamoad:
This PR improve performance of `escape` and `hyperlink` functions of `Name` struct by reducing number of `String` allocation.

---
#### TODO

- [x] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry

### DIFF
--- a/src/meta/name.rs
+++ b/src/meta/name.rs
@@ -4,6 +4,7 @@ use crate::icon::Icons;
 use crate::meta::filetype::FileType;
 use crate::print_error;
 use crate::url::Url;
+use std::borrow::Cow;
 use std::cmp::{Ordering, PartialOrd};
 use std::ffi::OsStr;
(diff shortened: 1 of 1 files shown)


meain:
Is it really making much of a difference? I was getting mostly mixed results with original version being better(within error margin).

TeamTamoad:
Hmm. Maybe this changes doesn't significantly improve performance of the program. However, I'll try writing a benchmark to test performance `render`, `escape` and `hyperlink` functions of `Name` struct.
What do you think about `Cow<str>`? Do you have any concern about its complexity?

meain:
> Do you have any concern about its complexity?

As long as it provides some perf improvement, I don't mind the added complexity.

meain:
I'm closing this PR for now. We can revisit it later if you think this could improve things here.

PR #698, opened 2022-07-10
docs(readme): make example config file collapsible

TeamTamoad:
Because the example config files is quite long, it's annoying to scroll all the way down to the bottom.
This PR makes example config file collapsible. However, it still shows the config file by default.

---
#### TODO

- [ ] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,9 +81,9 @@ You can also provide a configuration file from a non standard location:
 
 ### Config file content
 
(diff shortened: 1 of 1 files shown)


meain (approved):
I guess since it is open by default, there is no harm in adding this anyways.

PR #700, opened 2022-07-10
docs(readme): add repology badge

tranzystorek-io:
Adds repology badge for a quick overview of different available packages of `lsd`

---
#### TODO

- [ ] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,13 @@ The project is heavily inspired by the super [colorls](https://github.com/athity
 
 ## Installation
 
+<details>
+<summary>Packaging status</summary>
+<a href="https://repology.org/project/lsd/versions">
+    <img src="https://repology.org/badge/vertical-allrepos/lsd.svg?columns=3" alt="Packaging status">
+</a>
+</details>
+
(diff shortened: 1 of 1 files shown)


meain:
This looks like a good addition,but having this gigantic thing right at the starts feel like a bit too much. What is your opinion on moving it to the very end and maybe linking to a heading from the installation section?

tranzystorek-io:
I added the parameter to the link so that it's split into three columns, shouldn't be too big:

obraz

But of that's still too sizable then yeah, sure linking seems fine

meain:
I did see that 3 column layout, but I feel like it is still too big. Something like 5 would be good, but then it becomes unusable on a mobile. I think we will just link to it.

tranzystorek-io:
> I guess, we can have it as a collapsed section here as well using `<details>` thingy.

Switched to the html version of the badge wrapped in details, IMO looks fine now 😄

meain:
I guess that ain't that bad.

Issue #706, opened 2022-07-18
Adding new file types


bearcatsandor:
Is there a process for users to add file types that aren't recognised.  For instance, I work with  wavpack and dsd files and it would be great to have them recognised as audio file types.

meain:
If those are common enough extensions, feel free to just add those into `lsd`. You can do that by adding them in https://github.com/Peltoche/lsd/blob/946f2430661e57e0eda641a3f70601d4b46a3b71/src/icon.rs#L225

meain:
We also have #707 in the pipeline for enabling custom icons.

dreamerlzl:
Sorry for asking in a closed channel, but is "justfile" legit to be a common extension?

meain:
Sure, feel free to add an icon if that is what you are going for.

Issue #655, opened 2022-05-04
cannot modify columns colors


greweln:
- os: Fedora 35/36
- `lsd --version`:0.20.1 
- `echo $TERM`: Alacritty
- `echo $LS_COLORS`: :di=001:ex=0;32:

The colors of the columns won't change despite any changes on the default theme.
Any sugestions?

meain:
There is a possibility that you might have some issues in your config. Could you post your configs here.

greweln:
config.yaml
`
# == Classic ==
# This is a shorthand to override some of the options to be backwards compatible
# with `ls`. It affects the "color"->"when", "sorting"->"dir-grouping", "date"
# and "icons"->"when" options.
# Possible values: false, true
classic: false

# == Blocks ==
# This specifies the columns and their order when using the long and the tree
# layout.
# Possible values: permission, user, group, size, size_value, date, name, inode
blocks:
  - permission
  - user
  - group
  - size
  - date
  - name

# == Color ==
# This has various color options. (Will be expanded in the future.)
color:
  # When to colorize the output.
  # When "classic" is set, this is set to "never".
  # Possible values: never, auto, always
  when: auto
  # How to colorize the output.
  # When "classic" is set, this is set to "no-color".
  # Possible values: default, <theme-file-name>
  # when specifying <theme-file-name>, lsd will look up theme file
  # XDG Base Directory if relative, e.g. ~/.config/lsd/themes/<theme-file-name>.yaml,
  # The file path if absolute
  theme: default

# == Date ==
# This specifies the date format for the date column. The freeform format
# accepts an strftime like string.
# When "classic" is set, this is set to "date".
# Possible values: date, relative, '+<date_format>'
# `date_format` will be a `strftime` formatted value. e.g. `date: '+%d %b %y %X'` will give you a date like this: 17 Jun 21 20:14:55
date: date

# == Dereference ==
# Whether to dereference symbolic links.
# Possible values: false, true
dereference: false

# == Display ==
# What items to display. Do not specify this for the default behavior.
# Possible values: all, almost-all, directory-only
# display: all

# == Icons ==
icons:
  # When to use icons.
  # When "classic" is set, this is set to "never".
  # Possible values: always, auto, never
  when: auto
  # Which icon theme to use.
  # Possible values: fancy, unicode
  theme: fancy
  # Separator between icon and the name
  # Default to 1 space
  separator: " "

# == Ignore Globs ==
# A list of globs to ignore when listing.
[...]

Issue #566, opened 2021-10-14
lsd not reading the config file


greweln:
Please excuse my ignorance but how can I get lsd to read the config & theme file?
I install it with cargo and I do have a ~/.config/lsd/config.yml  and  ~/.config/lsd/themes/my_theme

thanks

meain:
1) You have to build from `master`. Themes is not yet in a release.
2) Have this is `~/.config/lsd/config.yaml`
```
color:
  when: auto
  theme: mytheme
```
3) Have theme in `~/.config/lsd/themes/mytheme.yaml` (`.yaml` at the end)

Issue #343, opened 2020-04-07
[Feature Suggestion] Custom icons for files and folders


j-maynard:
I can think of many ways to implement this but simply put I'd like to have a mechanism for assigning custom icons to files and directories and overriding those already defined within LSD.  I'd also like to be able to configure LSD to match its directory icon assignment and this feature suggestion would allow me to do that.  See the example below to see the difference between colorls and LSD:

![image](https://user-images.githubusercontent.com/29251905/78699791-5a484980-78fc-11ea-82ed-853c76def664.png)

As for implementation the two ideas I have are to either have an LSD config somewhere where these options can be defined along with the request to configure colors which has been suggested in the past.  Alternatively would be to follow my suggestion for `.hidden` files.  Define a file format in say a `.icon` or `.lsd` file which specifices either an icon or hex which people can get from [https://www.nerdfonts.com/cheat-sheet](https://www.nerdfonts.com/cheat-sheet) followed by the name of the file they want to change.  The best maybe to support both options.

KSXGitHub:
I prefer this feature work out-of-the-box. If a it has to be configurable, I would prefer global configuration (e.g. ~/.config/lsd/something.json).

j-maynard:
I prefer global configuration as well generally not least of all I can throw it in to my dot-files repo and have the same experience wherever I work.

neezer:
I'd also love this feature, but IMO it should support wildcards for the filename. As an example, I have a folder with a bunch of Dockerfiles in it that all have various extensions, eg.

```
Dockerfile.eslint
Dockerfile.prettier
Dockerfile.typescript-language-server
...
```

In vim, I have one `autocmd` to correctly set the filetype for all of them:

```
autocmd BufRead,BufNewFile Dockerfile.* set filetype=dockerfile
```

Would be great if I could do the same for `lsd` so they all have the Docker icon without requiring me to name them all explicitly.

---

> The best maybe to support both options.

👍 on this suggestion, with the understanding that local `.hidden` files supersede the top-level configuration. 

As an example: this would help in JS projects with extension-less multi-format configuration files, eg. `.eslintrc` or `.prettierrc`. These can be configured with JSON or YAML (or in some cases TOML), so I could set a top-level default icon for `lsd` to always treat them as whatever I think they should be, then override on a per-project basis.

zwpaper:
[...]

PR #707, opened 2022-07-26
:sparkles: theme: bring icon theme to lsd

zwpaper:
refactor color to be under theme so that we can add icon theme

**IMPORTANT**

this breaks the theme config by moving the origin one to under `color:`

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry


meain:
@zwpaper I think we should probably hash out the details for a breaking change before you put in effort to implementing it. Can we start a thread on this, or continue here. Let's finalize about what exactly the change would be, why it is required and any alternate approaches so as to avoid a breaking changes.

zwpaper:
related:
- https://github.com/Peltoche/lsd/issues/343

maybe related:
- https://github.com/Peltoche/lsd/issues/706

zwpaper:
@meain this is why I created this draft PR, let's discuss it here directly.

the changes I would like to introduce are that move the current contents of the theme file to become a sub-item of `color`.
the reason is that I want to use one theme file for both color and icons.

we now have color options in the root of the theme file(introduced by me, what a bad choice🤣) for example:

```yaml
user: 230
group: 187
permission:
  read: dark_green
```

I would like to make it:

```yaml
color:
  user: 230
  group: 187
  permission:
    read: dark_green
icon:
  suffix:
    rs: ABCD
    md: 0123
```

zwpaper:
there may be an option to support both 

```yaml
user: 230
group: 187
permission:
  read: dark_green
```

and 

```yaml
color:
  user: 230
  group: 187
  permission:
    read: dark_green
```

in the same time, but I think it is not worthy to keep this compatibility code as we can easily notify user and easily update the theme file.

meain:
I don't think there is anything necessarily bad about having multiple files, one for colors and one for icons. Besides, I see a few positives for this:
- No breaking changes
- Theme files will not get big
- Easier to copy/paste theme(of icon set) from wiki(once we have it)

zwpaper:
fine, I did not have a strong opinion on the unique theme file, let's leave it split.

zwpaper:
@meain, how about dropping this theme option?
image

as we discussed in the other issue, the Unicode option is not used in most cases, and the `when` option should be enough for us.

if a user want to use an Unicode theme, they can create a custom one by this PR's feature.

meain:
How about for icons theme, we explicitly call it `overrides` and keep the `theme` one with existing functionality. I am not sure if that is the best approach though.

_For folks without patched fonts, `unicode` is an option to get some icons without much hurdle._

---

One other option that I strongly prefer is something different. We don't let users specify a file, instead if a file called `colors.yaml` is preset in `$XDG_CONFIG_HOME/lsd`, we load colors from there and if a file called `icons.yaml` is present, we load icons from there. It feels much simpler. We can support specifying themes in both ways for color for some time with a deprecation message on the old way while people switch. The current way of specifying a file path seems a bit clunky overall.

zwpaper:
for the Unicode part, we currently have ONLY 2 icons used, one for folders, and one for files, even more, these two Unicode icons can NOT work in MOST of the systems including the widely used macOS.
even that, There is only one issue we received to discuss.

zwpaper:
for the default theme file part, I am totally ok with the `color.yaml` and `icon.yaml`, the default theme we offered works like those two files, the difference is users can not update the default theme.

BUT offering the ability to specify a file(or path) is a necessary one, as users can save many themes and switch in need.
for example, my terminal background color may change between day and night, then I would like to use a different color theme for lsd.

zwpaper:
the default theme files could be another enhancement, I have created one to discuss https://github.com/Peltoche/lsd/issues/709

meain:
> for the default theme file part, I am totally ok with the color.yaml and icon.yaml, the default theme we offered works like those two files, the difference is users can not update the default theme.

The idea is that these files will be how they customize it, not the default theme. I don't think adding the default theme values as files in would be all that useful.

> BUT offering the ability to specify a file(or path) is a necessary one, as users can save many themes and switch in need.
for example, my terminal background color may change between day and night, then I would like to use a different color theme for lsd.

For that usecase, having the file with a consistent name is a much better approach than changing the path in `config.yaml`. For example if I can have both `colors.yaml` and `icons.yaml` as symlinks to actual themes and just change the symlink paths in a script, or even just copy the original file to the default locations. The alternate way when we have to specify paths in `config.yaml` would be for them to use some hacky `sed` script to to use something like `yq` to change up the value which would be much harder.

zwpaper:
> I don't think adding the default theme values as files in would be all that useful.
I am saying that the file works like users' default theme, but putting the real default theme in it. we have no divergence in this part.

> For that usecase, having the file with a consistent name is a much better approach than changing the path in config.yaml.

I don't really think that a symbol link or copying files are better than changing the `config.yaml`.
as we have the `config.yaml`, it is meant to be the config center, I don't see any problems updating the config.yaml when users want to change the config.
a symbol link is more like a workaround if some apps did not have a config file.

so that my point is we can keep both of the features, we reserved the `color.yaml` and `icon.yaml` to be the default config, and use the custom one if use specified in the `config.yaml`.
so that

1. did not have any break changes.
2. we satisfied both of the people like you and me.

zwpaper:
One more question appeared, should we use the char directly or Hex in config file?

specifying chars is supported now. we have to do something more if we want to support Hex format.

meain:
> a symbol link is more like a workaround if some apps did not have a config file.

It does not really have to be symlink, it could even be the user replacing the file with a new one there. Here is an example situation. Say the user has a script which runs in the background and change the theme everywhere in the evening. If we have separate files, they can just replace the file. But if the config is just a path, they will have to find some way to edit content within a yaml file which I believe would be harder and might require external tools like [yq](https://github.com/mikefarah/yq) for example.

> so that my point is we can keep both of the features, we reserved the color.yaml and icon.yaml to be the default config, and use the custom one if use specified in the config.yaml.

I was thinking about deprecating the old option as that has been confusing to people. Example: https://github.com/Peltoche/lsd/issues/655 , https://github.com/Peltoche/lsd/issues/566

zwpaper:
I still believe that specifying the theme in config file is the proper config file usage.
no matter symlink or moving files, both seems like the workaround.

as I mentioned above, we can offer the default file, but I don't think we should deprecate the previous one.
which also will not make any break changes, I believe that I treat the breaking important more than me.

meain:
> I still believe that specifying the theme in config file is the proper config file usage.
no matter symlink or moving files, both seems like the workaround.

I feel like there is some communication gap here. You can think of `colors.yaml` and `icons.yaml` as just two other config files instead of having it be something that you have to link from the "main" config file. For example you can check config files for calcurse(has multiple files like keys, conf), git (has files like attributes, config), mpv (input.conf and mpv.conf), ncmpcpp (bindings, config) etc. These don't link from one another.

> as I mentioned above, we can offer the default file, but I don't think we should deprecate the previous one.
which also will not make any break changes, I believe that I treat the breaking important more than me.

The reason why I thought we could deprecate the old way was because having multiple ways can possibly be confusing to folks. We have to decide which route we are taking before we go and implement it for icons.

zwpaper:
@meain I am still not convinced, but I dropped the custom file option and continued the PR as you insisted.

I check the default icon theme file if no icon theme flag is present, but not handling the color theme file, one reason is that the color theme file has gone public with the previous release, and another is that this PR should limit to icons.

meain:
> I check the default icon theme file if no icon theme flag is present, but not handling the color theme file, one reason is that the color theme file has gone public with the previous release, and another is that this PR should limit to icons.

Makes sense. We can do the color theme change in a different PR. The idea for color theme file is to support both for a while with only the newer one advertised and depreciate the old option after some time and remove.

zwpaper:
@meain all works should be ready now. I will leave the prefix/regex to other issues.

meain:
Hey @zwpaper , sorry about the delay. I have been pretty busy the past few weeks plus wanted to make a release to help get a deb package if possible. I'll get to this soon.

zwpaper:
@meain never mind, it is absolutely ok, look into this any time when you are free

README.md
```diff
@@ -296,6 +298,71 @@ and then change its colors, the items missed would fallback to use the default c
...
+- icons-by-extension
+
+The default icon theme scheme shipped with `lsd` can be check in [icon theme source code](src/theme/icon.rs), we will load the default theme, and overwrite it with user defined parts, here is a example for icon theme.
+
+lsd icon theme support both nerd font and Unicode in the same time, you can use any one to config the theme, or even combine them.
+
+nerd font:
+
+```yaml
+icons-by-name:
```
meain, line 325: Since it will be a separate file, what do you think about just `name` here?

README.md
```diff
@@ -296,6 +298,71 @@ and then change its colors, the items missed would fallback to use the default c
 Please also notice that an empty theme is **NOT** supported due to
 [a bug in serde lib](https://github.com/dtolnay/serde-yaml/issues/86).
 
+### Icon Theme
+
+Icon theme can be configured in a fixed location, `LSD_CONFIG_DIR/themes/icons.yaml`,
+for example, `~/.config/lsd/themes/icons.yam` on macOS,
```
meain, line 304:
```suggestion
for example, `~/.config/lsd/themes/icons.yaml` on macOS,
```

README.md
```diff
@@ -296,6 +298,71 @@ and then change its colors, the items missed would fallback to use the default c
 Please also notice that an empty theme is **NOT** supported due to
 [a bug in serde lib](https://github.com/dtolnay/serde-yaml/issues/86).
 
+### Icon Theme
+
+Icon theme can be configured in a fixed location, `LSD_CONFIG_DIR/themes/icons.yaml`,
```
meain, line 303: Why not `LSD_CONFIG_DIR/icons.yaml`, do we need a subdirectory. The way I see it, we will have 3 top level files:
- `config.yaml`
- `colors.yaml`
- `icons.yaml`

> We will have to make changes to the color one as well, but yeah.

README.md
```diff
@@ -296,6 +298,71 @@ and then change its colors, the items missed would fallback to use the default c
...
+icons-by-filetype:
+  dir: 📂
+  file: 📄
+  pipe: 📩
+```
+
+When creating a theme for `lsd`, you can specify any part of the default theme,
+and then change its colors, the items missed would fallback to use the default colors.
+
+Please also notice that an empty theme is **NOT** supported due to
```
meain, line 363: I don't remember how the code looks like but should we check if the file is empty and only then pass it on to serde?

src/core.rs
```diff
@@ -47,11 +44,8 @@ impl Core {
             _ => flags.color.theme.clone(),
         };
 
-        let icon_theme = match (tty_available, flags.icons.when, flags.icons.theme) {
-            (_, IconOption::Never, _) | (false, IconOption::Auto, _) => icon::Theme::NoIcon,
-            (_, _, IconTheme::Fancy) => icon::Theme::Fancy,
-            (_, _, IconTheme::Unicode) => icon::Theme::Unicode,
-        };
+        let icon_when = flags.icons.when;
```
meain, line 47: Any reason to move this logic to inside `Icons::new`?

README.md
meain, line 303: I think we can mention `$XDG_CONFIG_DIR/lsd` here instead of `LSD_CONFIG_DIR` similar to how it is in colors.

src/display.rs
```diff
@@ -408,11 +408,11 @@ mod tests {
     use super::*;
     use crate::color;
     use crate::color::Colors;
-    use crate::flags::HyperlinkOption;
+    use crate::flags::{HyperlinkOption, IconOption, IconTheme as FlagTheme};
```
meain, line 411: Any reason for `as FlagTheme`?

meain (commented):
I did a quick pass, just a few things.


- Should we have an option to fallback to existing ones if none was found in the specified theme file? For example if the custom icons.yaml file only has icons for `.rs`, we can use icons for other filetypes from `lsd` builtins.
- Some tests are failing if we have a custom theme file in the users system, could you make it so that it load the default themes.

src/core.rs
zwpaper: the logic here is all about how to set icons, it should be handled inside the Icons package

zwpaper:
> Should we have an option to fallback to existing ones if none was found in the specified theme file? For example if the custom icons.yaml file only has icons for .rs, we can use icons for other filetypes from lsd builtins.

this is how icon theme works now

> Some tests are failing if we have a custom theme file in the users system, could you make it so that it load the default themes.

I have a both icon and color theme in my mac, and no error happened, could you point out a case so that I can test it?

zwpaper:
I have pushed one commit so that it is easy to review.

after your review is done, I will fix the conflict and force push again.

@meain

src/theme.rs
```diff
@@ -49,7 +52,7 @@ impl Theme {
             real
         } else {
             match config_file::Config::config_file_path() {
-                Some(p) => p.join("themes").join(real),
```
meain: We can probably support both `~/.config/lsd/themes/<name>.yaml` and `~/.config/lsd/colors.yaml` initially for a while. Dropping it without any warning might not be the best idea. Also, it will be `~/.config/lsd/colors.yaml` and not `~/.config/lsd/<name>.yaml` essentially making `config.color.theme`  unnecessary.

zwpaper: Oops, sorry for mistakenly changing the location, I have changed it back to the origin one and added a TODO sign, and a new issue https://github.com/Peltoche/lsd/issues/749 to track.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] - ReleaseDate
 ### Added
+- Add support for icon theme from https://github.com/Peltoche/lsd/pull/707 [kWeiZh](https://gitHub.com/zwpaper)
```
meain, line 9: `[kWeiZh](https://gitHub.com/zwpaper)` : Typo `kWeiZh` ?

We can probably also remove the PR number for consistency.

zwpaper: that's my other username, let me change it back to the origin one, drop the PR number and do the rebase work.

zwpaper: happened to find a typo, `gitHub` -> `github` https://github.com/Peltoche/lsd/blob/master/CHANGELOG.md?plain=1#L23

I will also change it for this minor change🤣

meain: Woops, that was my Emacs being helpful :D. Thanks for that.

zwpaper: not intelligent enough to also change the capital `G` for `GitHub`😁

zwpaper:
this now should be good to go after CICD if there are no other problems.

zwpaper:
@meain

README.md
```diff
@@ -244,12 +246,12 @@ The valid theme configurations are:
 when configured with the `theme-file-name` which is a `yaml` file,
 `lsd` will look up the theme file in the following way:
 
-- relative name: check the themes under XDG Base Directory, e.g. ~/.config/lsd/themes/<theme-file-name>.yaml
```
meain, line 247: Colors themes are still under `themes`, just the `icons.yml` is in the base dir.

src/color.rs
```diff
@@ -131,16 +128,24 @@ impl Elem {
...
 }
 
 impl Colors {
     pub fn new(t: ThemeOption) -> Self {
         let theme = match t {
             ThemeOption::NoColor => None,
-            ThemeOption::Default | ThemeOption::NoLscolors => Some(Theme::default()),
-            ThemeOption::Custom(ref file) => Some(Theme::from_path(file).unwrap_or_default()),
+            ThemeOption::Default | ThemeOption::NoLscolors => Some(Theme::default().color),
+            ThemeOption::Custom(ref file) => {
```
meain, line 140: I guess the only additional thing that we would need to support https://github.com/Peltoche/lsd/issues/749 would be to add an additional check here in case filename is not mentioned but we can take that up in another PR.

README.md
```diff
@@ -297,6 +299,68 @@ and then change its colors, the items missed would fallback to use the default c
...
+#### Icon Theme file content
+
+lsd support 3 kinds of icon configuration:
+- filetype
+- name
+- extension
+
+The default icon theme scheme shipped with `lsd` can be check in [icon theme source code](src/theme/icon.rs), we will load the default theme, and overwrite it with user defined parts, here is a example for icon theme.
+
+lsd icon theme support both nerd font and Unicode in the same time, you can use any one to config the theme, or even combine them.
```
meain, line 321: "combine them" might be a bit confusing, how about we say something like, "this is the default but you can customize this to use glyphs from nerd-font or unicode emojis" and give the example that you have there.

README.md
```diff
@@ -297,6 +299,68 @@ and then change its colors, the items missed would fallback to use the default c
...
+extension:
+  rs: 🦀
+filetype:
+  dir: 📂
+  file: 📄
+  pipe: 📩
+```
+
+When creating a theme for `lsd`, you can specify any part of the default theme,
+and then change its colors, the items missed would fallback to use the default colors.
```
meain, line 362: I guess you meant icons instead of colors here.

meain (commented):
Just some small things that I missed before.

src/color.rs
zwpaper: https://github.com/Peltoche/lsd/issues/749 updated

zwpaper:
thanks for the nice catch! @meain 

please take another look.

README.md
```diff
@@ -297,6 +299,96 @@ and then change its colors, the items missed would fallback to use the default c
...
+  executable: 
+  symlink-dir: 
+  symlink-file: 
+  device-char: 
+  device-block: ﰩ
+  special: 
+```
+
+When creating a theme for `lsd`, you can specify any part of the default theme,
+and then change its icons, the items missed would fallback to use the default icons.
```
meain, line 390: This doc block felt a bit obtuse and thought I would rework it. What do you think?

```suggestion
`lsd` support 3 kinds of icon overrides, by `name`, by `filetype` and by `extension`.
The final set of icons used will be a combination of what is shipped with in `lsd` with overrides from config applied on top of it.
*You can find the default set of icons [here](src/theme/icon.rs).*

Both nerd font glyphs and unicode emojis can be used for icons. You can find an example of icons customization below.

```yaml
name:
  .trash: 
  .cargo: 
  .emacs.d: 
  a.out: 
extension:
  go: 
  hs: 
  rs: 🦀
filetype:
  dir: 📂
  file: 📄
  pipe: 📩
  socket: 
  executable: 
  symlink-dir: 
  symlink-file: 
  device-char: 
  device-block: ﰩ
  special: 
```
```

Writing this actually gave me another idea, we could rename the `--icon-type unicode` to `--icon-type simple` and with this we will only use icons defined in `filetype` and not use anything by extension or name. Their configs will also affect in this case. As of now, it is kinda as if those don't do much. I'll add it to the issue, but just thought I would mention it here.

zwpaper:
@meain suggestion committed

meain (approved):
And now we have customisible icons. Thanks for this :D.

PR #708, opened 2022-07-26
refactor(flags): refactor parsing arguments

TeamTamoad:
- Make code more idiomatic and readable and have the same pattern
- Use `unreachable` to clearly show internal logic error
- Rename function `from_str` to `from_arg_str` to make it more specific

#### TODO

- [x] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -6,7 +6,7 @@ env:
   PROJECT_NAME: lsd
   PROJECT_DESC: "An ls command with a lot of pretty colors."
   PROJECT_AUTH: "Peltoche <peltoche@halium.fr>"
-  RUST_MIN_SRV: "1.56.0"
+  RUST_MIN_SRV: "1.62.1"
 
 on: [push, pull_request]
 
--- a/src/flags/blocks.rs
+++ b/src/flags/blocks.rs
@@ -71,24 +71,24 @@ impl Blocks {
     /// This errors if any of the parameter arguments causes [Block]'s implementation of
     /// [TryFrom::try_from] to return an [Err].
     fn from_arg_matches(matches: &ArgMatches) -> Option<Result<Self, Error>> {
-        if matches.occurrences_of("blocks") > 0 {
-            if let Some(values) = matches.values_of("blocks") {
-                let mut blocks: Vec<Block> = vec![];
-                for value in values {
-                    match Block::try_from(value) {
-                        Ok(block) => blocks.push(block),
-                        Err(message) => {
-                            return Some(Err(Error::with_description(
-                                &message,
-                                ErrorKind::ValueValidation,
-                            )))
-                        }
+        if matches.occurrences_of("blocks") == 0 {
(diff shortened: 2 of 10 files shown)


TeamTamoad:
I'm not sure if the custom message in `unreachable!` is good enough. Should I put the invalid value in the message?

meain:
I think probably adding something like `Invalid value {value} for <thing>` would be better for message in `unreachable` as they might make them more debuggable in case we miss something in clap. What do you think? _I did a [quick search on Sourcegraph](https://sourcegraph.com/search?q=context%3Aglobal+unreachable+lang%3ARust+&patternType=literal) but looks like most packages don't include much in there._

As for the change from `from_str` to `from_arg_string`, I guess the original idea was to mimic [FromStr](https://doc.rust-lang.org/std/str/trait.FromStr.html) but then again, we are not satisfying the proper constraints so I guess we can switch it.

TeamTamoad:
> I did a [quick search on Sourcegraph](https://sourcegraph.com/search?q=context%3Aglobal+unreachable+lang%3ARust+&patternType=literal) but looks like most packages don't include much in there.

That's a cool way to learn more about Rust. I've leaned something new 😎 

> I think probably adding something like Invalid value {value} for <thing> would be better for message in unreachable as they might make them more debuggable in case we miss something in clap. What do you think?

I think I can do that. There is no reason not to do it anyway. 😄

TeamTamoad:
Hmmm. I tested on my machine with latest Rust (1.62.1) and it's fine. The CICD failed because it use Rust 1.56. Should we update the MSRV or change the code ?

zwpaper:
hi @TeamTamoad, maybe the failing is about the clippy `cargo clippy -- -D warnings`, not the version, did not try clippy locally?

TeamTamoad:
I didn't get any warning on my machine.
image

I believed it's due to the fact that `unreachable!` is changed in Rust 1.59
https://github.com/rust-lang/rust/blob/master/RELEASES.md#compatibility-notes-3

meain:
> Hmmm. I tested on my machine with latest Rust (1.62.1) and it's fine. The CICD failed because it use Rust 1.56. Should we update the MSRV or change the code ?

If changing the CI to a newer version fixes it, then I think we could just bump it.

TeamTamoad:
I'm not familiar with GitHub workflow. Can I just push a commit changing `RUST_MIN_SRV` to 1.62.1 in this file?
https://github.com/Peltoche/lsd/blob/2bc55dd608b37418d8b7aca6d9e367771d88b59d/.github/workflows/CICD.yml#L1-L11

meain:
Yup, that should do it.

meain (approved):
LGTM. Thanks again for consistently helping improve the codebase. :D

TeamTamoad:
> LGTM. Thanks again for consistently helping improve the codebase. :D

You're welcome. You can call me Master of Refactoring.

PR #711, opened 2022-08-06
refactor: implement `Configurable` for `blocks`

TeamTamoad:
Previously, `blocks` struct try to mimic the `Configurable` trait by having functions with the same name and similar signature. It cannot implement `Configurable` because it propagate an `Error` to caller in case of invalid values.
However, those invalid values should be handled by `clap` in the first place. So, we can implement `Configurable` for `blocks` with custom `configure_from` instead of the default one.

---
#### TODO

- [x] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry

### DIFF
--- a/src/flags/blocks.rs
+++ b/src/flags/blocks.rs
@@ -51,7 +51,7 @@ impl Blocks {
         self.0.contains(&Block::Size)
     }
 
-    /// Tnserts a [Block] of variant [INode](Block::Context), if `self` does not already contain a
+    /// Inserts a [Block] of variant [INode](Block::Context), if `self` does not already contain a
     /// [Block] of that variant. The positioning will be best-effort approximation of coreutils
     /// ls position for a security context
     fn optional_insert_context(&mut self) {
--- a/src/flags/blocks.rs
+++ b/src/flags/blocks.rs
@@ -1,123 +1,19 @@
 //! This module defines the [Blocks] struct. To set it up from [ArgMatches], a [Config] and its
 //! [Default] value, use its [configure_from](Blocks::configure_from) method.
 
+use super::Configurable;
 use crate::config_file::Config;
 use crate::print_error;
 
 use std::convert::TryFrom;
 
-use clap::{ArgMatches, Error, ErrorKind};
+use clap::ArgMatches;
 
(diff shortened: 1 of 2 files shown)


src/flags/blocks.rs
```diff
@@ -1,18 +1,75 @@
...
+        if !self.contains_inode() {
+            self.prepend_inode()
+        }
+    }
+
+    pub fn displays_size(&self) -> bool {
+        self.0.contains(&Block::Size)
+    }
+
+    /// Tnserts a [Block] of variant [INode](Block::Context), if `self` does not already contain a
```
meain, line 54: Small typo: Tnserts → Inserts

src/flags/blocks.rs
```diff
@@ -23,72 +80,53 @@ impl Blocks {
...
-                    Err(message) => {
-                        return Some(Err(Error::with_description(
-                            &message,
-                            ErrorKind::ValueValidation,
-                        )))
-                    }
-                }
+                blocks.push(Block::try_from(value).unwrap_or_else(|_| {
+                    unreachable!(
+                        "Invalid value '{value}' for 'blocks' flag should be handled by `clap`"
```
meain, line 125: This is also present in the previous PR, but I was re-evaluating if we should mention clap?
How about just `Invalid value '{value}' for 'blocks'` ?

src/flags/blocks.rs
TeamTamoad: Oops I'll fix that.

src/flags/blocks.rs
TeamTamoad: I think we should make it clear why this statement should be unreachable because it is not obvious for new contributors. So, `clap` or `ArgMatchs` should be mentioned. 
I think it's fine to mention them in either the comment or in the panic message

meain: This message would be something that the users would see (if we mess something up). What I was thinking was that we could skip mentioning clap in the message but have it in a comment right next to it?

TeamTamoad: Yeah. That's good. I'll change this and make another PR to change others `unreachable`

meain (approved):

PR #714, opened 2022-08-07
refactor: use default attribute for enum

TeamTamoad:
Use the new feature from Rust 1.62 to make code more concise
https://blog.rust-lang.org/2022/06/30/Rust-1.62.0.html#default-enum-variants

Since we just bump the CICD to Rust 1.62.1, I think it's ok to use the new feature from Rust.

---
#### TODO

- [x] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -6,7 +6,7 @@ env:
   PROJECT_NAME: lsd
   PROJECT_DESC: "An ls command with a lot of pretty colors."
   PROJECT_AUTH: "Peltoche <peltoche@halium.fr>"
-  RUST_MIN_SRV: "1.62.1"
+  RUST_MIN_SRV: "1.62.0"
 
 on: [push, pull_request]
 
--- a/build.rs
+++ b/build.rs
@@ -18,11 +18,11 @@ use std::process::exit;
 include!("src/app.rs");
 
 fn main() {
-    match version_check::is_min_version("1.62.1") {
+    match version_check::is_min_version("1.62.0") {
         Some(true) => {}
         // rustc version too small or can't figure it out
         _ => {
-            writeln!(&mut io::stderr(), "'lsd' requires rustc >= 1.62.1").unwrap();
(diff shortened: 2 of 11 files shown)


meain:
That is a neat language addition. I was not aware of it. Could you bump up the min version in build script as well.
https://github.com/Peltoche/lsd/blob/1553aafe9db589dfe9d0385a2111127060af8db8/build.rs#L21

meain:
If `1.62.0` would do, lets keep it as that.

TeamTamoad:
> If `1.62.0` would do, lets keep it as that.

Could you clarify to me why it's better to use `1.62.0`?
I prefer `1.62.1` because
* It matches the Rust version in CICD.
* Some regression and vulnerability have been fixed in `1.62.1`. https://blog.rust-lang.org/2022/07/19/Rust-1.62.1.html

meain:
> Could you clarify to me why it's better to use 1.62.0?

The min version we support is not as important as it is not a lib but I would still like to let folks compile with the minimum version they can compile `lsd` with.

By setting this, we are in no way restricting us or users from using `1.62.1` and taking advantage of the new fixes but just mentioning that they won't be able to compile it with anything lower than `1.62.0`. Keeping it as low as possible (while updating it to meet our needs), we enable them to compile `lsd` even if they happen to not have the latest and greatest rust version(eg: their package manager might not have the latest version).

> It matches the Rust version in CICD.

Now that you mention it, I think we should switch it to `1.62.0` (if clippy is still happy) as well to make sure it compiles with that version as we change things.

TeamTamoad:
@meain Thank you so much for explaining it to me. I'll reduce the MSRV in both places.
I'm curious about when we release the `lsd`. What is the version of Rust that is used to compile code?

meain:
It will use the latest stable version IIRC. You can check the build step for the [CI run of previous release](https://github.com/Peltoche/lsd/runs/6848841574?check_suite_focus=true).

PR #722, opened 2022-08-13
refactor(test): remove unnecessary `vec!` macro

TeamTamoad:
---
#### TODO

- [x] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry

### DIFF
--- a/src/flags/permission.rs
+++ b/src/flags/permission.rs
@@ -97,8 +97,8 @@ mod test {
 
     #[test]
     fn test_from_arg_matches_short() {
-        let args = vec!["lsd", "--permission", "octal"];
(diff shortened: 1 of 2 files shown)


zwpaper:
Hi @TeamTamoad, maybe we should not change `args` to `argv`.

`args` seems to be more intuitive, `argv` seems to be a historically reserved choice of `c`

TeamTamoad:
> args seems to be more intuitive

I agree with you, but almost all code in the test use `argv`. I think it's should be the same for all tests.

meain:
As @zwpaper mentioned `args` is probably better here, but since this is just tests and we are already using `argv` pretty much everywhere else, it should be fine.

meain (approved):

Issue #574, opened 2021-11-01
Feature request: Color override setting for file/directory indicator column


adam-gaia:
Hello,

Currently the file/directory indicator column (the first character column with `--long` output) color is set to the color of the file type (which is picked up from LS_COLORS).

```
$ lsd --color=always
drwxr-xr-x agaia agaia 4.0 KB Fri Oct 29 09:27:51 2021 🗀 some_dir
.rw-r--r-- agaia agaia 1.9 KB Wed Oct 27 14:04:28 2021 🗀 some_file
^-----The colors of this first column match 'some_dir' and 'some_file's colors respectively
```

I would greatly appreciate a setting in the theme yaml to explicitly set the color of the file/directory indicator. 
I suggest an optional setting in the theme yaml. If this field is absent, fall back to the file type color from LS_COLORS.
```yaml
# excerpt from theme.yaml
directory-indicator: 15
```

My personal goal is to mimic gnu ls where only the file name is colored while retaining lsd's file icons. I would do so by setting the file/directory indicator to white. I already color the permissions, size, etc to white.

bryanwills:
Would like to have this feature. I cannot change the directory color and I can barely read the names they are in a horrible blue color. Please make this happen!

meain:
@bryanwills This issue is different. You can change the directory colors using LS_COLORS.

Reference: https://gist.github.com/thomd/7667642

PR #725, opened 2022-08-18
Feature: Color override setting for file/directory indicator column

themkat:
Attempt at fixing #574 . Provide two new theme options:
```yaml
file-indicator: 43
directory-indicator: 65
```
(numbers in example are arbitrary, and can be any color currently supported)

I thought one for each made more sense than a common one. Feel free to disagree.

The indicators use the theme in the file if found, if not they use the defaults for file/directory names. LS_COLORS color is only used if no theme setting is provided. This is based upon the following comment in the feature suggestion, which I found reasonable: 
> I suggest an optional setting in the theme yaml. If this field is absent, fall back to the file type color from LS_COLORS.  

I was unsure on how to test the LS_COLORS bits, so have briefly tested that manually. Please tell me if I'm missing something and there are good ways of writing tests for them 🙂 

In general I'm fairly new to Rust, so any feedback is greatly appreciated! 🙂 

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased] - ReleaseDate
 ### Added
 - Add icon for Zstandard from [nix6839](https://github.com/nix6839)
+- Add optional theme overrides for file indicator (`file-indicator`) and directory indicator (`directory-indicator`) in `--long` view from [themkat](https://github.com/themkat)
 ### Changed
 ### Fixed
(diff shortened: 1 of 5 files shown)


meain:
I think it would be better to keep just one color. This feature will be mostly if folks want to make everything consistent (you can see something similar in the original issue). When they want custom colors, I think it will be better to match `LS_COLORS`.

*PS: If we were to go the route of implementing separate colors we would need more than just file and dir colors, for example symlinks, block devices etc.*

PR #729, opened 2022-08-21
perf: reduces the binary size and improves the performance

sabify:
This PR reduces the binary size by ~25% and improves the performance slightly by ~2% on M1 Mac.

---
#### TODO

- [x] Use `cargo fmt`
- [ ] Add necessary tests
- [x] Add changelog entry


meain (approved):
I think I remember doing this before but deciding not to for some reason, but yeah I can't remember why and I can think of any reason why we shouldn't be doing this for release. I'm approving the PR, will merge as soon as the CI run is complete.

meain:
Thanks for the PR. :D

PR #748, opened 2022-10-02
fix: do not quote filename when piping to another program

TeamTamoad:
According to GNU `ls` behavior, when piping the output to another program, the filename with special characters should not be quoted.

**Old**
image

**New**
image

**Reference**
image

## Note
- I put `should_quote` field in `Flag` struct because I think users should be able to configure this behavior. I'll submit another PR  implementing that when I have time. What do you think?
- I'm investigating the behavior of GNU `ls` about quoting and I'm not confident that the current implementation is full compatible with GNU `ls`. So, I think we should not release a new version soon.
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [ ] Add changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 - In keeping with the coreutils change, add quotes and escapes for necessary filenames from [merelymyself](https://github.com/merelymyself)
 
+### Fixed
+- Do not quote filename when piping into another program from [TeamTamoad](https://github.com/TeamTamoad)
+
 ## [0.23.1] - 2022-09-13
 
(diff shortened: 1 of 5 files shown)


merelymyself:
Probably relevant: #660

meain:
> I put should_quote field in Flag struct because I think users should be able to configure this behavior. I'll submit another PR implementing that when I have time. What do you think?

Not sure if being able to decide if we quote would be useful. But, if we are decide on doing it, it could be `[auto, always, never]` like other options.

> I'm investigating the behavior of GNU ls about quoting and I'm not confident that the current implementation is full compatible with GNU ls.

I had done some testing when the original PR was added, but do let me know if you find anything else.

*PS: Add a changelog entry and I think we should be good with this PR.*

Issue #763, opened 2022-11-02
Theme colours for `inode` and `links` are not applied to the output


UrsaDK:
- os: macOS 12.6
- `lsd --version`: lsd 0.23.1
- `echo $TERM`: xterm-256color
- `echo $LS_COLORS`: fi=0:ex=1;32:su=1;4;32:sg=1;4;32:di=1;34:st=1;34:tw=1;4;34:ow=1;4;34:ln=1;36:or=1;4;36:mi=1;31:pi=35:bd=1;35:cd=1;35:so=1;35

<details>
<summary>~/.config/lsd/config.yaml</summary>

```
# This is a shorthand to override some of the options to be backwards compatible
# with `ls`. It affects the "color"->"when", "sorting"->"dir-grouping", "date"
# and "icons"->"when" options.
# Possible values: false, true
classic: false

# == Blocks ==
# This specifies the columns and their order when using the long and the tree
# layout.
# Possible values: permission, user, group, size, size_value, date, name, inode
blocks:
  - permission
  - links
  - user
  - group
  - size
  - date
  - name

# == Color ==
# This has various color options. (Will be expanded in the future.)
color:
  # When to colorize the output.
  # When "classic" is set, this is set to "never".
  # Possible values: never, auto, always
  when: auto
  # How to colorize the output.
  # When "classic" is set, this is set to "no-color".
  # Possible values: default, <theme-file-name>
  # when specifying <theme-file-name>, lsd will look up theme file
  # XDG Base Directory if relative, e.g. ~/.config/lsd/themes/<theme-file-name>.yaml,
  # The file path if absolute
  theme: light

# == Date ==
# This specifies the date format for the date column. The freeform format
# accepts a strftime like string.
# When "classic" is set, this is set to "date".
# Possible values: date, relative, '+<date_format>'
# `date_format` will be a `strftime` formatted value. e.g. `date: '+%d %b %y %X'` will give you a date like this: 17 Jun 21 20:14:55
date: '+%e %b %H:%M'

# == Dereference ==
# Whether to dereference symbolic links.
# Possible values: false, true
dereference: false

# == Display ==
# What items to display. Do not specify this for the default behavior.
# Possible values: all, almost-all, directory-only
# display: all

# == Icons ==
icons:
  # When to use icons.
  # When "classic" is set, this is set to "never".
  # Possible values: always, auto, never
  when: never
  # Which icon theme to use.
  # Possible values: fancy, unicode
  theme: fancy
[...]

PR #768, opened 2022-11-13
Theme colors for inode and links are now being applied to the output

d-sonuga:
This is a fix for #763

### DIFF
--- a/src/color.rs
+++ b/src/color.rs
@@ -116,8 +116,8 @@ impl Elem {
             Elem::FileLarge => theme.size.large,
             Elem::FileMedium => theme.size.medium,
             Elem::FileSmall => theme.size.small,
-            Elem::INode { valid: false } => theme.inode.valid,
-            Elem::INode { valid: true } => theme.inode.invalid,
+            Elem::INode { valid: true } => theme.inode.valid,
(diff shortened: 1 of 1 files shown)


meain:
Thanks for the PR! :D

brainsik:
Looking at this code, it only fixes inode colorings. Coloring links according to theme is still broken. Trying the latest from master confirms that's the case:

```sh
$ cargo install --list
lsd v0.23.1 (https://github.com/Peltoche/lsd.git?branch=master#ea56a7c3):
    lsd
$ unset LS_COLORS # to be sure this isn't causing problems
```

image

cochon-git:
> Coloring links according to theme still doesn't work.

Only just arrived here as a recent convert to lsd, searching for this exact colour problem. However there seems to be confusion here over terminology and expectations with respect to the term 'link'.

The original issue, my wish and indeed this fix was to control the colour of the inode and [hard] link count columns displayed by `lsd -1 --blocks links,inode` not to highlight valid/hanging symlinks which can at least be controlled by LS_COLORS.

The terminology in colors.yaml certainly seems to imply symlink validity however, which is confusing, as the 'invalid:' setting seems irrelevant to the links column, and would be a fringe abnormality for the inode column too.

PR #769, opened 2022-11-15
Fix showing every file as executable on Windows

### LINKED ISSUE #712: Stop lsd on windows from showing every regular file as green
- os: Windows 11
- `lsd --version`: lsd 0.22.0
- `echo $TERM`: N/A
- `echo $LS_COLORS`: N/A

Windows NTFS doesnt have POSIX file permissions, so they need to be emulated.  This results in all files being marked as executable, which is necessary but annoying, since every regular file is shown in green.  Is there a way to make executable files white?

Icxolu:
This PR determines the executable state of a file on Windows using the file extension instead of permissions.
This fixes that every regular file was marked executable (green).

- ~~added `--exec-ext <ext>` to change mark extensions as executable on the fly~~
- ~~added `executable-extensions` config to modify them more permantly~~ `LS_COLORS` can be used to customize (other) extentions
- ~~cli flags and config file options will be merged~~
- `exe`, `msi`, `cmd` and `ps1` will be marked as executable on windows

Fixes #712 

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry

### DIFF
--- a/src/meta/permissions.rs
+++ b/src/meta/permissions.rs
@@ -127,7 +127,7 @@ impl Permissions {
         ColoredString::new(Colors::default_style(), res)
     }
 
-    #[cfg_attr(windows, allow(dead_code))]
+    #[cfg(not(windows))]
     pub fn is_executable(&self) -> bool {
         self.user_execute || self.group_execute || self.other_execute
     }
--- a/src/meta/permissions.rs
+++ b/src/meta/permissions.rs
@@ -127,6 +127,7 @@ impl Permissions {
         ColoredString::new(Colors::default_style(), res)
     }
 
+    #[cfg_attr(windows, allow(dead_code))]
     pub fn is_executable(&self) -> bool {
         self.user_execute || self.group_execute || self.other_execute
     }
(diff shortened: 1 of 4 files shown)


meain:
I like the idea of having predefined file extensions being marked as executable. I don't think we should have an option to configure it though. If they want to add more file extensions to be marked as green, one can always control it with LS_COLORS(or PR to upstream if it is really common). Could you rework it so that we don't have a new cli flag or config option?

Icxolu:
I dropped the configuration option and cli flag. I was not really happy with them either, it felt like a little too much, but i haven't thought of just using `LS_COLORS` as the customizable option. For now i marked `exe`, `msi`, `bat` and `ps1` as executable.

Icxolu:
With this in place, we could probably also revisit fetching ACL permission only on demand, since we don't use any other `Permission` info for coloring. This could potentially improve performance on (domain connected) Windows machines significantly. As a replacement, we could introduce a `Mode` block, like PowerShell does, which displays the file attributes.

src/meta/permissions.rs
```diff
@@ -127,6 +127,7 @@ impl Permissions {
         ColoredString::new(Colors::default_style(), res)
     }
 
+    #[cfg_attr(windows, allow(dead_code))]
```
meain, line 130: Nit, we can change this to `#[cfg(not(windows))]`.

meain (changes requested):
Looks good. Add a test for verifying this behavior on windows and we should be good to go. :D

src/meta/permissions.rs
Icxolu: done :innocent:

meain (approved):

PR #777, opened 2022-11-19
check isatty for underline in --header

### LINKED ISSUE #775: lsd -l --header doesn't filter ANSI sequences
I would mark this as a bug. But my issue template was not accepted. Couldn't figure out why, didn't think it matters much either.
This is a problem in the latest version, possibly all versions with --header.

## Current Behaviour

`lsd -l --header` doesn't filter ANSI sequences
`lsd -l --header | less` for example

## Expected Behaviour:

It should filter ANSI sequences.

GnikDroy:
Fixes #775 

The isatty check code is copied from core.rs

This PR was drafted very quickly with ripgrep and minimal reading. Hence, it might not be the ideal way to do things. Let me know if you want anything to be changed.

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -8,6 +8,11 @@ use term_grid::{Cell, Direction, Filling, Grid, GridOptions};
 use terminal_size::terminal_size;
 use unicode_width::UnicodeWidthStr;
 
+#[cfg(not(target_os = "windows"))]
+use std::io;
+#[cfg(not(target_os = "windows"))]
+use std::os::unix::io::AsRawFd;
+
 const EDGE: &str = "\u{251c}\u{2500}\u{2500}"; // "├──"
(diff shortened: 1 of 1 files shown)


meain:
I think it might be better to control it [here](https://github.com/meain/lsd/blob/f5ee0a2b5143b975c277a1c53870c08c6e211c37/src/core.rs#L62) by changing [`Header`](https://github.com/meain/lsd/blob/f5ee0a2b5143b975c277a1c53870c08c6e211c37/src/flags/header.rs#L12) to take more states just to keep it consistent. We should retain the external cli interface the same though.

GnikDroy:
Okay, I will take a look at this later in the day.

PR #790, opened 2023-01-13
Add installation information for debian sid and bookworm

alexanderkjall:
I packaged `lsd` for Debian, adding install information to the list. It will be part of the next Debian stable, and imported into Ubuntu also.

https://packages.debian.org/sid/lsd
https://packages.debian.org/bookworm/lsd


meain:
Thanks for the PR! :D

PR #792, opened 2023-01-14
Icon theme with overrides from config

### LINKED ISSUE #780: Icon theme with overrides from config
- os: debian
- `lsd --version`: lsd 0.23.1
- `echo $TERM`: Windows Terminal
- `echo $LS_COLORS`: 

## Current behavior
Created icons.yaml in ~/config/lsd folder based on example in documentation. Docs says
>"The final set of icons used will be a combination of what is shipped with in lsd with overrides from config applied on top of it".

However this does not work for me. Icons from what is shipped is gone anď only this in icons.yaml is used.

Using icon theme. Config.yaml and icons.yaml has a default icon.

image

Removing icon theme and now config.yaml and icons.yaml have the icon that is shipped.

image

sudame:
This PR will close #780 

## Background
:eyes: A detailed discussion can be found here: #780

When a user chooses to use custom icons, lsd should apply the custom icons for the conditions specified by the user, and use the default icons for all other conditions.
However, currently lsd ignores the default icons when a user sets custom icons.

## Cause
The struct `IconTheme` has HashMap fields named `name` and `extension`. The Serde (Rust's deserialization module) can merge default values for structs, but it cannot do so for HashMap. As a result, the `name` and `extension` fields will not be merged if a custom configuration exists.

## What I did
- Created the `ByFilename` struct which wraps a HashMap for the `name` and `extension` fields.
  - `ByFilename` has the IntoIterator trait to minimize the impact on other parts of the code.
- Set a custom deserializer for `ByFilename` .
- Added 4 new tests covering the situations of this PR.

## Comments
The number of added and removed lines in this PR is significant because I had to move the default icon settings.

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry
- [x] ~Update default config/theme in README (if applicable)~
- [x] ~Update man page at lsd/doc/lsd.md (if applicable)~


src/theme/icon.rs
```diff
@@ -1,13 +1,16 @@
...
+use std::hash::Hash;
 
 #[derive(Debug, Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "kebab-case")]
 #[serde(deny_unknown_fields)]
 #[serde(default)]
 pub struct IconTheme {
-    pub name: HashMap<String, String>,
-    pub extension: HashMap<String, String>,
+    #[serde(deserialize_with = "deserialize_by_name")]
```
meain, line 10: If we are using a custom deserialize, I think we can avoid creating separate structs.

sudame: I totally agree with you. I updated the PR. (5e81871)

zwpaper:
thanks so much, @sudame, it works like a charm

PR #820, opened 2023-03-09
Localization support

scarf005:
image

find locale set with [sys-locale](https://crates.io/crates/sys-locale), then format with locales.

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry


zwpaper:
thanks for your help! @scarf005

Issue #7, opened 2018-12-04
Create a git integration


Peltoche:

alichtman:
Can you explain what kind of git integration you're looking to add?

Peltoche:
Hello @alichtman,

There is no defined specs yet. This feature is already implemented by the colorls and the exa projects, I think it can be a good start to check how they do it.

ysf:
There is an example on the exa website: https://the.exa.website/features/git

Basically there is a new column that displays the git status for the files in the listed directory. (new, modified etc..)

Peltoche:
Seems cool. I think that we can do something even better with the icons support. I don't know if we should add a new icons for the git status or just override the existing one. We can also add a new column of course.

cjbassi:
Another tool to check out that displays the git status of files is [ranger](https://github.com/ranger/ranger). I like their use of icons, and I also like that they display the git status of entire repos too, so you could `ls` the current directory and see which repos are dirty. [Screenshot](https://ranger.github.io/screenshots/vcs.png).

leomikezee:
Maybe add a option or a flag to hide files based on the .gitignore file

patevs:
I definitely support this. Would be useful to be able to see the current git status of files and also being able to hide files based on the .gitignore file.

aseifert:
+1 from me! I'd really love to have this as well, it's the only thing I still miss from exa/colorls!

sherifabdlnaby:
+1!!!!

hpwxf:
I like this idea; maybe I can do it. Is there already someone on this issue ?

About the icons, if we want funny glyphs, my suggestions are (from https://www.nerdfonts.com/cheat-sheet)
        * check (-): nf-fa-check f00c
        * modified (M): nf-fa-gear f013
        * uknown (?): nf-fa-user_secret f21b 
        * ignored (!) : nf-mdi-ghost f79f (or nf-fa-exclamation f12a)
        * new (+) : nf-mdi-login_variant fafb 
<img width="688" alt="Capture d’écran 2021-02-28 à 13 11 49" src="https://user-images.githubusercontent.com/26146722/109418054-a4c82e00-79c6-11eb-848c-8f0ace7fea05.png">

The display should be with two columns (staged, not staged)

To get git info, I can use libgit2 (https://github.com/rust-lang/git2-rs)

Is there any other constraint ?

zwpaper:
[...]

PR #822, opened 2023-03-11
Git integration

hpwxf:
Add git integration following a behaviour very close to what `exa` do.

It adds a `--git` option and a block `git` which is only display when `--long` option is enable.

`git` is called only once for the current tree (if the current directory is a subdirectory of git, the status is displayed but if you use a deeper display, e.g. using `--tree`, the git status of repositories in subdirectories is not done). 

---
- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry


hpwxf:
this PR follows a discussion in https://github.com/Peltoche/lsd/issues/7#issuecomment-1453364120

hpwxf:
`git` is enabled by default (when available). The feature `no-git` may be used to force disable it.

The only unsupported target is `i686-pc-windows-gnu` (compilation failure during the link; ex: https://github.com/hpwxf/lsd/actions/runs/4271505990/jobs/7436019267)

hpwxf:
Updated with master

zwpaper (changes requested):
I have done some of the reviews, and allow me to leave the rest to a next time.

I have checked the simple part, it only has some nit requests to change, and I am trying to walk through the complicated parts, trying my best to get this merged

src/app.rs
```diff
@@ -96,21 +96,24 @@ pub struct Cli {
...
     #[arg(short = 'v', long)]
     pub versionsort: bool,
 
     /// Sort by TYPE instead of name
     #[arg(
-        long,
-        value_name = "TYPE",
-        value_parser = ["size", "time", "version", "extension", "none"],
-        overrides_with_all = ["timesort", "sizesort", "extensionsort", "versionsort", "no_sort"]
+    long,
```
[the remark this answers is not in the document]
hpwxf: I 100% agree. Done.

hpwxf:
Thx @zwpaper for the review. I'm sorry for the format issues; my conclusion: don't blindly trust code formatting

hpwxf:
@zwpaper is everything ok for you?

src/git.rs
```diff
@@ -0,0 +1,464 @@
...
+    pub fn get(&self, _filepath: &PathBuf, _is_directory: bool) -> Option<GitFileStatus> {
+        None
+    }
+}
+
+#[cfg(not(feature = "no-git"))]
+impl GitCache {
+    pub fn new(path: &Path) -> GitCache {
+        let _cachedir = std::fs::canonicalize(path).unwrap();
+        // log::info!("Trying to retrieve Git statuses for {:?}", _cachedir);
```
zwpaper, line 50: please drop those debug log comments

src/git.rs
```diff
@@ -0,0 +1,464 @@
...
+
+    pub fn get(&self, _filepath: &PathBuf, _is_directory: bool) -> Option<GitFileStatus> {
+        None
+    }
+}
+
+#[cfg(not(feature = "no-git"))]
+impl GitCache {
+    pub fn new(path: &Path) -> GitCache {
+        let _cachedir = std::fs::canonicalize(path).unwrap();
```
zwpaper, line 49: log error by https://github.com/lsd-rs/lsd/blob/8c62e9ed12ccf7a737bd7fc24ccfb0b36cf7b7da/src/main.rs#L66 and return empty instead of unwrap directly

src/git.rs
```diff
@@ -0,0 +1,464 @@
...
+        // log::info!("Trying to retrieve Git statuses for {:?}", _cachedir);
+
+        let repo = match git2::Repository::discover(path) {
+            Ok(r) => r,
+            Err(_e) => {
```
zwpaper, line 54: `Err(e)` and log by https://github.com/lsd-rs/lsd/blob/8c62e9ed12ccf7a737bd7fc24ccfb0b36cf7b7da/src/main.rs#L66

src/git.rs
```diff
@@ -0,0 +1,464 @@
...
+                        let str_path = status_entry.path().unwrap();
+                        let path: PathBuf =
+                            str_path.split('/').collect::<Vec<_>>().iter().collect();
+                        let path = workdir.join(path);
+                        let elem = (path, status_entry.status());
+                        // log::debug!("{:?}", elem);
+                        statuses.push(elem);
+                    }
+                }
+                Err(_e) => {
```
zwpaper, line 76: log error

src/meta/git_file_status.rs
```diff
@@ -0,0 +1,71 @@
...
+            },
+        }
+    }
+
+    pub fn render(
+        &self,
+        colors: &crate::color::Colors,
+        git_symbols: &crate::git_symbol::GitSymbols,
+    ) -> crate::color::ColoredString {
+        use crate::color::{ColoredString, Colors};
```
zwpaper, line 49: move `use` and `crate` to the top of the file

src/theme/git_symbol.rs
```diff
@@ -0,0 +1,52 @@
...
+    pub renamed: String,
+    pub ignored: String,
+    pub typechange: String,
+    pub conflicted: String,
+}
+
+impl Default for GitSymbolTheme {
+    fn default() -> GitSymbolTheme {
+        GitSymbolTheme {
+            default: "_".into(),
```
zwpaper, line 23: It would be better to differ the `default` and `unmodified` to distinguish them

src/theme/git_symbol.rs
```diff
@@ -0,0 +1,52 @@
...
+            unmodified: "_".into(), // "\u{f00c}
+            new_in_index: "\u{f067}".into(),
+            new_in_workdir: "?".into(),
+            deleted: "\u{f014}".into(), // or f06
+            modified: "\u{f8ea}".into(),
+            renamed: "\u{f8ea}".into(),
+            ignored: "!".into(),
```
zwpaper, line 30: I would prefer a less eye catch symbol for `ignore`, like `I` in `exa`

src/git_symbol.rs
```diff
@@ -0,0 +1,38 @@
...
+impl GitSymbols {
+    pub fn new(theme: FlagTheme, icon_separator: String) -> GitSymbols {
+        let git_theme = match theme {
+            IconTheme::Unicode => GitSymbolTheme::unicode(),
+            IconTheme::Fancy => GitSymbolTheme::default(),
+        };
+
+        Self {
+            theme: git_theme,
+            separator: icon_separator,
```
zwpaper, line 18: I would prefer not to add this separator here, it looks better without it:

image

zwpaper (changes requested):
hi @hpwxf, sorry again for the late reply, I was swamped recently, I had added more comments and it would be ok to go after we resolve them.

thanks so much for the PR, it looks great and we only have some nit to discuss.

let's make the magic happen!

zwpaper:
hi @hpwxf, do you have some cycle for this PR, we are planning for 1.0 release, we can try to make it ship with the 1.0

hpwxf:
> hi @hpwxf, do you have some cycle for this PR, we are planning for 1.0 release, we can try to make it ship with the 1.0

Sorry for my discontinued availability. I will do the requested changes today or tomorrow.

src/git.rs
hpwxf: Dropped.
`_cachedir` has been also removed since only used as log

src/git.rs
hpwxf: removed line. Only used as log.

src/theme/git_symbol.rs
hpwxf: Done.

src/git_symbol.rs
hpwxf: Done. I also removed the separator option (always no space).

src/theme/git_symbol.rs
hpwxf: Now default is a simple space ` `

src/meta/git_file_status.rs
hpwxf: Done for all in this file.

src/git.rs
hpwxf: Done.

src/git.rs
hpwxf: The `Err` case is not a real error. Git2 cannot retrieve git status info. Usually it means that the directory is not a git directory.

hpwxf:
Thx for the review. Everything solved (see comments for details)

zwpaper:
hi @hpwxf, thanks for the great work again, I have only two more nit questions

1. the style is different in `unicode` and `default`
   image
2. the space may cause confusion when most of the files are unchanged, but actually I like the clean spaces personally 🤣
   image

Maybe a more consistent choice for the default would be better.

how about a dash `-` for default, and dot `.` for unmodified?

image

hpwxf:
@zwpaper Thank you for your feedback.
Do you prefer ASCII chars in any mode?
Without the space separator, the symbols are sometimes too close:
Capture d’écran 2023-04-26 à 19 33 11
Never this issue with simple ASCII chars:
Capture d’écran 2023-04-26 à 19 33 40

zwpaper:
hi @hpwxf, I would prefer icons, or Unicode, actually, but as you said that it may be too close in some cases, that would be a problem, and it would be too far away if a space added🤣

how about we use ASCII in this first version, and try to figure out a theme that fits all cases?

and we can also let users create a theme file to custom it themselves later.

hpwxf:
Hi @zwpaper ;
An idea seems to be:
* default symbols using ASCII chars
* custom symbols using a theme (the user will be able to add a space for large symbol)

I will work on that. (I will first update to only use ASCII chars and then about the theme; if you prefer the git theme can be done in a next PR).

hpwxf:
Git theme done:

In config file, you can define:
```yaml
# == Git ==
git:
  # Whether to display git status
  # Possible values: false, true
  enabled: false
  # How to display git status
  # When "classic" is set, this is set to "default".
  # Possible values: default, <theme-file-name>
  # when specifying <theme-file-name>, lsd will look up theme file
  # XDG Base Directory if relative, e.g. ~/.config/lsd/themes/<theme-file-name>.yaml,
  # The file path if absolute
  theme: git
```

and `git.yaml` can be:
```yaml
default: "-"
unmodified: "."
new-in-index: 
new-in-workdir: "?"
deleted: D
modified: ✍
renamed: R
ignored: I
typechange: T
conflicted: C
```

I still have a small bug that always enable git when `--config-file` is used (even if `enabled: false`)

zwpaper:
Hi, @hpwxf thanks so much for implementing the theme in this short time!

But I prefer a separate PR for the theme features.

One reason is that we would have more discussions about it like, we were planning to use a fixed theme file name, theme should be optional so that user doesn't have to offer all items, eg., keeping it simple would let us release the 1.0 in a shorter time.
Another one is that I am planning to squash the PR as this is one feature and tens of commits may be a nightmare for others to rebase onto.

zwpaper:
Ps it would also be great if we make the git theme (in a separate PR) in time to release with the 1.0

hpwxf:
Ok @zwpaper ; all done.

* First I finalize a functional code about a theme for git (only symbols; it could be only with colors)
* Then I did a commit to remove the git theme and I saved the previous work in another branch for a next PR
* I squashed all the commits before your latest review (the latest commits are not yet squashed, but one reviewed it could be a good thing for sure)

tummetott:
@hpwxf any news when the git-theme branch will be merged? Is there an issue which tracks the announced discussion around the git-theme?

PR #830, opened 2023-03-29
Fix obsolete Nerd Font icons

nix6839:
Replace the old code point with the new code point.
See: https://github.com/ryanoasis/nerd-fonts/releases/tag/v2.3.3

> Material Design Icons
> The Material Design Icons have to move codepoints. Historically we patch them in between F500 - FD46; but part of that range is needed for Chinese and other glyphs. With https://github.com/ryanoasis/nerd-fonts/pull/773 and Plan 5+ of https://github.com/ryanoasis/nerd-fonts/issues/365 their new home will be after F0001. And they grow considerably.

> The v2.3.x release has both, the 'old' and the 'new' codepoints to enable you an easier transision. Please change already now your prompts and other uses to the new codepoints. Release v3.0.0 will finally drop the old codepoints (F500 - FD46).

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add many icons from https://github.com/Peltoche/lsd/issues/764 [@TruncatedDinosour](https://ari-web.xyz/gh)
 - Add support for localization from [scarf](https://github.com/scarf005)
 - Add icons for cjs, cts and mts from [Han Yeong-woo](https://github.com/nix6839)
+- Fix obsolete Nerd Font icons from [Han Yeong-woo](https://github.com/nix6839)
 
 ### Fixed
 - Do not quote filename when piping into another program from [TeamTamoad](https://github.com/TeamTamoad)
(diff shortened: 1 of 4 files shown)


zwpaper (changes requested):
can you help to add some notice under FAQ about this change?

nix6839:
> can you help to add some notice under FAQ about this change?

Where is the FAQ? Should I add it to the README section?

nix6839:
@zwpaper  I've added a FAQ. I'm not a native English speaker so please let me know if you see anything strange in the FAQs.

zwpaper:
hi @nix6839, I have one more question, Are the pre-patched fonts updated to this latest v2.3.x version? https://github.com/ryanoasis/nerd-fonts#patched-fonts

looks like the [Hack Nerd](https://github.com/ryanoasis/nerd-fonts/tree/master/patched-fonts/Hack) I was using did not show the expected icons

nix6839:
If you look at https://washnix.com/hack-nerd/, it is patched correctly.

The source code of the page: https://github.com/nix6839/hack-nerd/blob/main/index.html

zwpaper:
hi @nix6839, I double check for the icons, it works like a charm, will merge after CI passed, thanks again!

nix6839:
Thank you for your hard work.

PR #833, opened 2023-04-06
:sparkles: chore: add new bug template

zwpaper:
ref: https://github.com/zwpaper/lsd/issues/new?assignees=zwpaper&labels=bug&template=bug.yml&title=%5BBug%5D%3A+

this can help to gather information easily.

image


zwpaper:
the form also notifies users to reproduce the bug in latest release, add tell us how they install lsd

PR #834, opened 2023-04-14
change executable file extentions, add cmd, com, vbs, vbe, and dll

OldWorldOrdr:
cmd is another way of naming bat files, com is an old filetype that afaik isn't really used anymore but its still technically treated as a program by windows.  msi on the other hand isn't actually a program, its just an installer package that gets interpreted by msiexec.exe, it would be like saying that deb files are executable on linux.

### DIFF
--- a/src/meta/filetype.rs
+++ b/src/meta/filetype.rs
@@ -16,7 +16,7 @@ pub enum FileType {
 
 impl FileType {
     #[cfg(windows)]
-    const EXECUTABLE_EXTENSIONS: &[&'static str] = &["exe", "bat", "ps1", "cmd", "com", "msi" ];
+    const EXECUTABLE_EXTENSIONS: &[&'static str] = &["exe", "msi", "bat", "ps1", "cmd", "com" ];
 
     #[cfg(unix)]
     pub fn new(
--- a/src/meta/filetype.rs
+++ b/src/meta/filetype.rs
@@ -16,7 +16,7 @@ pub enum FileType {
 
 impl FileType {
     #[cfg(windows)]
-    const EXECUTABLE_EXTENSIONS: &[&'static str] = &["exe", "msi", "bat", "ps1"];
+    const EXECUTABLE_EXTENSIONS: &[&'static str] = &["exe", "bat", "ps1", "cmd", "com"];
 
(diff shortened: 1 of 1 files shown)


zwpaper:
hi, @OldWorldOrdr, you are right about the msi extension, but it also feels like some kinda widely used executable, how about we still treat the msi as executable?

OldWorldOrdr:
@zwpaper done, last thing I'm doing before going to bed.

OldWorldOrdr:
ok that was really the last thing

OldWorldOrdr:
Should I add script extensions like py, sh, or lua to the list (when I wake up)

OldWorldOrdr:
Considering we have bat/cmd and ps1 as executable file extensions, I'll add other scripts too.

OldWorldOrdr:
I decided to not include stuff that's usually run with `someprogram file.whatever` rather then `file.whatever` or `./file.whatever` or `.\file.whatever`, since those kind of programs usually aren't made executable when running on linux, you don't see js files marked as executable most of the time for example.  I added dll files because it seems to be tradition to mark shared libraries as executable in the unix world.

OldWorldOrdr:
I found something very interesting, https://github.com/rmyorston/busybox-w32 this guy ported busybox to windows, and it can somehow determine executable files from the contents
Screenshot (25)
and its still faster then lsd
Screenshot (26)

OldWorldOrdr:
if there are a lot of files or the files are really big then its slower the first time, but faster every time after that.
Screenshot (27)

OldWorldOrdr:
I'm closing this because I found a better way to do it, but I don't know enough rust to make it happen.  The environment variable `PATHEXT` contains a list of file extensions to expand when run from the shell, so if theres a file called test.exe in the current directory and you run ./test it will run test.exe.  The default value is `.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC`

Also I found that lsd doesn't pick up extensions if the case is different, so .exe would be detected but .EXE or .Exe would not

OldWorldOrdr:
also some stuff that is definitely executable should be added to that list, like PS1, even if it's not in the env variable

PR #843, opened 2023-04-30
Add Version Sort

juansc:
# Description

Addresses https://github.com/lsd-rs/lsd/issues/801. The root cause is that the external library that is being used does not have sufficient test coverage and is no longer maintained.

I took a look at how [GNU sort is implemented](https://github.com/coreutils/coreutils/blob/master/doc/sort-version.texi) and it looked simple enough to be included into this repo. I added a few test cases, including the one from the issue. 

I'm still pretty new to Rust, so any code/organizational feedback is greatly appreciated

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests (You tell me 🙂)
- [x] Add changelog entry


juansc:
It looks like I have some refactoring artifacts. I'll fix and add more documentation

src/sort.rs
```diff
@@ -1,6 +1,6 @@
 use crate::flags::{DirGrouping, Flags, SortColumn, SortOrder};
 use crate::meta::Meta;
-use human_sort::compare;
+use crate::versionsort::compare;
```
nim65s, line 3: I think you didn't meant to simply remove human-sort, but this is what you are doing here. 

Did you forget to implement some cli / config switch ?

nim65s (changes requested):

juansc:
I think it's cleaner to break out the sorting logic into a [separate crate](https://github.com/juansc/vsort/). Once it's published I'll update this PR to use that crate instead. The code in the create has been updated and has more robust testing

src/sort.rs
juansc: I do mean to replace the sorting library. The new code is now its own separate crate and the PR now reflects that.

zwpaper:
hi @juansc, I have read the PR and vsort crate, and it seems great!

please add a changelog here https://github.com/lsd-rs/lsd/blob/master/CHANGELOG.md and we should be good to go.

/lgtm

juansc:
@zwpaper Changelog has been added

zwpaper:
thanks so much @juansc

Issue #749, opened 2022-10-05
should change the color theme path to `$XDG/lsd/colors.yaml` instead of current `$XDG/lsd/themes/NAME.yaml`


zwpaper:
as mentioned here https://github.com/Peltoche/lsd/pull/707#issuecomment-1198461950 and here https://github.com/Peltoche/lsd/pull/707#discussion_r976023899, we should change the color theme path to `$XDG/lsd/theme.yaml` instead of current `$XDG/lsd/themes/NAME.yaml`

should add a check and warning? if theme file has a wrong location.

PR #851, opened 2023-05-28
Added new `custom` option for color config, marked themes folder as deprecated.

peppidesu:
#749: when `color: custom` is specified, `lsd` will look for $XDG/lsd/colors.yaml and load the theme. The old custom theme behaviour is marked as deprecated and will show a warning when used.

- Renamed `ThemeOption::Custom` to `ThemeOption::CustomLegacy`
- Added new `ThemeOption::Custom`

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry


README.md
```diff
@@ -121,10 +121,8 @@ color:
   when: auto
   # How to colorize the output.
   # When "classic" is set, this is set to "no-color".
-  # Possible values: default, <theme-file-name>
-  # when specifying <theme-file-name>, lsd will look up theme file
-  # XDG Base Directory if relative, e.g. ~/.config/lsd/themes/<theme-file-name>.yaml,
-  # The file path if absolute
+  # Possible values: default, custom
+  # When "custom" is set, lsd will look in the config directory for `colors.yaml`.
```
zwpaper, line 125: I have reconsidered the logic, how about we drop the custom logic, and detect the `colors.yaml` directly, if the colors file is found, we use it as a custom theme.

so that we don't have to update the colors theme file and both this configuration, it seems a little bit duplicated

zwpaper (changes requested):

README.md
peppidesu: Can you please clarify what you mean by that? I don't see the issue of the code duplication here, since one is deprecated anyway and should be phased out at some point.

zwpaper: sorry for not explaining myself.

I was saying that we should drop the `custom` option for the theme option because we could detect whether the `colors.yaml` file existed.

it's not about codes, it is about usage

zwpaper: hi @peppidesu, did you have some time to work on this PR again?

let me explain myself more clear,
1. we deprecated all the theme configuration
2. we detect the existence of `colors.yaml` file every run
3. if `colors.yaml` existed, we use it directly
4. if not, we fallback to default

WDYT?

peppidesu: In that case we should do the same for icons, since it has a field for default/custom.

zwpaper: Hmmm...make sense, let's leave those to later improvement

PR #852, opened 2023-06-02
Added complete color theming support for Git

k4yt3x:
I've noticed that the recently merged Git feature only has one default color, so I completed it. I've added a default color theme for it that looks like this (ignore the conflicted, I didn't bother actually making a repo for it):

image

... and with this theme of mine:

image

... it looks like this, works fine:

image

Edit: also suppressed ErrorKind::NotFound for Git status since if you don't this could happen when you are in a directory that's been deleted:

image

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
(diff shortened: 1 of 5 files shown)


zwpaper:
thanks @k4yt3x! nice work!

PR #855, opened 2023-06-12
Add zsh fish bash completion support

niT-Tin:
Add zsh fish bash completion support
---
Place the completion file of the corresponding shell in the completion folder in the specified location

### DIFF
--- a/completions/bash/lsd
+++ b/completions/bash/lsd
@@ -1,0 +1,93 @@
+# Place the content of this file in .bashrc or related bash configuration files, or activate the file by using the source command in the configuration file
+
+parse_long() {
+  local -n pl_ret="$1"
+  pl_ret=()
+  local -A lt
+  local exp_l='(\-\-[[:alnum:]-]+)'
+  : "$(lsd --help)"
+  for res_l in ${_}
+  do
+     [[ "$res_l" =~ $exp_l ]] && { [[ -z ${lt[${BASH_REMATCH[1]}]} ]] && lt[${BASH_REMATCH[1]}]=1; }
+  done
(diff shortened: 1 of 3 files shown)


niT-Tin:
and the specific effect is as follows
bash
lsdbash
zsh
lsdzsh
fish
lsdfish

nim65s:
Hi,

What is the point of adding those generated files here ?

I think this should be closed in favor of #861, or something like that.

niT-Tin:
> 

Oh, thank you for your reminder. I did not notice that there was already a completion operation, but using LSD on my computer did not have this effect, so I wrote these shell completion myself. If there is no problem running lsd completion on someone else, then this pr can indeed be closed

nim65s:
Arch already provide them : https://gitlab.archlinux.org/archlinux/packaging/packages/lsd/-/blob/main/PKGBUILD#L42-L44
Nix too: https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/misc/lsd/default.nix#L28

You probably just need to set `SHELL_COMPLETIONS_DIR`

niT-Tin:
> Arch already provide them : https://gitlab.archlinux.org/archlinux/packaging/packages/lsd/-/blob/main/PKGBUILD#L42-L44 Nix too: https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/misc/lsd/default.nix#L28
> 
> You probably just need to set `SHELL_COMPLETIONS_DIR`

Wow, thanks, I'll try

zwpaper:
https://github.com/lsd-rs/lsd/blob/8acaabec723cdd71b498212c8a47ac952d022738/build.rs#L32-L35 

we already support this

/close

Issue #841, opened 2023-04-24
Filesizes shown in bytes not colored with size.medium and size.large


pstray:
- os: linux
- `lsd --version`:  lsd 0.23.1

## Expected behavior
I had hoped for files colors be applied to sizes even when '--size bytes'.

## Actual behavior
With '--size bytes', all files sizes are colored with size.small

zwpaper:
/close

as #856 merged

PR #856, opened 2023-06-15
Fixed file size colorizing with `--size=bytes` argument

bells307:
Fixed incorrect colorizing with `--size=bytes` described in issue #841 

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add changelog entry


zwpaper:
thanks for contributing @bells307

PR #860, opened 2023-06-21
:tada: add zwpaper to code owner

zwpaper:
it was needed to be listed on codeowner file in order to use the `/approve` command, so I am adding myself to the CODEOWNER file

/assign @meain


meain:
We should have done this earlier 😅

meain (approved):

Issue #857, opened 2023-06-15
[ENHANCEMENT] Implementing ls -p argument 


D3vil0p3r:
Is it possible to implement the `-p` argument of 'ls' in 'lsd'?

zwpaper:
do you mean the following?

```
       -p, --indicator-style=slash
              append / indicator to directories
```

D3vil0p3r:
Yes

PR #863, opened 2023-06-25
add indicator style

nim65s:
Hi,

I saw a "good first issue" on #857, so I gave it a try :)
---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry


nim65s:
NB: I tried to keep compatibility with previous options / config files, but it might be cleaner to allow some refactoring here

PR #873, opened 2023-08-06
lsd does not seem to be available in Ubuntu repositories

JJ:
This is the error I get:

```
No hay un paquete apt "lsd", pero hay un snap con ese nombre.
Intente «snap install lsd»
```

Spanish for: no `lsd` package, try... what I've written here

---
#### TODO

- [ ] Use `cargo fmt`
- [ ] Add necessary tests
- [ ] Add changelog entry


zwpaper:
hi @JJ, thanks for the update, but we found some problems using snap, and we already deprecated it(check the line below your modified).

and refer to the package status here: https://github.com/lsd-rs/lsd#installation, we should have the 0.23.1 in ubuntu, can you have a look or try `apt install rust-lsd`?

JJ:
I'm in Ubuntu 22.04, and snap seems to work without a hitch. I've tried apt search for lsd, and it seems not to be available in any form. Closing this anyway since it was in the wrong row.

Issue #894, opened 2023-09-03
Unable to disable quotes (no support for: `--quoting-style`, `QUOTING_STYLE`, or `--literal`)


globau:
### Version

- [X] latest

### version

lsd 1.0.0

### What OS are you seeing the problem on?

macOS

### installation

homebrew

### term

xterm-256color

### ls-colors

di=1;36:ln=35:so=32:pi=33:ex=31:bd=34;46:cd=34;43:su=30;41:sg=30;46:tw=30;42:ow=30;43

### What happened?

```markdown
`error: unexpected argument '--literal' found`
```

### What expected?

```markdown
`lsd` should honour `lsd --literal`, `lsd --quoting-style=literal`, and/or `QUOTING_STYLE=literal lsd`
```

### What else?

```markdown
Sadly #746 added quotes around filenames without any way to opt out.

See https://www.gnu.org/software/coreutils/quotes.html for the various options that GNU's `ls` provides.

Adding just `literal` initially would be valuable to those who are not in favour of this somewhat controversial change.
```

zwpaper:
thanks for the advice! 
/kind feature
/good-first-issue
/help

PR #900, opened 2023-09-11
Add support for --literal flag and literal entry in config

PanGan21:
Add support for `--literal` in order to opt out showing filenames with quotes
Closes: #894

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [v1.0.0] - 2023-08-25
 
 ### Added
+- Add support for `--literal` from [PanGan21](https://github.com/PanGan21)
 - Add CI to build aarch64 macOS target and skip on test [#878](https://github.com/lsd-rs/lsd/pull/878) from [zwpaper](https://github.com/zwpaper)
 - Add complete color theming support for Git [k4yt3x](https://github.com/k4yt3x)
 - Add [Git integration](https://github.com/Peltoche/lsd/issues/7) from [hpwxf](https://github.com/hpwxf)
--- a/doc/lsd.md
+++ b/doc/lsd.md
@@ -140,8 +140,8 @@ lsd is a ls command with a lot of pretty colours and some other stuff to enrich 
 `--header`
 : Display block headers
 
-`--literal`
-: Do not show quotes on filenames
+`-N --literal`
+: Print entry names without quoting
 
 # ARGS
 
--- a/doc/lsd.md
+++ b/doc/lsd.md
@@ -140,6 +140,9 @@ lsd is a ls command with a lot of pretty colours and some other stuff to enrich 
 `--header`
 : Display block headers
 
+`--literal`
+: Do not show quotes on filenames
+
 # ARGS
 
 `<FILE>...`
--- a/src/app.rs
+++ b/src/app.rs
@@ -181,8 +181,8 @@ pub struct Cli {
     #[arg(long, hide = !cfg!(windows))]
     pub system_protected: bool,
 
-    /// When showing files do not quote their names
-    #[arg(long)]
+    /// Print entry names without quoting
+    #[arg(short = 'N', long)]
     pub literal: bool,
 
     /// Print help information
--- a/src/app.rs
+++ b/src/app.rs
@@ -181,6 +181,10 @@ pub struct Cli {
     #[arg(long, hide = !cfg!(windows))]
     pub system_protected: bool,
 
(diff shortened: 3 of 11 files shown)


doc/lsd.md
```diff
@@ -140,6 +140,9 @@ lsd is a ls command with a lot of pretty colours and some other stuff to enrich
 `--header`
 : Display block headers
 
+`--literal`
```
zwpaper, line 143: according to https://man7.org/linux/man-pages/man1/ls.1.html

we can add `-N` as short name

doc/lsd.md
```diff
@@ -140,6 +140,9 @@ lsd is a ls command with a lot of pretty colours and some other stuff to enrich
...
+: Do not show quotes on filenames
```
zwpaper, line 144: > print entry names without quoting

ref: https://man7.org/linux/man-pages/man1/ls.1.html

src/app.rs
```diff
@@ -181,6 +181,10 @@ pub struct Cli {
     #[arg(long, hide = !cfg!(windows))]
     pub system_protected: bool,
 
+    /// When showing files do not quote their names
+    #[arg(long)]
+    pub literal: bool,
```
zwpaper, line 186: `-N` and description align with https://man7.org/linux/man-pages/man1/ls.1.html

src/core.rs
```diff
@@ -71,7 +73,7 @@ impl Core {
             // or require a raw output (like the `wc` command).
             inner_flags.layout = Layout::OneLine;
 
-            flags.should_quote = false;
```
zwpaper, line 74: how about changing `should_quote` to `literal`?

zwpaper (changes requested):
thanks for the contribution!

it mostly looks great! only with some nit we can discuss

doc/lsd.md
PanGan21: Should be fixed now!

doc/lsd.md
PanGan21: Fixed it as well according to the manual

src/app.rs
PanGan21: Fixed it!

src/core.rs
PanGan21: Changed!

src/meta/name.rs
```diff
@@ -153,21 +153,21 @@ impl Name {
                 format!(
                     "{}{}",
                     icons.get(self),
-                    self.hyperlink(self.escape(self.file_name(), quote), hyperlink)
+                    self.hyperlink(self.escape(self.file_name(), !quote), hyperlink)
```
PanGan21, line 156: Let me know if you think `quote` arg should change to literal as well in order to avoid this negation here

zwpaper: Yes! it would be great if you can also update this one!

PanGan21: Fixed it as well!

zwpaper:
hi @PanGan21, can you fix the CI errors

PanGan21:
> hi @PanGan21, can you fix the CI errors

I think I am confused a bit with the errors of the ci. Could they be caused because I did `cargo update` and updated the `cargo.lock` file?
I did that because when building the project locally I was getting the following error:
`error[E0635]: unknown feature `proc_macro_span_shrink` which was fixed after updating cargo versions

PanGan21:
> hi @PanGan21, can you fix the CI errors

I was using wrong rustc version, so I had to update `Cargo.lock`. Now I reverted it and used the correct rust version.
Could you run the ci one more time to check if everything is fixed now?

zwpaper:
thanks for the contribution! @PanGan21

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [v1.0.0] - 2023-08-25
 
 ### Added
+- Add support for `--literal` from [PanGan21](https://github.com/PanGan21)
```
shilangyu, line 17: Should this have gone to an already-released v1.0.0?

zwpaper: good catch! actually, I was planning to use the GitHub auto-generate release note, this file may not be necessary, can you help create a PR to fix it and add a notice that we are no longer using this file, we use the GitHub auth-generate one instead.

shilangyu: Not exactly sure what you meant, but opened one here: https://github.com/lsd-rs/lsd/pull/907

Let me know if it is what you meant.

PR #901, opened 2023-09-13
RPM spec file for lsd

daniejstriata:
Basic SPEC file that can be used to build a RPM for lsd. 
  - UPX compression was disabled.

I have precompiled packages on my fork for EL9 and AMZN AL2023.
https://github.com/daniejstriata/lsd/releases


zwpaper:
hi @daniejstriata, thanks for the contribution, can you also add a ci to build the rpm? so that the spec file could be used and we could maintain a rpm release

daniejstriata:
> hi @daniejstriata, thanks for the contribution, can you also add a ci to build the rpm? so that the spec file could be used and we could maintain a rpm release

Hi @zwpaper I've not built rpms before using ci on github. I'll investigate it.

zwpaper:
hi @daniejstriata, how is this PR going? maybe we could merge it first if the spec is completed.

and let the ci be another PR

daniejstriata:
@zwpaper I've built lsd in copr here
https://copr.fedorainfracloud.org/coprs/faramirza/lsd/
for various rpm based distros. The difference between the spec I must here and the one in my fork is that the on I'm pushing here assumes that rust and cargo is already installed whereas the other version will install the version for rust available to the distro in its repos.

I'm still unsure about the ci on github. I've never used it before. I've not yet been able to look into that. Do you know of a recommend Guide I can look at? 

We could cancel this PR and redo it once more with the newer spec?

zwpaper:
let's merge this one and you could always work on the CI later, for CI, please check: https://github.com/features/actions

daniejstriata:
@zwpaper I had a look. How is the Fedora package built and should that SPEC file not be added back here rather than mine?

PR #902, opened 2023-09-15
Add windows-acl-permissions feature to fix perf issue on windows

domsleee:
Attempted fix for performance issue mentioned in https://github.com/lsd-rs/lsd/issues/200, for machines on a domain.

## Benchmark
For a folder with 50 items.

Before change:
`lsd` took ~40 seconds

After change: 
```pwsh
❯ hyperfine "lsd" --runs 50
Benchmark 1: lsd
  Time (mean ± σ):      39.4 ms ±   8.3 ms    [User: 0.6 ms, System: 0.6 ms]
  Range (min … max):    33.6 ms …  91.8 ms    50 runs
```

## Other notes
I think it makes more sense for this to be disabled by default on windows due to performance, but I could be swayed.
My suggestion would be to use https://github.com/lsd-rs/lsd/issues/866#issuecomment-1721012397 as the default windows permissions behaviour.

P.S thanks for this great tool 🎉

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests (Not applicable imo)
- [ ] Add changelog entry

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -69,6 +69,7 @@ serial_test = "0.5"
 default = ["git2"]
 sudo = []
 no-git = [] # force disabling git even if available by default
+windows-acl-permissions = []
 
 [profile.release]
 lto = true
--- a/src/meta/windows_utils.rs
+++ b/src/meta/windows_utils.rs
@@ -5,7 +5,7 @@ use std::os::windows::ffi::{OsStrExt, OsStringExt};
 use std::path::Path;
 
 use windows::Win32::Foundation::PSID;
-use windows::Win32::Security::{self, Authorization::TRUSTEE_W, ACL};
+use windows::Win32::Security::{self, ACL};
 
 use super::{Owner, Permissions};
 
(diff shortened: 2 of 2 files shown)


zwpaper:
hi @domsleee, thanks for the contribution, I am working on https://github.com/lsd-rs/lsd/pull/882 to add a runtime flag instead of a build time feature for this.

can you have a look at that, and we can discuss which one would be good for users.

domsleee:
Hi @zwpaper, a runtime flag is probably the better approach. I see in your PR that you are handling this issue as well, nice:
> lsd: .: The specified domain either does not exist or could not be contacted. (os error 1355).

I have been able to reproduce this os error 1355 by disconnecting from the internet from my work machine.

I found a similar issue here: https://github.com/docker/for-win/issues/2131#issuecomment-402941794

I used [process monitor](https://learn.microsoft.com/en-us/sysinternals/downloads/procmon) and found, somewhat hilariously, that `GetEffectiveRightsFromAclW` is calling `FSCTL_PIPE_TRANSCEIVE` to fetch a file from a remote server!

Each read takes about a second, which explains why it was taking 45s for my 50 files 😄

domsleee:
Since https://github.com/lsd-rs/lsd/pull/882 is merged, I will close this one and open another PR for this comment: https://github.com/lsd-rs/lsd/pull/882#issuecomment-1722126112

PR #905, opened 2023-09-16
Add configuration and CLI options: truncate owner

bbc2:
Hello! This is my first attempt to contribute something meaningful to this project so I don't know I'm doing it right. Thank you in advance for your feedback.

This adds the following CLI flags:

- `--truncate-owner-after`
- `--truncate-owner-marker`

And the following configuration fields:

```yaml
truncate-owner:
  after:
  marker: ""
```

The default behavior of LSD is unchanged.

The problem this change attempts to solve is the usability of the `-l` flag on systems where some user or group names are long but cannot be changed (e.g. the user is not admin and the account is managed in a central directory). In such cases, even with a decently sized terminal (90+ characters wide), lines often overflow, making the directory listing hard to read.

Without this change, the only mitigation would consist in turning off the display of file ownership (via the `blocks` configuration field) which is unsatisfactory because ownership information is very useful.

---

Default settings (the length of user/group name may seem exaggerated but it's actually the same length as my user/group name on machines I have to use):

lsd_before

With `--truncate-owner-after 7 --truncate-owner-marker …`:

lsd_after

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+- Add CLI parameters `--truncate-owner-after` and `--truncate-owner-marker` (and equivalent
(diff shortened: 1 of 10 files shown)


zwpaper:
Thanks for the great work!

PR #906, opened 2023-09-17
human readable flag (-h) enabled

sdorr0:
This re-enables a quick, easy to type command line flag that `ls` users are familiar with for convenience.

In the event a user has changed the default configuration of `lsd` to use a size setting of something other than `default`, this lets a user use the historically established CLI flag (`-h`) to enable the human readable option without having to type `--size=default` on the command line.

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [x] Add changelog entry

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -54,7 +54,7 @@ pub struct Cli {
 
     /// For ls compatibility purposes ONLY, currently set by default
     #[arg(short, long)]
-    human_readable: bool,
+    pub human_readable: bool,
 
     /// Recurse into directories and present the result as a tree
     #[arg(long)]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
+## [Unreleased]
+
+### Added
+
+### Fixed
+
+### Changed
+- Human readable size flag `-h` re-enabled. [sdorr0](https://github.com/sdorr0)
+
 ## [v1.0.0] - 2023-08-25
 
 ### Added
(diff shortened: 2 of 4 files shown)


zwpaper:
hi @sdorr0, thanks for the contribution.

I have one more question, what is users used `-h` with `--size` option? this would make `--size` not working?

sdorr0:
> hi @sdorr0, thanks for the contribution.
> 
> I have one more question, what is users used `-h` with `--size` option? this would make `--size` not working?

Hmm.  Yeah, the way the code is written now, `-h` will override a `--size` on the command line.  I see that subsequent `--size` flags on the command line will override one another based on position (the 'last' one wins).

Never done Rust before.  I'm not sure if there's a way to set up `#[arg()]` to also incorporate a short `-h` so it can play nicely with those positional dependencies, but I can research.

Would this be a requirement to merge this change in?

sdorr0:
I may have a solution

sdorr0:
Never mind.  I've tried all kinds of approaches and can't seem to get clap to let `-h` work with `--size=<mode>` where the last one on the command line always wins.

So, yes, using `-h` will take precedence over a supplied `--size=<mode>` argument.

zwpaper:
but it does not make sense for `-h` as a general flag to overwrite the `--size` as a specific flag.

many people may set the `-h` as default and it may lost `--size` if that

src/app.rs
```diff
@@ -52,9 +52,9 @@ pub struct Cli {
     #[arg(short = 'R', long, conflicts_with = "tree")]
     pub recursive: bool,
 
-    /// For ls compatibility purposes ONLY, currently set by default
+    /// This will take precedence over any --size=<mode> option
```
spenserblack, line 55: IMO this can be described as "alias for --size=default (included for compatibility with ls)" or something along those lines, then configure this argument and `--size` to conflict with each other. So instead of choosing precedence when both are used, one must use either one or the other (or none).

zwpaper:
thanks for the contribution, but we currently do not plan to change this default option.

I am closing this PR and feel free to find another issue to work on.

PR #907, opened 2023-09-20
Add changelog deprecation notice

shilangyu:
Created out of: https://github.com/lsd-rs/lsd/pull/900#discussion_r1329966830

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 # Changelog
+
+NOTE: This file is no longer being updated. For release notes follow [GitHub releases](https://github.com/lsd-rs/lsd/releases).
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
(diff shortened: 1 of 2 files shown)


shilangyu:
The pull request template still mentions that a changelog entry should be added. Should I remove it?

CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 # Changelog
+
+NOTE: This file is no longer being updated. For release notes follow [GitHub releases](https://github.com/lsd-rs/lsd/releases).
```
zwpaper, line 3: by saying GitHub auto-generates release notes, I mean https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes

We don't have to update this file and will generate by PRs when releasing.

zwpaper:
can you also help drop this line: https://github.com/lsd-rs/lsd/blob/master/.github/PULL_REQUEST_TEMPLATE#L9

CHANGELOG.md
shilangyu: Not sure what you would like me to do here. I understand what you are trying to achieve, but not what is the future of this CHANGELOG file. What should the notice say? If there should be a notice at all?

Issue #866, opened 2023-06-28
[Feature request] Show Windows Attributes when running on Windows


RaffaeleBianc0:
- os: Win10pro
- `lsd --version`: lsd 0.23.1

## Expected behavior
**dar--**  28/06/2021 13:33:00 4.0 KB  Users/
When run from a Windows machine, the first column could show the attributes the same way "Get-ChildItem" does on PowerShell on Windows.

## Actual behavior
**drwxrwxrwx**  28/06/2021 13:33:00 4.0 KB  Users/

(thanks for this neat piece of software, i am using it every time!)

zwpaper:
Hi @RaffaeleBianc0, I am not a windows user, can you explain some more about the `dar--`?

domsleee:
There is also `l` for links (e.g. symlinks / junctions)

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-childitem?view=powershell-7.3

> By default Get-ChildItem lists the mode (Attributes), LastWriteTime, file size (Length), and the Name of the item. The letters in the Mode property can be interpreted as follows:

> l (link)
d (directory)
a (archive)
r (read-only)
h (hidden)
s (system)

[eza](https://github.com/eza-community/eza) uses these attributes for windows

ofersadan85:
@zwpaper Shall i write a PR for this?

PR #911, opened 2023-09-25
feat: Windows safe default permissions (fixes ACL errors/performance)

domsleee:
Fixes: #200 by implementing #866 as the default permission setting for windows.

### Context
The current default (`--permission rwx`) uses ACL on windows, which can throw errors and has performance problems, described in #200 and also in https://github.com/lsd-rs/lsd/pull/882#issuecomment-1722126112:

> * `GetEffectiveRightsFromAclW` can't access the domain, and os error 1355 is thrown
> * `GetEffectiveRightsFromAclW` works, but it can be very slow when the information about permissions requires files from a remote server [1 second per file for me, worse for others]

### Suggestion
1. A new `--permission attributes` is introduced
2. It is used as the default for windows

### Prior art
* powershell uses it in [Get-ChildItem](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/get-childitem?view=powershell-7.3#example-1-get-child-items-from-a-file-system-directory)
* [eza](https://github.com/eza-community/eza) uses it as the default for windows (`eza -al`)

#### Checklist

- [x] Use `cargo fmt`
- [x] Add necessary tests

### Screenshot & comparison
screenshot

### DIFF
--- a/src/meta/mod.rs
+++ b/src/meta/mod.rs
@@ -12,8 +12,9 @@ mod permissions;
 mod permissions_or_attributes;
 mod size;
 mod symlink;
-mod windows_attributes;
 
+#[cfg(windows)]
+mod windows_attributes;
 #[cfg(windows)]
 mod windows_utils;
 
--- a/src/meta/mod.rs
+++ b/src/meta/mod.rs
@@ -9,8 +9,10 @@ mod locale;
 pub mod name;
 mod owner;
 mod permissions;
+mod permissions_or_attributes;
 mod size;
 mod symlink;
+mod windows_attributes;
 
 #[cfg(windows)]
 mod windows_utils;
@@ -25,6 +27,7 @@ pub use self::links::Links;
 pub use self::name::Name;
 pub use self::owner::Owner;
 pub use self::permissions::Permissions;
+use self::permissions_or_attributes::PermissionsOrAttributes;
 pub use self::size::Size;
 pub use self::symlink::SymLink;
 pub use crate::icon::Icons;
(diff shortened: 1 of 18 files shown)


domsleee:
btw - I'm not sure why the `work-in-progress` label is still on this PR.
I think [muniu-bot](https://github.com/apps/muniu-bot) should have removed it when I marked the PR is ready, or perhaps I missed something 😄

Could you please review this one when convenient @zwpaper

roqvist:
Anyway to help this forward?

zwpaper:
sorry for the late reply, I will look into this recently

/assign
/milestone v1.1.0

src/meta/mod.rs
```diff
@@ -9,8 +9,10 @@ mod locale;
 pub mod name;
 mod owner;
 mod permissions;
+mod permissions_or_attributes;
 mod size;
 mod symlink;
+mod windows_attributes;
```
zwpaper, line 15: how about `#[cfg(windows)]` here, and drop the `#cfg` inside `windows_attributes`?

src/meta/windows_attributes.rs
```diff
@@ -0,0 +1,126 @@
+#[cfg(windows)]
```
zwpaper, line 1: maybe we can drop all the `#[cfg(windows)]` in this file, if we set it in `mod.rs`

zwpaper (changes requested):
@domsleee really sorry for the late reply, thanks so much for the great work! it LGTM mostly but only a nit

zwpaper:
hi @domsleee, as it has been some time since you responded, I have pushed a commit to make this PR go forward.

hope you don't mide

domsleee:
Hi @zwpaper, sorry I missed your review - thanks for reviewing and helping out with this PR 🎉

PR #935, opened 2023-10-16
Fix `literal` flag not being recognized from config file

Akmadan23:
The `literal` option was not correctly recognized from config file due to `Literal::from_cli` returning `Some(Literal(false))` instead of `None`, so `Literal::from_config` would never have been executed.

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [ ] Update default config/theme in README (not applicable)
- [ ] Update man page at lsd/doc/lsd.md (not applicable)

### DIFF
--- a/src/flags/literal.rs
+++ b/src/flags/literal.rs
@@ -14,12 +14,12 @@ impl Configurable<Self> for Literal {
     /// Get a potential `Literal` value from [Cli].
     ///
     /// If the "literal" argument is passed, this returns a `Literal` with value `true` in a
-    /// [Some]. Otherwise this returns `Literal` with value `false` in a [Some].
+    /// [Some]. Otherwise this returns [None].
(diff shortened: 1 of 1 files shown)


Akmadan23:
Sorry @zwpaper, I noticed you have been active in merging PRs recently, could you please take a look at this? It's a very simple fix to a real problem.

zwpaper:
hi @Akmadan23 , sorry for the late reply, I have started the test and will take a deeper look into it tomorrow, it looks good at a glance 😃

zwpaper:
hi @Akmadan23, I have checked your fix, and it works like a charm! thanks for the contribution!

please rebase your code onto master, so that CI can pass, and then we should be good to go

Akmadan23:
Now everything should be good to go

zwpaper:
thanks for the contribution!

PR #938, opened 2023-10-19
update various dependencies

decathorpe:
I'm the maintainer of lsd for Fedora Linux, and some outdated dependencies are making maintenance increasingly difficult.

- deps: update chrono-humanize to v0.2

We have this patched for almost two years in Fedora and it has not caused issues.

- deps: update assert_cmd to v2

Same here, this patch has been in the Fedora package for a while.

- deps: update sys-locale to v0.3

This dependency seems to have been added recently, not sure why an old version was chosen.

- deps: update vsort to v0.2

Same here, this was added recently but 0.1 was used instead of 0.2, not sure why.

- deps: update git2 to v0.18

Using old versions of git2 is not a good idea, since the bundled libgit2 C library often has CVE issues.

- deps: migrate from users to uzers

The "users" crate is unmaintained. The "uzers" crate is an API-compatible fork that also fixes some bugs and security issues.

- deps: update serial_test to v2

The current dependency (v0.5) is **reeeeeally** old. Not sure why this was never updated.

- deps: update predicates to v3

Same here, predicates v1 is **reaally** old.

- deps: allow newer versions of url, wild, and xdg crates

Not sure why strange `x.0.*` style dependencies were used here. It's holding back various updates for both url and xdg crates, and makes maintaining lsd in Fedora more difficult. We have built lsd against the latest versions of all three crates forever, and it has not caused issues.

### DIFF
--- a/.github/workflows/CICD.yml
+++ b/.github/workflows/CICD.yml
@@ -6,7 +6,7 @@ env:
   PROJECT_NAME: lsd
   PROJECT_DESC: "An ls command with a lot of pretty colors."
   PROJECT_AUTH: "Peltoche <peltoche@halium.fr>"
-  RUST_MIN_SRV: "1.69.0"
+  RUST_MIN_SRV: "1.70.0"
 
 on: [push, pull_request]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,6 +12,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "anstyle"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7079075b41f533b8c61d2a4d073c4676e1f8b249ff94a393b0595db304e0dd87"
+
+[[package]]
 name = "assert_cmd"
 version = "2.0.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


decathorpe:
PS: I'm not sure why you're keeping Cargo.lock pinned to really old versions. It looks like "cargo update" hasn't been run in ages. I'm sure there's tons of bugfixes (and potentially also security fixes) that happened between the versions pinned in Cargo.lock and the latest compatible versions.

zwpaper:
Hi @decathorpe, thanks so much for raising this, it would definitely help a lot!

I found predicates v3 need msrv 1.69, I have updated the msrv in master, can you help rebase your PR onto it?

decathorpe:
Ah, the MSRV is only checked in CI - it might be a good idea to set `package.rust-version` in Cargo.toml as well?

Thanks for taking a look, I'll rebase my PR.

zwpaper:
hi @decathorpe can you help to update the msrv to 1.70, so that we can pass the CI

decathorpe:
I've bumped the MSRV in the GitHub action and in Cargo.toml.

I also removed the hand-written MSRV check that was present in build.rs that is redundant with defining `package.rust-version` in Cargo.toml.

zwpaper:
thanks so much for your help @decathorpe!

zwpaper:
btw, I have sent you an invitation for the lsd-rs org package maintainers team, welcome aboard if you would like to 😀

decathorpe:
Great, thank you :)

PR #984, opened 2024-02-05
update various dependencies

decathorpe:
This is a followup for https://github.com/lsd-rs/lsd/pull/938.

- update dirs from v4 to v5 (no code changes)
- update term_grid from v0.1 to v0.2 (small code changes)
- update terminal_size from v0.1 to v0.3 (no code changes)
- update lscolors from v0.15 to v0.16 (no code changes)
- update serde_yaml from v0.8 to v0.9 (no code changes)

The term_grid update required small changes to the use of `term_grid::Cell` because with v0.2, it is possible to specify the alignment within the cell. Adding `Alignment::Left` should preserve the current behaviour.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,28 +4,86 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.18"
+version = "1.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
+checksum = "b2969dcb958b36655471fc61f7e416fa76033bdd4bfed0678d8fee1e2d07a1f0"
 dependencies = [
  "memchr",
 ]
 
 [[package]]
-name = "anstyle"
-version = "1.0.4"
+name = "android-tzdata"
+version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


zwpaper:
Hi @decathorpe, thanks for the update, as you can see CI, once we upgrade the `dirs` dep to `v5`, the Windows build will fail.

can you help to fix it?

decathorpe:
I briefly looked at the failure, and it seems to be a toolchain issue? I have no idea why updating Rust code should cause linker failures.

If it's dirs v5 that causes the issue, I can drop that one from the batch and we can look into the failure separately.

zwpaper:
that would also be okay to leave dirs in a separate PR

decathorpe:
I've dropped the dirs v4 → v5 update for now.

One thing that looks suspicious is that with v5, it now uses `windows-sys` instead of `winapi` ...

decathorpe:
I tried something else, maybe it changes things, maybe not. 🤞🏼

decathorpe:
Looks like good ol' `cargo update` fixed it. Judging by the previous contents of `Cargo.lock`, it hadn't been done in years, so it's not surprising that it pinned some crates to versions that were missing fixes.

(After `cargo update` I made small adjustments to keep the MSRV at 1.70 - clap v4.5 bumped its MSRV to 1.74, so I kept it at v4.4.)

zwpaper:
thanks so much for the update!

decathorpe:
Great, thanks!

PR #985, opened 2024-02-08
📝 document `--date=locale`

zica87:
This flag is added by #840


zwpaper:
good catch @zica87, thanks

PR #1019, opened 2024-03-25
rollback term grid to 0.1 and release v1.1.2

zwpaper:


decathorpe:
This is reverting part of one of my previous changes - is there a specific reason this dependency is downgraded again?

PR #1025, opened 2024-03-26
Update permission options in README; minor tweaks

adrian5:
The `permissions` option appears to be `disable`.


zwpaper:
thanks @adrian5

PR #1028, opened 2024-04-09
Disable classify when piped

### LINKED ISSUE #918: [Feature]: add an option to turn of --classify
### Version

- [X] latest

### version

lsd 1.0.0

### What OS are you seeing the problem on?

Linux

### installation

NixOS

### term

_No response_

### ls-colors

_No response_

### What happened?

```markdown
I have `lsd` configured with `indicators = true`. The indicators stay present `lsd` is used non-interactively, i.e it's output is piped. 


$ mkdir -p test; cd test
$ touch foo; chmod +x foo
$ lsd | xargs rm
rm: cannot remove 'foo*': No such file or directory
$ lsd
  foo*
```

There should be no files remaining. Instead, `foo` was never removed.
```


### What expected?

```markdown
We should be able to run `lsd --classify --indicator-style=none` and the two options would cancel each other out, showing no suffixes for special files. This would mirror the option in `ls`, the recommended way to turn off `--classify` according to https://superuser.com/a/220680 

In general, each configurable setting that can be turned on with an option should have an option to turn it off.
```

### What else?

```markdown
Another potential approach is to hide the classifiers when stdout is not a tty, maybe configurable by a setting? A downside is compatibility with `ls` would break.
```

bryanjhv:
The `--classify` flag (`indicators` option) is turned on when piped to other commands.
This PR fixes that, disabling `display_indicators` when output is not a TTY. Fixes #918.

---
#### TODO

- [x] Use `cargo fmt`
- [ ] Add necessary tests

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -1,7 +1,8 @@
 use crate::color::Colors;
 use crate::display;
 use crate::flags::{
-    ColorOption, Display, Flags, HyperlinkOption, Layout, Literal, SortOrder, ThemeOption,
+    ColorOption, Display, Flags, HyperlinkOption, Indicators, Layout, Literal, SortOrder,
+    ThemeOption,
 };
 use crate::git::GitCache;
 use crate::icon::Icons;
@@ -74,6 +75,7 @@ impl Core {
             inner_flags.layout = Layout::OneLine;
 
             flags.literal = Literal(true);
+            flags.display_indicators = Indicators(false);
         };
 
         let sorters = sort::assemble_sorters(&flags);


zwpaper:
hi @bryanjhv, thanks for the contribution, but the classify is enabled only when you turn it on, this is also how the GNU ls works

the issue is more about adding a flag `--indicator-style=none`.

so I am closing this PR for now, and feel free to open one for the `--indicator-style=none` flag

bryanjhv:
Oh, now I see. Seems the most viable option to remain compatible with GNU ls.

PR #1056, opened 2024-06-01
Fixed lack of space between executable construction icon and the start of file name

Mersid:
Makes executable files go from looking like this:
image

To this:
image

Notice in the original image there's no spacing between the construction crane icon and the file's name.

---
#### TODO

- [X] Use `cargo fmt`
- [X] Add necessary tests
- [X] Update default config/theme in README (if applicable)
- [X] Update man page at lsd/doc/lsd.md (if applicable)

### DIFF
--- a/src/meta/mod.rs
+++ b/src/meta/mod.rs
@@ -308,7 +308,7 @@ impl Meta {
                 ),
                 Err(e) => {
                     eprintln!(
-                        "lsd: {}: {}(Hint: Consider using `--permission disabled`.)",
+                        "lsd: {}: {}(Hint: Consider using `--permission disable`.)",
                         path.to_str().unwrap_or(""),
                         e
                     );


zwpaper:
hi @Mersid, thanks for the PR, but this is related to the icons, can we should not fixed it like this, because when the icon sits with other icons of the same size, it will have extra space.

maybe we should try to detect the width of icons, but it would be much more complicated

Mersid:
Heya @zwpaper! Probably. I can't say for certain what the issue is, but if you try using unicode-width, it appears that the construction icon is marked as having a width of 1, but others like the file and folder icons have a width of 2. However, when rendering, this is obviously not the case, so thus the construction icons are rendered too close together. Just adding the space may be a viable option until we find something better. For now, having some stuff being misaligned does look quite odd.
image

zwpaper:
Hi @Mersid, thanks for the info on unicode-width, it helps!

it seems reasonable for the default unicode theme to add this space, but we should calculate the width and handle the icon width dynamically.

can you create a issue for use unicode-width to handle this, and add as TODO comment here, so that we can fix this when we have unicode-width

Issue #1063, opened 2024-06-16
Lsd long output is not properly aligned


pwnalone:
- os: `Linux kali 6.6.15-amd64 #1 SMP PREEMPT_DYNAMIC Kali 6.6.15-2kali1 (2024-05-17) x86_64 GNU/Linux` (also observed on Ubuntu under WSL)
- `lsd --version`: 1.1.2 (also observed with 1.0.0)
- `echo $TERM`: `tmux-256color` (also observed with `alacritty`)
- `echo $LS_COLORS`: (empty)

## Expected behavior

The output should be properly aligned as it is with vanilla `ls`.

![image](https://github.com/lsd-rs/lsd/assets/3609458/4e1bd588-afb2-4dd5-9120-9e33944078f1)

## Actual behavior

The output is not aligned properly.

![image](https://github.com/lsd-rs/lsd/assets/3609458/6072b9c2-58de-44c1-a231-98ac85558db8)

ehwan:
I'm having same issue on OSX Terminal.app

But it worked good when I put `--color=never`.

![image](https://github.com/lsd-rs/lsd/assets/15500385/7b136a71-6a5e-4842-92fd-d882e9416a85)

![image](https://github.com/lsd-rs/lsd/assets/15500385/80c40abe-2d18-4484-a6d2-686b4b47a06b)

zwpaper:
is there any chance I can reproduce this?

thomcc:
I have this bug too, it reproduces easily macOS and Linux on every folder I've tried.

efvincent:
same issue. version 1.1.2, installed today with 
```bash
cargo install --git https://github.com/lsd-rs/lsd.git --branch master
```
KDE Neon (Ubuntu) 6.0
Plasma 6.1.2
KDE Framework 6.3.0
Qt 6.7
Kernel 6.5.0-41-generic 64bit
Wayland

terminal: Kitty 0.21.2
font: Fira Code Nerd Font Mono Regular

msrd0:
@efvincent Add `--locked` to your command and the issue will go away

efvincent:
> @efvincent Add `--locked` to your command and the issue will go away

That did it... didn't notice the related issue that had the answer - thanks for pointing it out! 👍🏽

SamuelLarkin:
Hi,
  just to add another data point.  In my case I get the directories to be misaligned with the files or is it the other way around 😕 

```sh
lsd --version
lsd 1.1.2
[...]

Issue #1054, opened 2024-05-28
Names not aligned when using  nerd font glyphs and Unicode emojis cusomization file


vernj123:
- os: linux mint mate 21.3
- `lsd --version`: 1.1.2
- `echo $TERM`: xterm-256color
- `echo $LS_COLORS`: 

## Expected behavior
Names to be aligned when using nerd fort glyphs and unicorn emojis

## Actual behavior
Miss-aligned

 Noticed If I use the customization file option provided in my icon/yaml the names are no longer aligned.
I am using Fancy setting.  
Not the end of the world but would be great to know if there is a solution for this specific issue.  

![Alignment](https://github.com/lsd-rs/lsd/assets/6043608/d8499642-4e85-41e9-99fe-2614917a8a4f)
 .

hoyhoy:
Everything is misaligned on Terminal.app.

PR #1064, opened 2024-06-18
Fixed misaligned outputs and unicode-width dependency version

ehwan:
This could possibly fix misaligned output issues #1063

The problem occured only if when I `cargo install lsd`, not when I build it directly from the clone or install it via `brew`.

This is because of the dependency `unicode-width`'s version. The version installed via `cargo install` was using the latest `unicode-width`, v0.1.13, but others were using v0.1.11 because of `Cargo.lock`.

The dependency `unicode-width` had a [breaking change](https://github.com/unicode-rs/unicode-width/issues/55) between v0.1.11 and v0.1.13 that returns a different result when [counting unicode characters](https://github.com/lsd-rs/lsd/blob/5d02a59e29a755baeb33a546eabfc6914fdedf76/src/display.rs#L458). 

The new behavior in `unicode-width` v0.1.13 is more intuitive for calculating the length of a unicode string, so I fixed the logic accordingly.

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1257,9 +1257,9 @@ dependencies = [
 
 [[package]]
 name = "unicode-width"
-version = "0.1.11"
+version = "0.1.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e51733f11c9c4f72aa0c160008246859e340b00807569a0da0e7a1079b27ba85"
+checksum = "0336d538f7abc86d282a4189614dfaa90810dfc2c6f6427eaf88e16311dd225d"
 
 [[package]]
 name = "unsafe-libyaml"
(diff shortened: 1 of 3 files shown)


zwpaper:
hi @ehwan, thanks so much for fixing the issue, but it seems not to be helping https://github.com/lsd-rs/lsd/issues/1054, can you add some tests so that we can make sure it works?

I cloned your code and change one of my icon to unicode:

CleanShot 2024-06-22 at 11 49 22@2x

ehwan:
Hi @zwpaper, and thank you for reviewing.

I apologize that I misunderstood the term 'misaligned' in #1054. This PR does not fix the issue.
I've deleted the mention in the description.

ehwan:
And by the way, for #1054, should we add extra space after a 1-width icon if a 2-width icon is present in the same column?

hucklesberries:
This fix worked for me, running Debian on an rPi5.

zwpaper:
sorry for the late review, I have tested it and made sure it was really caused by the dependency break.

https://github.com/unicode-rs/unicode-width/issues/55

let's merge this and create a release for it!

zwpaper:
thanks so much for the help @ehwan

PR #1069, opened 2024-06-26
Update README.md

tomterl:
serde-0.9 no longer has this bug; 
testing lsd with a zero-byte colors.yaml works just fine showing the default theme.

---
#### TODO

- [ ] Use `cargo fmt`
- [ ] Add necessary tests


tomterl:
This is a readme only change - how important/possible are the standard todos above?

zwpaper:
hi @tomterl, good catch! it's ok to ignore them for a doc update

PR #1099, opened 2024-10-13
Draft: Add a --json flag to output json.

jorikvanveen:
This PR adds a --json flag which causes lsd to output the Metas as JSON.

It's a basic proof of concept I put together in an evening, so there are some `.unwrap()`s and rough edges. It's meant to show what I imagine this feature to look like so you can decide whether this is something you actually want implemented in lsd. If it is I can start working on a more serious implementation.

## Why?
Combining this flag with something like nushell allows users to create ls aliases that output structred nu data but still look pretty.

For example, this alias:
```nu
def --wrapped ls [...rest] {
        lsd ...$rest --json --icon always --icon-theme fancy --color always
        | from json
        | get 0.content
        | select type display size date
        | update date {|row| $row.date | into datetime}
        | update size {|row| $row.size | into filesize}
      }
```

results in this output: 
image

and you can then use other nushell operations to do something with this data:
image

#### TODO

- [x] Use `cargo fmt`
- [ ] Add necessary tests


dolence:
Will this be merged? Another nushell user here

PR #1138, opened 2025-05-26
Allow sorting by name

### LINKED ISSUE #1121: [Bug]: `--sort name` is not a valid option
### Version

- [x] latest

### version

lsd 1.1.5

### What OS are you seeing the problem on?

macOS

### installation

brew install lsd

### term

xterm-256color

### ls-colors

_No response_

### What happened?

```markdown
Ran `lsd --sort name`, no bueno.
```

### What expected?

```markdown
I expected the option to work as normal.
```

### What else?

```markdown
Currently, if you specify a default sort in the config, there is no way to revert it back to a name sort on the command-line as the option is not accepted.
```

AdamGaskins:
Closes #1121 

In the issue above, @5HT2 has a config that sets a different default sort field, which is what they want most of the time. The rare times that they do want to actually sort by name, there's no way to do this without modifying the config.

This PR adds support for sorting by name.

---

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [ ] ~Update default config/theme in README (if applicable)~


5HT2:
Thank you so much!! I really appreciate you taking the time to put this together :)

Issue #1014, opened 2024-03-22
File list is totally vertical if the list doesn't fit into one line unless run with `-a`


Un1q32:
- os: macOS 14.4
- `lsd --version`: lsd 1.0.0 (it's the 1.1.0 binary, see #1012)
- `echo $TERM`: xterm-kitty
- `echo $LS_COLORS`: (none)

## Expected behavior

Wrapping behavior should be the same as 1.0.0, or the same as when run with `-a`

## Actual behavior
<img width="1280" alt="Screenshot 2024-03-22 at 5 29 28 PM" src="https://github.com/lsd-rs/lsd/assets/57970114/09763c84-f28c-4957-9175-ecfd634e6f70">

Zooming in shows a normal run can also have this problem if one line isnt enough to fit

<img width="1280" alt="Screenshot 2024-03-22 at 5 30 40 PM" src="https://github.com/lsd-rs/lsd/assets/57970114/ec4edc64-266d-4b7d-a80f-112e60f98998">

wwwjfy:
It seems `-a` doesn't matter (much) here.
My observation is if non-`.` items cannot fit in one line, it'll show one item per line.

In the above case non-`.` items are not many, so even though with both it's wider than one line, it still behaves as expected.

I'm using iTerm2. This is new in the last release.

nogweii:
* `lsd --version`: lsd 1.1.1
* `echo $TERM`: xterm-kitty
* Arch Linux

Running into the same issue.

I've aliased ls to `/usr/bin/lsd --icon=auto --color=auto -F`, and indeed adding -a will make wrap files as expected.

oneElectron:
I am also having this issue.
- ```lsd --version```: lsd 1.1.1
- ```echo $TERM```: alacritty
- MacOS 14.4

FischLu:
I have the same issue

* `lsd --version`: lsd 1.1.1
* `echo $TERM`: xterm-256color
* MacOS 14.4

nogweii:
Hm, I think I've found the root cause - `term_grid` [doesn't ignore ANSI codes when calculating the width of each entry](https://github.com/ogham/rust-term-grid/issues/13). That could be tricky to solve...

CarloCattano:
came here looking for others having the same. 
lsd 1.1.1

ls returns a vertical list of files if the size cant hold them all in one line

CarloCattano:
it works with the -v flag for some reason
```bash
lsd -v
```

[...]

PR #1185, opened 2026-01-26
Fix Issue #1014 and Improve Header Underline Styling

rly0nheart:
---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [ ] Update README (if applicable)
- [ ] Update config sample file in `doc/samples` (if applicable)
- [ ] Update icon sample file in `doc/samples` (if applicable)
- [ ] Update color sample file in `doc/samples` (if applicable)

---

## Description
So I've been working on a similar ls utility myself and I've been taking a lot of inspiration from this project and eza. While poking around the code, I noticed a couple of things that I thought I could 'fix'.

## Header Underline Fix

The main thing was the header underlines in long format (`lsd -l --header`). Previously, the underlines would stretch across the entire column width with padding, which looked kind of off when you had varying column widths. I thought it made sense for the underline to stay within the bounds of the actual header text, that way, it looks visually even.

### Before:
Screenshot From 2026-01-26 03-29-07

---

Before the initial header underline change,  I updated term_grid to 0.2.0, and ran into errors relating to a missing alignment specification, this led me to change the header alignment so text columns (like Name, User, Group) are left-aligned and numeric columns (like Size, Date, INode, Links) are right-aligned. This made it easier to fix the header underline 'problem'.

The fix was in `add_header()` in `src/display.rs`, so instead of manually padding headers and then underlining the whole thing, I now just underline the header text and let term_grid handle the alignment via its `Alignment` enum.

### After:
Screenshot From 2026-01-26 03-22-51

> [!Note]
> If the wide header underline is intentional, then you should just ignore this bit :)
 
## Issue #1014 Fix

After I made that change, I thought it would be nice to fix an issue from lsd, so I noticed issue #1014.

The root cause was that term_grid 0.2 doesn't automatically strip ANSI escape codes when calculating cell widths. The fix is actually already in the code, when creating cells, we use `get_visible_width()` which properly calculates the visible width excluding ANSI codes:

```rust
cells.push(Cell {
    width: get_visible_width(&block, flags.hyperlink == HyperlinkOption::Always),
    contents: block,
    alignment: Alignment::Left
});
```

This way term_grid gets the correct width (e.g., 6 for "f1.txt") instead of the full string length including escape codes (e.g., 22 for "\x1b[38;5;184mf1.txt\x1b[39m").

Added a test `test_grid_layout_with_colors_issue_1014` to make sure this doesn't regress.

## Files Changed

- `src/display.rs` - Header alignment fix and issue #1014 test
- `src/flags/blocks.rs` - Added `is_numeric()` method to determine column alignment
- `Cargo.toml` - Updated term_grid to 0.2.0, and once_cell to 1.21

## Testing

All 437 tests pass (392 unit + 45 integration).


rly0nheart:
@zwpaper I reverted the header underline style, but the #1014 fix remains. All tests passed.

PR #1202, opened 2026-04-13
Fix recursion output indentation and spacing

Shonas301:
I've been using `lsd -R` for deep walks and it's hard to follow — entries sit
flush-left regardless of depth, and there are multiple blank lines between
subdirectory blocks. No visual nesting at all.

Changes:
- indent each recursive entry by `2 * depth` spaces (both headers and content)
  so the nesting reads visually
- collapse runs of blank lines between directory blocks down to a single separator
- tests covering per-depth indent at depth 1 and 2, plus a no-double-blank invariant

Before:

```
subdir:
file.rs

subdir/nested:


deep.rs
```

After:

```
  subdir:
    file.rs

    subdir/nested:
      deep.rs
```

Display-layer only — nothing in traversal or meta collection changed.

---
#### TODO

- [x] Use `cargo fmt`
- [x] Add necessary tests
- [ ] Update README (if applicable)
- [ ] Update config sample file in `doc/samples` (if applicable)
- [ ] Update icon sample file in `doc/samples` (if applicable)
- [ ] Update color sample file in `doc/samples` (if applicable)

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -14,6 +14,7 @@ const EDGE: &str = "\u{251c}\u{2500}\u{2500}"; // "├──"
 const LINE: &str = "\u{2502}  "; // "│  "
 const CORNER: &str = "\u{2514}\u{2500}\u{2500}"; // "└──"
 const BLANK: &str = "   ";
+const INDENT_STEP: usize = 2;
 
 pub fn grid(
     metas: &[Meta],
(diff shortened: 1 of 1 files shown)


Shonas301:
image

Sorry about the cold open of this and #1203 ! screen shot is attached for what it looks like! If there are changes requested just lmk, nbd, if you don't think it belongs in the upstream just lmk.

PR #1205, opened 2026-04-18
sanitize filesystem strings before terminal output

curious-rabbit:
Attacker-controlled filenames, symlink targets and paths could inject ANSI sequences when lsd rendered them to stdout or stderr.

Adds a sanitize_for_terminal helper and a SafePath display wrapper in a new display_util module. Applies them to symlink rendering, folder header output, error messages and the name escape path. Also covers bidi formatting characters used for filename spoofing

Update:
The patch now also sanitizes xattr contexts and replaces two panic sites for non-utf8 symlink targets and non-utf8 filennames in git status that were reachable via filesystem input.

### DIFF
--- a/src/core.rs
+++ b/src/core.rs
@@ -124,7 +124,7 @@ impl Core {
                 match Meta::from_path(&path, self.flags.dereference.0, self.flags.permission) {
                     Ok(meta) => meta,
                     Err(err) => {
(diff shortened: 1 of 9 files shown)


curious-rabbit:
Is there any interest in merging this? coreutils "ls" takes significant care to escape control sequences and this is still entirely missing from lsd

PR #1229, opened 2026-08-08
Use cargo_bin!() macro instead of deprecated cargo_bin fn in integration tests

### LINKED ISSUE #1178: Tests rely on deprecated `assert_cmd::cargo_bin` which will break on future Cargo versions
https://github.com/lsd-rs/lsd/blob/351acdaa80a4ebd1abdf194a8b2017ce8eed250f/tests/integration.rs#L620

`cargo_bin` relies on implementation details of cargo that will soon break (rust-lang/cargo#16147, rust-lang/cargo#15010).

You can test this by running:
```
CARGO_BUILD_BUILD_DIR="{cargo-cache-home}/build/{workspace-path-hash}" cargo test
```

The replacement is `assert_cmd::cargo_bin!`

MsfPablo:
Closes #1178.

The integration-test helper `cmd()` used `Command::cargo_bin(env!("CARGO_PKG_NAME")).unwrap()`, which relies on `assert_cmd` probing the `target/` layout. As tracked in #1178, that path is fragile under non-default build directories (e.g. `CARGO_BUILD_BUILD_DIR`) and the `cargo_bin` function has been deprecated in favor of the `cargo_bin!` / `cargo_bin_cmd!` macros.

This switches the helper to:

```rust
fn cmd() -> Command {
    Command::new(assert_cmd::cargo_bin!())
}
```

`cargo_bin!()` resolves the binary through the official `CARGO_BIN_EXE_lsd` env var that Cargo sets for the package's own bin target, so it is robust to the build directory and no longer needs an `.unwrap()`.

### Dependency change
`cargo_bin!()` landed in assert_cmd 2.0.17, so this bumps the `assert_cmd` dev-dependency in the lockfile from 2.0.14 → 2.2.2. The `Cargo.toml` constraint stays at `assert_cmd = "2"` (no tightening); this is a lockfile-only update.

### Verification
- `cargo build --tests` — clean
- `cargo test` — 42 passed; the only failure (`test_date_custom_format_supports_nanos_with_length`) is pre-existing and locale-dependent (`%.3f` emits a comma under `LC_TIME=es_ES.UTF-8\)); it fails identically on clean `main` and passes under the `en_US`/C locale used in CI.
- `cargo fmt --check` — clean
- `cargo clippy --all-targets` — no new warnings (the remaining ones are pre-existing in `src/display.rs`)


MsfPablo:
Ping on this. It's a mechanical swap to the `cargo_bin!()` macro, so it should be quick to eyeball. Let me know if you'd rather fold it into a broader test cleanup instead.

PR #1230, opened 2026-08-09
test: make date nanos test locale-agnostic (#1223)

### LINKED ISSUE #1223: test_date_custom_format_supports_nanos_with_length fails on macOS (aarch64-darwin)
### Environment
- lsd version: 1.2.0
- OS: macOS (aarch64-darwin / Apple Silicon)

### Description

The integration test `test_date_custom_format_supports_nanos_with_length` fails when building lsd on macOS. The test creates two files and expects their timestamps to differ at the nanosecond level, but APFS does not expose sub-millisecond timestamp precision to userspace — both files end up with timestamps that are identical at that resolution, so the assertion fails.

```
failures:
    test_date_custom_format_supports_nanos_with_length

test result: FAILED. 42 passed; 1 failed; 0 ignored
```

### Expected behavior

The test should either be skipped on platforms that don't provide nanosecond-precision timestamps (macOS, BSD), or rewritten to not rely on filesystem nanosecond resolution.

### Suggested fix

Guard the test with `#[cfg(not(target_os = "macos"))]`, or use a synthetic timestamp instead of relying on `mtime` nanoseconds from `std::fs`.

MsfPablo:
## Summary

Fixes #1223.

`test_date_custom_format_supports_nanos_with_length` asserts the output of `--date +testDateFormat%.3f` matches the regex `testDateFormat\.[0-9]{3}` (a literal `.`).

## Root cause

chrono's `format_localized` (used for `DateFlag::Formatted` in `src/meta/date.rs`) renders the `%.3f` fractional seconds using the **active locale's decimal separator**. That separator is `,` in locales such as `es_ES`, `de_DE`, `fr_FR` — not `.`. On those locales the literal-`.` regex matches 0 of the 2 `testDateFormat,NNN` lines, so the test fails:

```
assertion `left == right` failed
  left: 0
 right: 2
```

This reproduces on my machine (`LC_ALL=es_ES.UTF-8`).

## Fix

The production behaviour — localising the decimal separator — is arguably correct (a user in `es_ES` expects `,`), so this is a **test-only** fix: accept either `.` or `,` as the decimal separator:

```rust
predicate::str::is_match("testDateFormat[.,][0-9]{3}")
```

The assertion only cares that three fractional digits follow a decimal separator, which holds under both separators.

## Verification

- `cargo test --test integration test_date_custom_format_supports_nanos_with_length` passes under `LC_ALL=es_ES.UTF-8` (comma)
- same test passes under `LC_ALL=C` (dot)
- full integration suite under `es_ES.UTF-8`: **43 passed, 0 failed**
- `cargo fmt --check`: clean
- `cargo clippy --tests`: no new warnings introduced (7 pre-existing warnings are unchanged by this edit, confirmed via stash diff)

---

🤖 This is an AI-assisted contribution. The analysis, root-cause investigation, and fix were produced with the help of an AI assistant (Claude Code) and reviewed by me; the code is my own and I stand behind it.


MsfPablo:
Gentle ping — this one's still green and applies cleanly. Happy to rebase if it's gone stale on your end.

PR #1233, opened 2026-08-14
fix: make -h/--human-readable actually apply

### LINKED ISSUE #880: Feature request: `-h` should override previous `--size`
The `-h` flag should not do nothing, but be equivalent to `--size default`.

The reason is that if an alias sets e.g. `--size bytes`, to emulate the default behavior of `ls`, an option to the alias invocation `-h` is naturally expected to override that.

```bash
ls        # shows size in bytes
ls -h     # shows size human-readable

alias ls='lsd --size bytes'
ls        # = shows size in bytes (= 'lsd --size bytes')
ls -h     # should show size human-readable but doesn't (= 'lsd --size bytes -h')
````

- os: Debian GNU/Linux 12 'bookworm'
- `lsd --version`: 0.23.1

### PR DESCRIPTION
**Pull request by VXNCXNX**
Closes #880.

## The problem

`-h` / `--human-readable` is parsed and then never used:

```
$ grep -rn human_readable src/
src/app.rs:61:    human_readable: bool,
```

One hit, the declaration. Nothing reads it. So the flag has no effect at all:

```
$ lsd -l --size bytes -h
5000                 <- raw bytes, -h ignored
```

## The fix

Make `-h` and `--size` override each other, so the later flag wins:

```
lsd -l --size bytes -h            ->  4.9 KB     (was 5000)
lsd -l -h --size bytes            ->  5000       (--size wins, as it should)
lsd -l --size bytes -h --size short -> 4.9M      (last of three wins)
```

This uses clap's `overrides_with`, which the repo already relies on for `all`/`almost_all` in the same file, and `args_override_self = true` is already set on the struct, so it's the existing idiom rather than manual index comparison.

`--classic` still takes precedence over both, unchanged. Config-file `size:` still works and the CLI still beats it:

```
config size: bytes                 ->  5000
config size: bytes  +  CLI -h      ->  4.9 KB
```

## One change that needs a sentence

`human_readable` had to become `pub` so `flags/size.rs` can read it, which left `help: ()` as the struct's only private field, and that newly trips clippy's `manual_non_exhaustive`. I made `help` `pub` too, matching every other field.

I checked whether this is a public API concern: it isn't. lsd has no lib target (only `[[bin]]` in Cargo.toml) and `mod app;` is private in `main.rs`, so `Cli` is unreachable from outside the crate. The alternative clippy suggests, `#[non_exhaustive]`, would be wrong here since clap needs the field.

## Tests

Four tests added to `flags/size.rs`'s existing test module, covering both orderings, `--classic` precedence, and the config interaction.

```
cargo build:         clean
cargo test:          395 passed (unit), 44 passed (integration)
cargo fmt --check:   clean
cargo clippy --all-targets: no new warnings in the changed files
```

One integration test, `test_date_custom_format_supports_nanos_with_length`, fails in my environment on a locale decimal separator. It fails identically on a clean checkout, so it's unrelated to this change.

---
#### TODO

- [x] Use `cargo fmt` (`cargo fmt --check` clean)
- [x] Add necessary tests (four in `flags/size.rs`)
- [ ] Update README (if applicable) (not applicable: the README already documents `-h`/`--human-readable`; this makes the documented behaviour true rather than changing it)
- [ ] Update config sample file in `doc/samples` (if applicable) (not applicable: no config key is added or changed, the `size:` key is unchanged)
- [ ] Update icon sample file in `doc/samples` (if applicable) (not applicable: no icon change)
- [ ] Update color sample file in `doc/samples` (if applicable) (not applicable: no color change)
- [ ] Update man page at lsd/doc/lsd.md (if applicable) (not applicable: the man page already documents `-h`/`--human-readable`)

*Disclosure: written with AI assistance (Claude Code). I reproduced the issue, ran the change and the verification myself.*

### CODE REVIEW

**Comment by VXNCXNX** [comment]
Rebased both onto current `main`. Worth saying up front that the red X is not these changes: the failing jobs are the `use-cross` builds, and they fail the same way on `main` itself (run 32004887621, same `i686` / `riscv64gc` / `aarch64` targets). They die inside dependency crates, `libc`, `serde_core` and `zerofrom_derive`, in the cross image rather than in any lsd code. The last fully green cross run on this repo was 16 August.

Both PRs pass locally on the rebased base:

```
#1233   397 passed, 45 passed    0 failed
#1235   393 passed, 47 passed    0 failed
```

One note if you run these yourself: `test_date_custom_format_supports_nanos_with_length` is locale-dependent and fails under a comma-decimal locale, so it wants `LC_ALL=C`. That is pre-existing, not from these branches.

- #1233 closes #880: `-h` / `--human-readable` is parsed and then never read, so the flag has no effect at all.
- #1235: `--config-file /nope/x.yaml` panics with `Provided file path is invalid` and exits 101 instead of failing cleanly.

